PR #54, opened 2016-10-16
41 day usage graph

ldodds:
Created a PR to review the WIP branch


ldodds:
@dynamicworkplace 

Graphs starting to shape up, couple of bits of feedback:

* there are negative values displaying, but from a quick look at the sample data and what's created in the database, there's no negative values in the database? (e.g. `/schools/3/usage?period=daily&to_date=2012-11-30`)

* I think the above is caused by the logic in `usage.rb` this seems to be finding the first and last readings for a given day, for a given meter and then taking the difference? It should be adding up the 48 half-hourly readings for each meter to find the total for each day.

* as noted on Slack, the default should be to find and show the latest data that we have (based on the last full week of data), even if its slightly older, so that there's something to show by default. We can a note to say when we last received data.

* it looks like the charts are being delivered by `http://chart.apis.google.com/`. Google deprecated the image charts in 2012, so we should be using the JS API described here: https://developers.google.com/chart/. I don't know if there's rails helpers for that, but it'll involve some JS in the page to build the chart and then an AJAX call to the app to fetch the data (or at least that's how I've done it before

* it looks like when we have incomplete data for a week, then not all days of the week are display on the x-axis of the graph, (see, e.g. same link as above) The chart axes should be consistent, even if we have no data

* y-axis needs some labels

* can we add annotations on hovering over a bar so that it shows the date and value?

PR #98, opened 2016-11-28
Achievements

MattHelsby:


ldodds:
As discussed on Slack, when there are existing schools in the database the code fails because of missing data:

```
def badges_by_date(order: :desc, limit: nil)
    sash.badges_sashes.order(created_at: order)
      .limit(limit)
      .map(&:badge)
  end
```

PR #368, opened 2018-11-09
School groups and scoreboards

jamesalmond:


db/migrate/20181109120333_create_school_groups_and_scoreboards.rb
```diff
@@ -0,0 +1,22 @@
+class CreateSchoolGroupsAndScoreboards < ActiveRecord::Migration[5.2]
+  def change
+
+    create_table :scoreboards do |t|
+      t.string :name, null: false
```
jamesalmond: @jamesjefferies as far as I can see, this gets translated into an unrestricted `varchar`/`character varying`, which behaves the same as a a `text` type. I can change them to `text` if you think that makes it clearer?

lib/tasks/school_groups.rake
```diff
@@ -0,0 +1,16 @@
+namespace :school_groups do
+  desc 'Create default school groups and scoreboards'
+  task create_default_school_groups_and_scoreboards: :environment do
```
jamesalmond: @jamesjefferies I've just added a rake task for populating some example data. The after_party gem doesn't look like it's been updated to support Rails 5 (although, I think it _might_ work on Rails 5) so I'm wary of introducing it as a dependency https://github.com/theSteveMitchell/after_party

db/migrate/20181109120333_create_school_groups_and_scoreboards.rb
jamesjefferies: @jamesalmond nope - if string has unrestricted length then I'm happy for it to be string and I'll make a mental note that we can use string instead of text if we like

lib/tasks/school_groups.rake
jamesjefferies: @jamesalmond ok - well, delving in to producing a rails5 compliant version maybe a simple PR for after_party or it may be a flippin' nightmare - so rake task for now I guess

PR #376, opened 2018-11-15
Meter management

jamesalmond:


app/controllers/schools/meters_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+  private
+
+    def load_meters
+      @active_meters = @school.meters.active.order(:meter_no)
+      @inactive_meters = @school.meters.inactive.order(:meter_no)
+    end
+
+    def set_school
+      @school = School.friendly.find(params[:school_id])
+      authorize! :manage, @school
```
jamesalmond: Do you think we should authorize explicitly on the ability to manage meters?

config/routes.rb
```diff
@@ -28,6 +28,14 @@
     scope module: :schools do
       resources :contacts
       resources :alerts
+
+      resources :meters do
+        member do
```
jamesalmond: I'd considered adding these as one or two separate controllers instead of members. Any preference?

app/views/shared/_errors.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if subject.errors.any? %>
```
jamesalmond: There are some bits in here that I've replicated from my other PR. WIP etc etc....

jamesalmond:
@jamesjefferies is there something I need to do with the `school_times` functionality on the `schools#new` form? I just looks blank and there are no times to fill in.

app/controllers/schools/meters_controller.rb
jamesjefferies: Might be a question to posit for Claudia & Philip - can school admins set up the meters, or should a super admin only be able to do it? If the latter, then yes we should, the former I think it's ok...

config/routes.rb
jamesjefferies: as an overall preference, I like my controllers to have the usual restie kind of actions - but I'm happy to keep them here for now. I would be interested to see what those two separate controllers would be like though - would they just both have update actions?

jamesjefferies:
> @jamesjefferies is there something I need to do with the `school_times` functionality on the `schools#new` form? I just looks blank and there are no times to fill in.

This is on the 'we need to fix school onboarding' process - ideally, when you create a new school, it creates your school times for you - I don't think it does that at the moment, it's a behind the scenes task. So I don't think it needs doing as part of this PR, but it is something on the onboarding list which we might be best looking at sooner rather than later (due to new schools)

config/routes.rb
jamesalmond: A few options, off the top of my head:
`Schools::ActivationController#create` and `Schools::DectivationController#create`
or
`Schools::ActiveStatusController#update`
or
`Schools::ActivationController#create` and `Schools::ActivationController#destroy`

My preference would be for at least two separate actions, so not the middle option. All depends on which ones are more RESTy.

jamesjefferies: Yes, I think I'm with you on 1 or 3 - I think 3 might be my preference.. although is it anything to do with Schools ? Should it be Meters::ActivationController ? ;)

jamesalmond:
> > @jamesjefferies is there something I need to do with the `school_times` functionality on the `schools#new` form? I just looks blank and there are no times to fill in.
> 
> This is on the 'we need to fix school onboarding' process - ideally, when you create a new school, it creates your school times for you - I don't think it does that at the moment, it's a behind the scenes task. So I don't think it needs doing as part of this PR, but it is something on the onboarding list which we might be best looking at sooner rather than later (due to new schools)

Ok, thanks.

jamesalmond:
@jamesjefferies I've merged this into `aws-eb-test` and deployed.

PR #384, opened 2018-11-20
New structure for data feed 7 calendars, school types and KS4/5

jamesjefferies:


app/models/school.rb
```diff
@@ -69,7 +73,7 @@ class School < ApplicationRecord
   belongs_to :solar_pv_tuos_area
   belongs_to :school_group
 
-  enum school_type: [:primary, :secondary, :special, :infant, :junior]
+  enum school_type: [:primary, :secondary, :special, :infant, :junior, :middle]
```
jamesjefferies: @jamesalmond here is the new School Type!

PR #400, opened 2018-11-28
Fix for readings in a week bug

jamesjefferies:


spec/models/school_spec.rb
```diff
@@ -89,12 +89,58 @@
...
-      pending "further work on deprecating meter readings"
       meter_one = create(:gas_meter, school: subject)
       meter_two = create(:electricity_meter, school: subject)
 
       expect(subject.has_last_full_week_of_readings?).to be false
     end
+
+    it 'some readings' do
+      meter_one = create(:gas_meter, school: subject)
+      meter_two = create(:electricity_meter, school: subject, meter_type: :electricity)
```
jamesalmond: Don't need `meter_type` here

jamesjefferies: Correct - left over from previous test - will sort...

PR #406, opened 2018-11-30
Validating readings in app

jamesalmond:
screenshot 2018-11-30 at 13 18 06

### DIFF
--- a/app/views/schools/meters/index.html.erb
+++ b/app/views/schools/meters/index.html.erb
@@ -6,12 +6,21 @@
   </div>
 <% end %>
 
+<% if can?(:validate_meters, @school) && @school.meters_with_readings.any? %>
+  <div class="alert alert-secondary">
+    <%= link_to 'Validate meter readings', school_meter_readings_validation_path(@school), method: :post, class: 'btn btn-warning mr-3' %>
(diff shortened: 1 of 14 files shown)


jamesalmond:
@jamesjefferies I've moved the reports to this page too and left a few of the sensible tweaks to the testing and models that came from attempting to build genuine test data.

app/views/schools/meters/index.html.erb
```diff
@@ -27,9 +36,14 @@
           <td><%= meter.mpan_mprn %></td>
           <td><%= meter.name %></td>
           <td><%= meter.meter_type.humanize %></td>
+          <td><%= meter.amr_data_feed_readings.count %></td>
+          <td><%= meter.amr_validated_readings.count %></td>
           <td>
             <div class="btn-group">
               <%= link_to 'Edit', edit_school_meter_path(@school, meter), class: 'btn btn-sm btn-warning' %>
+              <% if current_user.admin? && meter.amr_validated_readings.any? %>
```
jamesalmond: This checks the user is an admin because that's what the reports controller checks (i.e. it doesn't use cancan). I can look changing that and updating this check as part of the permissions review task

PR #412, opened 2018-12-04
Refactor csv importer and validate persistence functionality

jamesjefferies:


app/models/amr_data_feed_config.rb
```diff
@@ -51,4 +51,16 @@ def array_of_reading_indexes(header = nil)
     header_array = this_header.split(',')
     reading_fields.map { |reading_header| header_array.find_index(reading_header) }
   end
+
```
jamesjefferies: Moved these in here rather than having them in the parser/importer

app/services/amr/csv_parser.rb
```diff
@@ -0,0 +1,45 @@
+require 'csv'
+
```
jamesjefferies: This just parses, including dropping the header row (if present) and doing the shift by one for Frome if appropriate. The output being a 'clean' array of rows for the importer to handle

app/services/amr/csv_parser_and_upserter.rb
```diff
@@ -0,0 +1,22 @@
+module Amr
+  class CsvParserAndUpserter
```
jamesjefferies: This now creates the log entry, calls the parser and calls the upserter

app/services/amr/csv_parser.rb
```diff
@@ -0,0 +1,45 @@
...
+module Amr
```
jamesjefferies: Moved these in to an Amr name space as it's amr csv files which are being handled here

app/services/amr/csv_upserter.rb
```diff
@@ -0,0 +1,66 @@
+require 'upsert'
+require 'upsert/active_record_upsert'
+require 'upsert/connection/postgresql'
+require 'upsert/connection/PG_Connection'
+require 'upsert/merge_function/PG_Connection'
+require 'upsert/column_definition/postgresql'
+
+module Amr
+  class CsvUpserter
```
jamesjefferies: This just does upserting now - possibly the parser could create an array of hashes with the relevant fields already set up rather than doing it here. Left it as is for now, but kind of feels like it would be cleaner for this not to need to now about the config...

app/services/amr/upsert_validated_readings.rb
```diff
@@ -0,0 +1,29 @@
+module Amr
+  class UpsertValidatedReadings
```
jamesjefferies: This has now been separated from the validate and persist structure we had before. This just upserts

app/services/amr/upsert_validated_readings_for_a_meter.rb
```diff
@@ -0,0 +1,31 @@
+module Amr
+  class UpsertValidatedReadingsForAMeter
```
jamesjefferies: This has also been separated out - upserts for a single meter at a time

app/services/amr/validate_and_persist_readings_service.rb
```diff
@@ -0,0 +1,17 @@
+module Amr
+  class ValidateAndPersistReadingsService
```
jamesjefferies: This used to do everything, it still does, but now delegates to the various services rather than doing it all by itself

app/services/amr/csv_parser.rb
```diff
@@ -0,0 +1,45 @@
...
+  class CsvParser
```
jamesjefferies: If we have services in the services folder, do we want to call them blah_service? or just blah? I think we have a mix of the two at the moment.... Happy to go either way

spec/factories/dashboard_meters_factory.rb
```diff
@@ -0,0 +1,51 @@
+require 'dashboard'
+
+FactoryBot.define do
```
jamesjefferies: This enables us to create dashboard meters (wot we get back from Analytics etc)

spec/factories/dashboard_one_day_amr_readings_factory.rb
```diff
@@ -0,0 +1,16 @@
+require 'dashboard'
+
+FactoryBot.define do
+  factory :dashboard_one_day_amr_reading, class: 'OneDayAMRReading' do
```
jamesjefferies: As above - these are the one day readings we get back from analytics

spec/services/amr/csv_parser_and_upserter_spec.rb
```diff
@@ -0,0 +1,293 @@
+require 'rails_helper'
+require 'fileutils'
+require 'fakefs/spec_helpers'
+
+module Amr
+  describe CsvParserAndUpserter do
```
jamesjefferies: THis is the old csv_importer spec - renamed - at the moment it's testing the underlying services, doing parsing and upserting - ideally at some point would be good to split those out (when a new test is required?)

app/services/amr/upsert_validated_readings_for_a_meter.rb
```diff
@@ -0,0 +1,52 @@
...
+        reading_date: one_day_reading.date,
+        kwh_data_x48: one_day_reading.kwh_data_x48,
+        one_day_kwh: one_day_reading.one_day_kwh,
+        substitute_date: one_day_reading.substitute_date,
+        status: one_day_reading.type,
+        upload_datetime: one_day_reading.upload_datetime
+      )
+    end
+
+    def existing_dates_to_be_deleted(new_validated_dates)
```
jamesjefferies: I think this functionality around deleting readings post upserting *should probably* be in a separate service, run after the upsert.

app/models/amr_data_feed_config.rb
jamesjefferies: Fair enough - I'll move it

app/services/amr/csv_upserter.rb
jamesjefferies: Yes - so a third part of the process to sit between the Parser and the Upserter - Validation & massgaging in to a hash or VO which can then be passed on

app/services/amr/upsert_validated_readings_for_a_meter.rb
jamesjefferies: yes - I wasn't sure about the size of where clauses - i.e. is there a data limit to what is allowed?

app/services/amr/upsert_validated_readings_for_a_meter.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      dates_to_delete = existing_dates_to_be_deleted(amr_data.keys)
+      deleted_count = delete_by_dates(dates_to_delete)
+      Rails.logger.info "Processing: #{@dashboard_meter} - deleted: #{deleted_count}"
+    end
+
+  private
+
+    def upsert_from_one_day_reading(meter_id, upsert, one_day_reading)
+      return if is_nan?(one_day_reading)
```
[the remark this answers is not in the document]
jamesjefferies: Yes - as below...

app/services/amr/validate_and_persist_readings_service.rb
```diff
@@ -0,0 +1,17 @@
...
+    def initialize(school)
+      @school = school
+      @meter_collection = AmrMeterCollection.new(@school)
+    end
+
+    def perform
+      return unless @school.meters_with_readings.any?
+      p "Validate and persist readings for #{@school.name} #{@school.id}"
+      AggregateDataService.new(@meter_collection).validate_meter_data
```
jamesjefferies: This should be reviewed in the Analytics code

app/services/amr/validate_and_persist_readings_service.rb
```diff
@@ -0,0 +1,17 @@
...
+      UpsertValidatedReadings.new(@meter_collection).perform
```
jamesjefferies: This does something in place and then the service returns it - which references @jamesalmond above

PR #424, opened 2018-12-12
Admin creation of school onboarding process

jamesalmond:


db/migrate/20181212110943_create_onboarding_records.rb
```diff
@@ -1,9 +1,8 @@
 class CreateOnboardingRecords < ActiveRecord::Migration[5.2]
   def change
-    enable_extension "pgcrypto" unless extension_enabled?("pgcrypto")
 
     create_table :school_onboardings do |t|
-      t.uuid :uuid, null: false, default: 'gen_random_uuid()'
```
jamesalmond: @jamesjefferies I've spent some un-fruitful time trying to get Travis to use PG 10 as this function doesn't exist on the version of PG that's running by default with no luck. Can never connect to the DB if I set a different version. And, I have no guarantee that this function is going to be available on production so I'm switching it out for a manual UUID for now. Unless you have any ideas?

jamesjefferies: @jamesalmond oh that's a pain - this is the only thing I saw from a quick look, they do now support 10 apparently, maybe it needs to be forced to use a certain OS version somewhere... 
```
addons:
  postgresql: "10"
  apt:
    packages:
    - postgresql-10
    - postgresql-client-10
env:
  global:
  - PGPORT=5433
```

jamesalmond: I've tried various combinations of this but it didn't seem to work. Might be worth trying it in isolation.

app/models/calendar_event.rb
```diff
@@ -67,20 +69,20 @@ def any_overlapping_here?(calendar_event, child_calendar, date_to_check)
   def update_neighbour_start
     if calendar_event_type.term_time
       following_holiday = calendar.holidays.find_by(start_date: end_date_before_last_save + 1.day)
-      following_holiday.update(start_date: end_date + 1.day) if following_holiday.present?
+      following_holiday.update!(start_date: end_date + 1.day) if following_holiday.present?
```
jamesjefferies: Should these ! be here @jamesalmond ?

app/views/admin/school_onboardings/email/new.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1>Onboarding: <%= @school_onboarding.school_name %></h1>
+
+<dl>
```
jamesjefferies: This does end up looking a bit small on the page - bootstrap default dl stuff I think

app/views/admin/school_onboardings/index.html.erb
```diff
@@ -0,0 +1,24 @@
...
+      <th>Last event</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @school_onboardings.each do |onboarding| %>
+      <tr>
+        <td><%= onboarding.school_name %></td>
+        <td><%= onboarding.contact_email %></td>
+        <td><%= onboarding.events.maximum(:created_at) %></td>
+        <td><%= SchoolOnboardingEvent.events.key(onboarding.events.maximum(:event)) %></td>
```
jamesjefferies: Might be nice to humanize (and check for null) this

app/views/admin/school_onboardings/index.html.erb
```diff
@@ -0,0 +1,24 @@
...
+      <th>Last activity</th>
+      <th>Last event</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @school_onboardings.each do |onboarding| %>
+      <tr>
+        <td><%= onboarding.school_name %></td>
+        <td><%= onboarding.contact_email %></td>
+        <td><%= onboarding.events.maximum(:created_at) %></td>
```
jamesjefferies: nice_dates helper call?

app/views/onboarding/consent/show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+      <div class="card-body">
+        <p>I would like to enrol <strong><%=@school_onboarding.school_name%></strong> in the Energy Sparks service to allow the school to better monitor our energy usage.</p>
+        <p>I understand that in order for this to happen, our energy usage data will need to be published under an open licence to Energy Sparks. This includes both ongoing and historical data.</p>
+        <p>I understand that publishing the data under an open licence will allow it to be accessed, used and shared by anyone, including the Energy Sparks team.</p>
+        <p>I would like to give permission to begin publishing our energy usage data at the earliest opportunity.</p>
+        <p>I understand that our school has the right to withdraw consent to our data being published to Energy Sparks at anytime. I can withdraw consent by emailing hello@energysparks.uk. We will then remove your ongoing and historical data from the Energy Sparks site.</p>
+      </div>
+    </blockquote>
+
+    <%= button_to 'I give permission', onboarding_consent_path(@school_onboarding), method: :post, class: 'btn btn-primary' %>
```
jamesjefferies: Maybe a back/cancel button or something? With the user knowing that they can return here without any probs.

db/schema.rb
```diff
@@ -10,10 +10,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_11_28_095621) do
+ActiveRecord::Schema.define(version: 2018_12_17_145048) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "hstore"
+  enable_extension "pgcrypto"
```
jamesjefferies: Not sure we need this do we?

PR #430, opened 2018-12-18
Rename category to the more meaningful fuel type

jamesjefferies:


app/services/alert_generator_service.rb
```diff
@@ -1,26 +1,28 @@
...
+    @gas_analysis_date = gas_analysis_date
+    @electricity_analysis_date = electricity_analysis_date
+    @send_sms_service = send_sms_service
 
-    account_sid = ENV['TWILIO_ACCOUNT_SID']
-    auth_token = ENV['TWILIO_AUTH_TOKEN']
-    @from_phone_number = ENV['TWILIO_PHONE_NUMBER']
-
-    @twilio_client = Twilio::REST::Client.new(account_sid, auth_token)
+    @aggregate_school = AggregateSchoolService.new(@school).aggregate_school
```
jamesjefferies: This should be passed in again, rather than instantiating it here

PR #450, opened 2019-01-15
New method (with test) to work out the last common date for a supply.

jamesjefferies:
Alerts can use this, else you end up with the aggregated data being
earlier than the alert date


app/views/schools/alert_reports/index.html.erb
```diff
@@ -64,12 +64,11 @@
   <div class="row padded-row">
```
jamesjefferies: Reword the message when an alert hasn't run as it looks like it's giving the reason why the alert hasn't run rather than which alert it is...

PR #453, opened 2019-01-17
NOT USING - Prototype downloading

jamesjefferies:
This is working, thinking about tests and where the to_csv and to_xls functionality could/should live


jamesalmond:
> 👍 looks good

If we're happy with the performance, perhaps it's enough to prove that we're working on downloading data.

jamesjefferies:
Yes, happy - but would like to look at some native Postgres CSV creation as well as that would be useful for other usecases too..

PR #456, opened 2019-01-18
Calendar event validation

jamesalmond:


app/services/holiday_factory.rb
```diff
@@ -17,7 +17,51 @@ def create
...
       holiday_end_date = next_term.start_date - 1.day
 
-      CalendarEvent.where(calendar: @calendar, calendar_event_type: holiday_type, start_date: holiday_start_date, end_date: holiday_end_date).first_or_create
+      if holiday_end_date >= holiday_start_date
+        CalendarEvent.where(calendar: @calendar, calendar_event_type: holiday_type, start_date: holiday_start_date, end_date: holiday_end_date).first_or_create!
+      end
     end
   end
+
+  def with_neighbour_updates(calendar_event, attributes)
```
jamesalmond: Unsurprisingly, Code Climate doesn't like this method....

jamesalmond: I'm not sure I do either

app/models/calendar_event.rb
```diff
@@ -66,23 +65,18 @@ def any_overlapping_here?(calendar_event, child_calendar, date_to_check)
...
-  def update_neighbour_end
-    if calendar_event_type.term_time
-      previous_holiday = calendar.holidays.find_by(end_date: start_date_before_last_save - 1.day)
-      previous_holiday.update(end_date: start_date - 1.day) if previous_holiday.present?
-    elsif calendar_event_type.holiday
-      previous_term = calendar.terms.find_by(end_date: start_date_before_last_save - 1.day)
-      previous_term.update(end_date: start_date - 1.day) if previous_term.present?
+  def no_overlaps
+    if (start_date && end_date && calendar_event_type) && (calendar_event_type.term_time || calendar_event_type.holiday)
+      holiday_or_term_events = calendar.calendar_events.joins(:calendar_event_type).where(calendar_event_types: { id: (CalendarEventType.holiday + CalendarEventType.term) })
```
jamesjefferies: Do you think some of these should or could be scopified?

app/services/holiday_factory.rb
jamesjefferies: Code Climate never is going to like how *similar* this code is I guess, but maybe extracting some methods may help?

jamesalmond: I briefly thought about what I could do. The trouble is the code is structurally similar but one almost does the opposite of the other (`start_date` v `end_date`, `>` vs `<` etc). I'll have another look at it and see where I get.

app/models/calendar_event.rb
jamesalmond: Yes, I can look at making those a scope.

app/services/holiday_factory.rb
jamesalmond: I've pushed some changes that reduce the complexity via private methods. I'd be hard pushed to get it much further down.

PR #458, opened 2019-01-18
Persisting alert results - Part 1

jamesjefferies:


app/controllers/schools/alert_reports_controller.rb
```diff
@@ -5,11 +5,11 @@ class Schools::AlertReportsController < ApplicationController
 
   def index
     authorize! :read, AlertType
+    @alerts = @school.alerts
```
jamesjefferies: This needs to change - looking at frequency and also the *latest* rather than all

app/models/alert.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def detail
+    data['detail']
+  end
+
+  def help_url
+    data['help_url']
+  end
+
+  def type
```
jamesjefferies: Type - don't forget reserved words - probably in analytics code.

app/services/alerts/framework_adapter.rb
```diff
@@ -0,0 +1,51 @@
...
+        data: data_hash(analysis_report),
+      )
+    end
+
+    def data_hash(analysis_report)
+      {
+        help_url: analysis_report.help_url,
+        detail: analysis_report.detail,
+        rating: analysis_report.rating,
+        type: analysis_report.type,
```
jamesjefferies: Rename this

app/services/alerts/framework_adapter.rb
```diff
@@ -0,0 +1,51 @@
...
+        term: analysis_report.term
```
jamesjefferies: Frequency? Rename things to make sense in our context rather than relying on analytics terminology

app/services/alerts/framework_adapter.rb
```diff
@@ -0,0 +1,51 @@
...
+    def initialize(alert_type, school, analysis_date, aggregate_school = AggregateSchoolService.new(school).aggregate_school)
+      @alert_type = alert_type
+      @school = school
+      @aggregate_school = aggregate_school
+      @analysis_date = analysis_date
+    end
+
+    def analyse
+      begin
+        analysis_report = alert_instance.new(@aggregate_school).analyse(@analysis_date)
```
jamesjefferies: Test this with a mock to return an Alert Report or No Method Error or whatever?

app/controllers/schools/alert_reports_controller.rb
jamesjefferies: If we don't upsert, we need the latest ones - that's ok

jamesjefferies: Also need to use the date here!

lib/tasks/alerts/create_holiday_alerts.rake
```diff
@@ -0,0 +1,15 @@
+namespace :alerts do
+  desc 'Run alerts job'
+  task create_holiday: [:environment] do
+    puts Time.zone.now
+    schools = School.active
+    schools.each do |school|
+      puts "Running alerts for #{school.name}"
+
```
jamesjefferies: Use school.calendar.holiday_approaching - this needs to match analytics code for now

jamesjefferies: More than a single day too, keep trying and the alert subscription manager can insure we don't do duplicates

PR #470, opened 2019-01-31
Alert subscriptions - Part 1 - no actual sending or SMSing unless admin user clicks button

jamesjefferies:


app/services/alerts/generate_subscription_events.rb
```diff
@@ -0,0 +1,47 @@
...
+    def perform
+      @school.alerts.latest.each do |alert|
+        process_alert(alert)
+      end
+    end
+
+  private
+
+    def process_alert(alert)
+      if any_subscriptions?(@school, alert.alert_type)
```
jamesalmond: Not sure this guard is necessary. If there are no subscriptions the `each` block in process subscriptions won't run which is effectively the same thing?

app/views/alert_mailer/alert_email.html.erb
```diff
@@ -1,27 +1,27 @@
 <h1>Energy Sparks alerts for <%= link_to @school.name, school_url(@school) %></h1>
 <p>You have been assigned as a contact for the following alerts from Energy Sparks.</p>
 <% if @school.school_admin.any? %>
```
jamesalmond: Could this just be a check for some unsubscribe emails? This way there's logic for them being from the school admin in two places.

app/services/alerts/generate_subscription_events.rb
jamesjefferies: Yep - good spot

app/views/alert_mailer/alert_email.html.erb
jamesjefferies: Yep, also good spot - best to keep school.school_admin out of it..

PR #471, opened 2019-02-13
Persisting alert results new analytics not on dashboard

jamesjefferies:


app/models/alert.rb
```diff
@@ -0,0 +1,62 @@
...
+  scope :gas,                 -> { joins(:alert_type).merge(AlertType.gas) }
+  scope :no_fuel,             -> { joins(:alert_type).merge(AlertType.no_fuel) }
+  scope :termly,              -> { joins(:alert_type).merge(AlertType.termly) }
+  scope :weekly,              -> { joins(:alert_type).merge(AlertType.weekly) }
+  scope :before_each_holiday, -> { joins(:alert_type).merge(AlertType.before_each_holiday) }
+  scope :usable,              -> { where(status: [:good, :poor])}
+
+  enum status: [:good, :poor, :not_enough_data, :failed]
+
+  scope :latest, -> { order(created_at: :desc).group_by { |alert| [alert.alert_type_id] }.values.map(&:first) }
```
jamesalmond: This might get inefficient when we have lots of alerts. Could we group in the DB?

spec/services/alerts/framework_adapter_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'rails_helper'
+
+GAS_FUEL_ALERT_TYPE = AlertType.create(fuel_type: :gas, frequency: :weekly, class_name: 'Alerts::DummyAlertClass', description: 'Woof', title: 'Broken')
```
jamesalmond: Can we scope this constant better? It would be available to all tests where it is now.

PR #484, opened 2019-02-21
Lambda ruby

jamesalmond:


jamesalmond:
@jamesjefferies I don't think we abosulutely need to keep the versions in sync. The only problems that may occur is if you're running a different version of Ruby locally to the version running in lambda. I've not locked the version of ruby in any way (e.g. it could be done with bundler or .ruby-version) but it's something we could add.

PR #493, opened 2019-03-01
Set lambda region to eu-west-2

jamesalmond:


jamesalmond:
@jamesjefferies I was hoping it would pick up the region from the credentials file but it doesn't. If you set and ENV variable it in your local environment then it would pick that up but I didn't know whether we could rely on that being set so I decided to hardcode it.

PR #495, opened 2019-03-01
Asset CDN

jamesalmond:


jamesalmond:
@jamesjefferies I've moved the favicon images to the asset pipeline so they can be called using `image_path` and therefore pulled from the CDN. This means they wouldn't be in the root `/public` folder any more and would have the hash suffix meaning we couldn't know the name of them to hardcode a link to them. I've removed them from the error pages for now so they don't get called at all. What do you think? Is it worth trying to render more dynamic 500/404 pages? In some ways I'd prefer them to be simpler but we've obviously lost the favicon from the error pages too...

app/views/shared/_favicon.html.erb
```diff
@@ -1,20 +1,20 @@
...
-<link rel="apple-touch-icon-precomposed" sizes="60x60" href="/public/favicon/apple-touch-icon-60x60.png" />
-<link rel="apple-touch-icon-precomposed" sizes="120x120" href="/public/favicon/apple-touch-icon-120x120.png" />
-<link rel="apple-touch-icon-precomposed" sizes="76x76" href="/public/favicon/apple-touch-icon-76x76.png" />
-<link rel="apple-touch-icon-precomposed" sizes="152x152" href="/public/favicon/apple-touch-icon-152x152.png" />
-<link rel="icon" type="image/png" href="/public/favicon/favicon-196x196.png" sizes="196x196" />
-<link rel="icon" type="image/png" href="/public/favicon/favicon-96x96.png" sizes="96x96" />
-<link rel="icon" type="image/png" href="/public/favicon/favicon-32x32.png" sizes="32x32" />
-<link rel="icon" type="image/png" href="/public/favicon/favicon-16x16.png" sizes="16x16" />
-<link rel="icon" type="image/png" href="/public/favicon/favicon-128.png" sizes="128x128" />
+<link rel="apple-touch-icon-precomposed" sizes="57x57" href="<%=image_path('favicon/apple-touch-icon-57x57.png')%>" />
```
jamesjefferies: You can also use https://apidock.com/rails/ActionView/Helpers/AssetTagHelper/favicon_link_tag - with asset path - but if this all works, then *great* !

config/environments/production.rb
```diff
@@ -37,7 +37,7 @@
   # `config.assets.precompile` and `config.assets.version` have moved to config/initializers/assets.rb
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  # config.action_controller.asset_host = 'http://assets.example.com'
+  config.action_controller.asset_host = ENV['ASSET_HOST'] if ENV['ASSET_HOST'].present?
```
jamesjefferies: Yep, and have two cdn subdomains set up for test & production?

jamesjefferies:
> @jamesjefferies I've moved the favicon images to the asset pipeline so they can be called using `image_path` and therefore pulled from the CDN. This means they wouldn't be in the root `/public` folder any more and would have the hash suffix meaning we couldn't know the name of them to hardcode a link to them. I've removed them from the error pages for now so they don't get called at all. What do you think? Is it worth trying to render more dynamic 500/404 pages? In some ways I'd prefer them to be simpler but we've obviously lost the favicon from the error pages too...

I think we keep a single favicon.ico in the public folder and have that referenced by the error pages? It's the classic thing about keeping 500/404 simple yet them requiring css etc..

PR #503, opened 2019-03-07
Routes templates for pupils and teachers

jamesjefferies:


config/routes.rb
```diff
@@ -113,4 +113,13 @@
     end
   end
 
+  namespace :teachers do
+    get 'schools/:school_id/chart', to: 'charts#show'
+    get 'schools/:id', to: 'schools#show', as: :school
```
jamesalmond: Why not resource routes here?

app/controllers/pupils/schools_controller.rb
```diff
@@ -0,0 +1,2 @@
+class Pupils::SchoolsController < SchoolsController
```
jamesalmond: I think this pattern is OK for now but I suspect we'll come across a a situation where we might want to inherit from a common `PupilsController` at some point and we'll have to decide what to do.

config/routes.rb
```diff
@@ -113,4 +113,13 @@
     end
   end
 
+  namespace :teachers do
+    get 'schools/:school_id/chart', to: 'charts#show'
```
jamesalmond: I'm not sure what these routes are for? Do they need to be namespaced?

config/routes.rb
jamesjefferies: Yep, now resource routes

app/controllers/pupils/schools_controller.rb
jamesjefferies: Yep, fair enough

config/routes.rb
jamesjefferies: They did, but with fresh eyes, now they don't.

PR #525, opened 2019-03-19
Radio button wasn't selected by default kwh

jamesjefferies:


spec/system/analysis_spec.rb
```diff
@@ -5,8 +5,46 @@
...
+      visit school_analysis_path(school)
+      expect(page).to have_content('Currently your measurements are in energy used in kilowatt-hours')
+      click_on('Change energy usage units', match: :first)
+      Measurements::MEASUREMENT_OPTIONS.values.each { |energy_description| expect(page.has_content?(energy_description.capitalize)).to be true }
+      expect(find('#measurement_kwh').selected?).to be true
+      expect(find('#measurement_co2').selected?).to be false
+    end
+  end
+
+  def stub_out_the_aggregation_etc
```
jamesjefferies: Pretty yucky that we have to do this - shows that my implementation of the analysis stuff needs to know too much!

PR #530, opened 2019-03-21
Alert reports and email previews

jamesjefferies:


app/services/alerts/generate_email_notifications.rb
```diff
@@ -4,9 +4,12 @@ def perform
       Contact.all.each do |contact|
         events = contact.alert_subscription_events.where(status: :pending, communication_type: :email)
         if events.any?
+          message_id = SecureRandom.uuid
           alerts = events.map(&:alert)
-          AlertMailer.with(email_address: contact.email_address, alerts: alerts, school: contact.school).alert_email.deliver_now
-          events.update_all(status: :sent)
+          alert_ids_as_string_parameter = alerts.pluck(:id).to_s
```
jamesalmond: Not sure I follow why these need to be a string?

app/models/alert_subscription_event.rb
```diff
@@ -9,6 +9,7 @@
 #  created_at            :datetime         not null
 #  id                    :bigint(8)        not null, primary key
 #  message               :text
+#  message_id            :uuid
```
jamesjefferies: Consider splitting this out in to a separate object, like the find out more calculation object

app/services/alerts/generate_email_notifications.rb
jamesjefferies: Yep, will sort this as discussed

PR #531, opened 2019-03-21
Let's get that lightbulb in

jamesjefferies:


app/controllers/pupils/schools_controller.rb
```diff
@@ -1,9 +1,42 @@
...
+
     @latest_alerts_sample = @school.alerts.usable.latest.sample(2)
     @scoreboard = @school.scoreboard
     if @scoreboard
       @surrounding_schools = @scoreboard.surrounding_schools(@school)
     end
+    # Temporary until equivalences are set up
+
+    if @school.meters?(:electricity)
+      electricity_card = MeterCard.create(school: @school, supply: :electricity)
```
jamesjefferies: Rather than creating the card, just call the class method:   def calulate_average_usage(school, supply, window)
And whilst we're at it, change them to be named params

PR #541, opened 2019-03-26
Storage heaters solar pv changes

jamesjefferies:


jamesjefferies:
I've now amended this so that it disables Solar PV and Storage Heaters at our end with an environment variable. This is currently set on production, but it means that we can merge and deploy this PR and keeps things tidier than having long running things on test...

spec/models/meter_spec.rb
```diff
@@ -1,6 +1,33 @@
...
 describe 'Meter', :meters do
+  describe 'meter attributes' do
+    let(:meter) { build(:electricity_meter) }
+
+    it 'can get them all' do
+      meter =  build(:electricity_meter)
+      expect(meter.meter_attributes(DummyMeterAttributes)).to eq DummyMeterAttributes::ATTRIBUTES
+    end
+
+    it 'drops storage heaters if env set' do
```
jamesjefferies: Change method call to receive ENV as a parameter with a default of env and use env.fetch

PR #543, opened 2019-03-26
Simpler format for pupil equivalences

jamesalmond:


app/controllers/pupils/schools_controller.rb
```diff
@@ -1,4 +1,6 @@
 class Pupils::SchoolsController < SchoolsController
+  include ActionView::Helpers::NumberHelper
```
jamesjefferies: It needs this because we're in a controller right?

jamesalmond: Yes, only available in views by default.

jamesalmond: If we did more work on the equivalences it might be worth moving the text creation somewhere else.

PR #544, opened 2019-03-27
Redirect old dashboard to new teacher dashboard

jamesalmond:


jamesjefferies:
Also, shall we remove the Teacher Dashboard link too for now?

PR #546, opened 2019-03-27
Find out more chart units

jamesalmond:


app/assets/javascripts/common_chart_options.js
```diff
@@ -185,6 +185,12 @@ function pie(d, c, chartIndex, seriesData, $chartDiv) {
   var chartHeight = $chartDiv.height();
   var yAxisLabel = d.y_axis_label;
 
+  if(yAxisLabel == '£'){
+    var pointFormat = yAxisLabel + '{point.y:.2f}';
+  } else {
+    var pointFormat = '{point.y:.2f} ' + yAxisLabel;
```
jamesalmond: @jamesjefferies do you think I should apply this change everywhere we use the `'{point.y:.2f} ' + yAxisLabel` pointFormat? I'm getting a bit lost in the chart configurations as to which ones are for specific unit types.

PR #548, opened 2019-03-28
Changes to cron jobs daily alerts etc

jamesjefferies:
This PR does the following

- Removes old weekly, termly, holiday alert cron jobs
- Removes ancient alert cron job
- Adds new daily alert job
- Moves find out more time to be after daily alerts
- Removes all extra validation of meter readings
- Adds new termly and weekly alert subscription events. Note these will only create one event per alert so you shouldn't get the same information twice. These both run weekly, but the termly one checks for approaching holidays
- Simplify services accordingly


.ebextensions/cronjob.config
```diff
@@ -15,16 +15,6 @@ files:
         content: |
             30 3,15 * * * root /bin/bash -l -c 'cd /var/app/current && RAILS_ENV=production bundle exec rake data_feeds:solar_pv_tuos_loader >> log/jobs.log 2>&1'
 
-    #
-    # Run alert cron job
-    #
-    "/etc/cron.d/run-alert-job":
```
jamesjefferies: Note this is the *original* one, so should have been deleted ages ago

.ebextensions/cronjob.config
```diff
@@ -95,54 +73,58 @@ files:
...
-            15 8 * * * root /bin/bash -l -c 'cd /var/app/current && RAILS_ENV=production bundle exec rake alerts:create_alert_subscription_events >> log/jobs.log 2>&1'
+            45 6 * * Wed root /bin/bash -l -c 'cd /var/app/current && RAILS_ENV=production bundle exec rake alerts:create_alert_subscription_events_weekly >> log/jobs.log 2>&1'
+
     #
-    # Find Out More
+    # Alert subscription events termly
+    # Runs weekly, but only generates if a holiday is approaching
     #
-    "/etc/cron.d/find-out-mores":
+    "/etc/cron.d/alert-subscription-events":
```
jamesalmond: Same filename as the one above?

.ebextensions/cronjob.config
```diff
@@ -95,54 +73,58 @@ files:
...
-        owner: root
-        group: root
-        content: |
-            0 8 * * Wed root /bin/bash -l -c 'cd /var/app/current && RAILS_ENV=production bundle exec rake alerts:create_weekly >> log/jobs.log 2>&1'
-    #
-    # Holiday alerts
+    # Generate alerts daily
     #
-    "/etc/cron.d/holiday-alerts":
+    "/etc/cron.d/daily-alert-generation":
```
jamesalmond: I think a pre-requisite for this task should be clearing of the cache. Or, we should do it after validation. If the following tasks (i.e. find out mores) use a cached aggregated school that doesn't match the current live validated data then we might be out of sync again.

lib/tasks/alerts/create_alert_subscription_events.rake
```diff
@@ -1,11 +1,31 @@
...
+      end
+    end
+    puts Time.zone.now
+  end
+
+  task create_alert_subscription_events_termly: [:environment] do
+    puts Time.zone.now
+    schools = School.active
+    schools.each do |school|
+      if school.holiday_approaching?
```
jamesalmond: To confirm, we're relying on this doing a 7 day check for an approaching holiday and we're running this cron task every 7 days so it should only get run once per "holiday approaching" and therefore we get termly emails for whatever the "latest" alert was (and those alerts may be ones we previously thought as "termly" but get run every day now)

PR #551, opened 2019-03-29
Set alert colours in the editor

jamesalmond:


app/services/template_interpolation.rb
```diff
@@ -1,18 +1,23 @@
...
+    pre_defined = @proxy.inject(@with_objects) do |collection, field|
+      collection[field] = @object.send(field)
+      collection
+    end
+    templated = fields.inject(pre_defined) do |collection, field|
       collection[field] = process(@object.send(field), with)
       collection
     end
-    OpenStruct.new(templated)
+    ClosedStruct.new(templated)
```
jamesalmond: Using an OpenStruct allows you to call _any_ method on the resulting class without error and it returns `nil`. ClosedStruct will do the usual method missing behaviour.

If we find manually defining the proxy variables up-front too cumbersome then we can look at some kind of decorator object with templating overrides if we need to.

PR #560, opened 2019-04-15
Alert updates

jamesalmond:


app/models/alert_type_rating_content_version.rb
```diff
@@ -0,0 +1,53 @@
...
+    presence: true,
+    if: ->(content) { content.alert_type_rating && content.alert_type_rating.find_out_more_active?},
+    on: :create
+  validates :sms_content,
+    presence: true,
+    if: ->(content) { content.alert_type_rating && content.alert_type_rating.sms_active?},
+    on: :create
+  validates :email_title, :email_content,
+    presence: true,
+    if: ->(content) { content.alert_type_rating && content.alert_type_rating.sms_active?},
```
jamesjefferies: Should be email_active?

app/views/admin/alert_types/ratings/index.html.erb
```diff
@@ -0,0 +1,35 @@
...
+    </tr>
+  </thead>
+  <tbody>
+    <% @ratings.each do |rating| %>
+      <tr>
+        <td><%= rating.rating_from %></td>
+        <td><%= rating.rating_to %></td>
+        <td><%= rating.description %></td>
+        <td class="lead">
+          <span class="<%='text-success' if rating.sms_active? %>"><%= fa_icon('sms') %></span>
```
jamesjefferies: You might be able to ```fa_icon('sms 2x')``` for example to double the size, but don't quote me on that

PR #562, opened 2019-04-17
Add chart titles and render full size chart if there are no tables

jamesjefferies:


app/views/schools/find_out_more/show.html.erb
```diff
@@ -2,14 +2,12 @@
 
 <h1><%= @content.page_title %></h1>
 
-<div class="row justify-content-md-center align-items-center">
-  <% @charts.each_with_index do |chart, index| %>
+<% if @tables.empty? %>
+ <%= render 'charts', school: @school, charts: @charts %>
```
jamesalmond: I think you'll have to fully qualify the partial name so we can render the charts from other places like the preview

PR #568, opened 2019-04-18
ON HOLD - Carbon feed work

jamesjefferies:
- [x] Set up daily job to load Carbon data with Upsert
- [x] Set up AMR code to use Carbon data from DB
- [x] Set up new Carbon data table and model
- [x] Set up bulk loader of historic information


app/services/schedule_data_manager_service.rb
```diff
@@ -63,4 +63,16 @@ def solar_pv
...
   end
+
+  def co2
+    cache_key = "co2-feed"
+    Rails.cache.fetch(cache_key, expires_in: 3.hours) do
+      uk_grid_carbon_intensity_data = GridCarbonIntensity.new
+      DataFeeds::CarbonIntensityReading.all.pluck(:reading_date, :carbon_intensity_x48).each do |date, values|
+        uk_grid_carbon_intensity_data.add(date, values.map(&:to_f))
+      end
+      CO2Parameterised.fill_in_missing_uk_grid_carbon_intensity_data_with_parameterised(uk_grid_carbon_intensity_data)
```
jamesjefferies: Not 100% sure we need this missing data thingy - have asked PH

lib/tasks/bulk_loaders/load_historic_carbon_intensity_data.rake
```diff
@@ -0,0 +1,10 @@
+namespace :bulk_load do
+  task historic_carbon_intensity_data: [:environment] do
+    pathname = Pathname.new("tmp/uk_carbon_intensity.csv")
+    raise ArgumentError, 'File is missing, should be in "tmp/uk_carbon_intensity.csv"' unless pathname.exist?
+    CSV.foreach(pathname, col_sep: ',') do |row|
+      date = row.shift
+      DataFeeds::CarbonIntensityReading.create(reading_date: date, carbon_intensity_x48: row)
```
jamesjefferies: This uses PH's file - possible we can switch this to just use the 'API' and call it with an after party job. Checking with PH when back

app/services/schedule_data_manager_service.rb
jamesjefferies: We do need it.

PR #573, opened 2019-04-23
Allow selection of chart type from list

jamesjefferies:
Awaiting analytics changes


app/controllers/admin/alert_types/ratings_controller.rb
```diff
@@ -53,14 +54,21 @@ def content_params
...
           ]
         )
       end
 
       def set_template_variables
         @template_variables = @alert_type.cleaned_template_variables
       end
+
+      def set_available_charts
+        available_chart_variables = @alert_type.class_name.constantize::TEMPLATE_VARIABLES.select { |_key, values| values[:units] == :chart }
```
jamesjefferies: This is obvs. a bit yuck at the moment - we get hold of all of the template variables, but as the chart ones reference the variable rather than the chart config we name, we then need to:

app/controllers/admin/alert_types/ratings_controller.rb
```diff
@@ -53,14 +54,21 @@ def content_params
...
+        dummy_alert_type_class = @alert_type.class_name.constantize.new(School.first)
```
jamesjefferies: Create a dummy alert of this instance

app/controllers/admin/alert_types/ratings_controller.rb
```diff
@@ -53,14 +54,21 @@ def content_params
...
+        @available_charts = available_chart_variables.map { |alert_chart_name, description_and_units| [description_and_units[:description].capitalize, dummy_alert_type_class.send(alert_chart_name)] }
```
jamesjefferies: To send it the instance method (which is the template chart name) which returns the chart config name.

app/controllers/admin/alert_types/ratings_controller.rb
```diff
@@ -53,14 +54,21 @@ def content_params
...
+        @available_charts << ["None", :none]
```
jamesjefferies: And we add None as an option, which is the default

app/models/alert_type.rb
```diff
@@ -46,6 +46,14 @@ def cleaned_template_variables
     end
   end
 
+  def available_charts
+    constant_class = class_name.constantize
```
jamesjefferies: This is obviously yuck, but will get better when fixed in analytics. We get the template variables, create an analytics alert, then send it the variable names to get the chart config names

spec/system/admin/alert_type_management_spec.rb
```diff
@@ -51,6 +50,7 @@
       visit root_path
       click_on 'Manage'
       click_on 'Alert Types'
+      allow_any_instance_of(AlertType).to receive(:available_charts).and_return([['chart description A', :chart_a], ['chart description B', :chart_b]])
```
jamesjefferies: Always a bit reluctant to add these, but happy to consider other alternatives...

app/models/alert_type.rb
jamesalmond: I think we need to revisit how the charts are referenced in the analytics. I think the `variable name` -> `chart name` translation should be available on the class of the alert and done when we want to save the alert or render the chart. If we do the translation at the point we display and save the content then it's effectively frozen at that point and any change to the analytics (e.g. `one_year_baseload_chart` references `:alert_1_year_baseload_with_carbon` instead of `:alert_1_year_baseload`) won't be reflected until we revisit the content again and `:alert_1_year_baseload` will no longer be an option in the list to select.

It feels like this way round we're saving the value of one of the variables in the content, not the name of the variable.

spec/system/admin/alert_type_management_spec.rb
jamesalmond: The alert type factory should create all alerts with a dummy/test alert class that conforms to all the interfaces we want to use in the tests.

jamesalmond: Possibly easier to think about when we have "System" alerts that we own in our code.

app/models/alert_type.rb
jamesalmond: Actually, we get the variable names in chart data now anyway, so I think we can just persist the variable name here and cross reference it when we pull the chart out.

app/controllers/schools/find_out_more_controller.rb
```diff
@@ -10,8 +10,14 @@ def show
...
       @content = TemplateInterpolation.new(@find_out_more.content_version).interpolate(:page_title, :page_content, with: @alert.template_variables)
-      @charts = @alert.charts
+      @chart = chart(@find_out_more.content_version.chart_type)
       @tables = @alert.tables
     end
+
+  private
+
+    def chart(chart_type)
+      chart_type == 'none' || chart_type.nil? ? nil : chart_type
```
jamesalmond: As per https://github.com/BathHacked/energy-sparks/pull/573/files#r278846110 we should be looking at using the variable name stored in the content to cross-reference against the alert's chart_data to get the chart name, like we do with template variables.

app/models/alert_type.rb
jamesalmond: So, ignore the bit about making it available on the class

PR #576, opened 2019-04-24
Holiday alert in app

jamesalmond:


jamesalmond:
@jamesjefferies I've made a start on the infrastructure for this. See what you think.

My aims were to be able to be able to swap out different alert providers, have consistent interfaces and outputs for each adapter and hide the gnarly bits for each alert source withing the adapter.

Here's a pseudo(ish)-code system alert for upcoming holidays. I've not had time to spec and properly write so it's just an example for now.
```
module Alerts
  module System
    class UpcomingHoliday
      def initialize(school:)
        @school = school
      end

      def report
        if @school.holiday_approaching?
          Adapters::Report.new(
            status: :good,
            rating: 2.0,
            summary: 'The school has a holiday approaching in their calendar',
            detail: '',
            template_data: {
              holiday_start_date: @school.next_holiday.start_date
            }
          )
        else
          Adapters::Report.new(
            status: :good,
            rating: 10.0,
            summary: 'The school has no holiday approaching',
            detail: ''
          )
        end
      end
    end
  end
end
```

app/services/alerts/adapters/report.rb
```diff
@@ -0,0 +1,17 @@
+module Alerts
+  module Adapters
+    class Report
```
jamesalmond: There has got to be a simpler way of defining classed like this but I'm not sure of what direction to take. Any preferences?

app/views/calendars/calendar_events/new.html.erb
```diff
@@ -1,3 +1,3 @@
 <h1>Create a new event</h1>
 
-<%= render 'calendars/calendar_events/form', calendar: @calendar, calendar_event: CalendarEvent.new %>
```
jamesalmond: This meant we never got error messages for any calendar events

app/services/alerts/adapters/report.rb
jamesjefferies: You mean without the module module?

app/views/calendars/calendar_events/new.html.erb
jamesjefferies: whoops!

app/services/alerts/adapters/report.rb
jamesalmond: No, I prefer module module over the rails autoloady magic way of doing modules.

I was thinking of an easier way to define boilerplate PORO-ish classes. Something activemodel based?

jamesjefferies: Ah yes, maybe, I always forget about what you get with ActiveModel when you use it without AR

jamesalmond: Unfortunately, you don't get value defaults with ActiveModel

PR #577, opened 2019-04-25
Rails6 rc1 - deploy first stand alone without ActionText

jamesjefferies:


jamesalmond:
@jamesjefferies Could we look at squashing this down to a single commit?

PR #584, opened 2019-05-01
New calendar area

jamesalmond:


app/controllers/admin/calendar_areas_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    def edit
+    end
+
+    def update
+      if @calendar_area.update(calendar_area_params)
+        redirect_to admin_calendar_areas_path, notice: 'Calendar area updated'
+      else
+        render :edit
```
jamesjefferies: it's always the edits which are lonely and untested ;)

lib/energy_sparks/csv_loader.rb
```diff
@@ -0,0 +1,25 @@
+require 'csv'
+module EnergySparks
+  class CsvLoader
+    def self.default_options
```
jamesjefferies: Now we have set this up, at some point I guess we have some other CSV parsers which we should switch to using this....

app/controllers/admin/calendar_areas_controller.rb
jamesalmond: I'll add a little tweak to the test to make sure it's not all red and embarassed

lib/energy_sparks/csv_loader.rb
jamesalmond: Yes, as we come across them I thought we could move them across.

PR #585, opened 2019-05-02
Amc loading page

jamesjefferies:


jamesalmond:
@jamesjefferies Looks like a good solution given the constraints we have. Couple of thoughts:

- it's quite a common user journey so I think we could do with some tests. Preferable to not have to have JS turned on all the time so maybe we could just hit the cache controller manually, or a _controller test_.....
- It doesn't play that well with the back button. Not a deal breaker and I'm not sure about what a fix might be for that at the moment.

app/controllers/schools/aggregated_meter_collections_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+      @number_of_solar_pv_readings = DataFeedReading.where(data_feed: data_feed).count
+    end
+
+    def post
+      # JSON request to load cache
+      ass = AggregateSchoolService.new(@school)
+      ass.aggregate_school unless ass.in_cache?
+
+      respond_to do |format|
+        format.json { render json: { referrer: session[:aggregated_meter_collection_referrer] }}
```
jamesjefferies: Need to change this in case the referrer is not set, in which case we'd go to the school home page

app/controllers/teachers/schools_controller.rb
```diff
@@ -7,9 +7,15 @@ class SchoolsController < ApplicationController
 
     def show
       redirect_to enrol_path unless @school.active? || (current_user && current_user.manages_school?(@school.id))
-      @charts = [:teachers_landing_page_electricity, :teachers_landing_page_gas]
-      setup_find_out_more
-      setup_activity_suggestions
+
+      if AggregateSchoolService.new(@school).in_cache? || ! Rails.application.config.action_controller.perform_caching
```
jamesjefferies: I don't like asking whether caching is switched on, but of courses tests, don't have it switched on...

app/controllers/schools/aggregated_meter_collections_controller.rb
```diff
@@ -0,0 +1,30 @@
+module Schools
+  class AggregatedMeterCollectionsController < ApplicationController
```
jamesjefferies: Need to sort out authorisation! for not logged in!

app/assets/javascripts/aggregated_meter_collection.js
```diff
@@ -0,0 +1,10 @@
+"use strict";
+
+$(document).ready(function() {
+  if ($("div.aggregated-meter-collection-loader").length) {
+    // Post to controller
+    $.post(window.location.href, function(data) {
+      window.location.href = data.referrer;
```
jamesjefferies: Also delete loading page from history so back button works

PR #586, opened 2019-05-03
Remove areas from AMR configs and import Oxford

jamesalmond:


lib/tasks/deployment/20190503121932_create_oxford_amr_config.rake
```diff
@@ -0,0 +1,24 @@
+namespace :after_party do
+  desc 'Deployment task: create_oxford_amr_config'
+  task create_oxford_amr_config: :environment do
+    puts "Running deploy task 'create_oxford_amr_config'"
+
+    reading_fields = '00:30,01:00,01:30,02:00,02:30,03:00,03:30,04:00,04:30,05:00,05:30,06:00,06:30,07:00,07:30,08:00,08:30,09:00,09:30,10:00,10:30,11:00,11:30,12:00,12:30,13:00,13:30,14:00,14:30,15:00,15:30,16:00,16:30,17:00,17:30,18:00,18:30,19:00,19:30,20:00,20:30,21:00,21:30,22:00,22:30,23:00,23:30,24:00'
```
jamesalmond: Not entirely sure whether the readings start at 00:00 or 00:30. Although, we don't have `handle_off_by_one` turned on for any configs so does it even matter? Claudia is going to investigate.

lib/tasks/deployment/20190503121932_create_oxford_amr_config.rake
```diff
@@ -0,0 +1,24 @@
...
+      description: 'Stark (Oxfordshire etc)',
+      s3_folder: 'stark',
+      s3_archive_folder: 'archive-stark',
+      local_bucket_path: 'tmp/amr_files_bucket/stark',
+      access_type: 'Email',
+      date_format: "%d/%m/%Y",
+      mpan_mprn_field: 'Meter Number',
+      reading_date_field: 'Reading Date',
+      reading_fields: reading_fields.split(','),
+      header_example: "Meter Number,Stark ID,Reading Date,#{reading_fields}"
```
jamesalmond: Not sure what the `Stark IDa field is for. Claudia looking into.

lib/tasks/deployment/20190503121932_create_oxford_amr_config.rake
```diff
@@ -0,0 +1,24 @@
+namespace :after_party do
+  desc 'Deployment task: create_oxford_amr_config'
+  task create_oxford_amr_config: :environment do
+    puts "Running deploy task 'create_oxford_amr_config'"
+
+    reading_fields = '00:30,01:00,01:30,02:00,02:30,03:00,03:30,04:00,04:30,05:00,05:30,06:00,06:30,07:00,07:30,08:00,08:30,09:00,09:30,10:00,10:30,11:00,11:30,12:00,12:30,13:00,13:30,14:00,14:30,15:00,15:30,16:00,16:30,17:00,17:30,18:00,18:30,19:00,19:30,20:00,20:30,21:00,21:30,22:00,22:30,23:00,23:30,24:00'
+    AmrDataFeedConfig.where(
+      description: 'Stark (Oxfordshire etc)',
+      s3_folder: 'stark',
```
jamesalmond: There's nothing that lets us know that it's for a particular area or school in the email so I've scoped these to the data provider. We might need to consider that when we add more schools or areas and get more data from Stark that we might have duplicate filenames.

jamesjefferies:
I’d assume it’s just their internal ID which should uniquely identify a meter in their system? Maybe?!

jamesjefferies:
Usually it is the 00:30 measuring the first 30 minutes of the day, but as we know it isn’t always the case

I thought we did have off by one turned on for, hmm, Frome?

lib/tasks/deployment/20190503121932_create_oxford_amr_config.rake
jamesjefferies: Hmm, I thought it was Frome which should have ```handle off by one``` turned on so I'm a bit confused now!

jamesjefferies: Ah, it should be on for Frome historical - no idea why it's off but we're not importing any historic data at the moment either...

lib/tasks/deployment/20190503121932_create_oxford_amr_config.rake
jamesjefferies: Yes, good point - something to think about, file format is currently: ```Transition Bath Limited TA Energy Sparks_20190716.csv```

lib/tasks/deployment/20190503121932_create_oxford_amr_config.rake
jamesjefferies: I guess it's their unique ID for a supply point? Or is it unique for a school/account?

PR #597, opened 2019-05-08
Online form for temperatures

jamesjefferies:
Early structure for recording temperatures, including Observation, Location, Temperature Recordings to, in the future just have Observations (as interventions). 

There is more to be done on this really, but as it was started a month ago, would be good to get this version sorted, then we can enhance it further.

### DIFF
--- a/app/models/temperature_recording.rb
+++ b/app/models/temperature_recording.rb
@@ -1,0 +1,31 @@
+# == Schema Information
+#
+# Table name: temperature_recordings
+#
+#  centigrade     :decimal(, )      not null
+#  created_at     :datetime         not null
+#  id             :bigint(8)        not null, primary key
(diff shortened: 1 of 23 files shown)


app/models/temperature_recording.rb
```diff
@@ -0,0 +1,31 @@
...
+#
+
+
+class TemperatureRecording < ApplicationRecord
+  belongs_to :observation
+  belongs_to :location
+
+  validates_presence_of :centigrade, :location
+
+  accepts_nested_attributes_for :location, reject_if: proc {|attributes| attributes['name'].blank? }
```
jamesalmond: The reject here is why the location disappears on invalid form submission. Could just let it be populated with blank validations and let the validation pick it up?

PR #605, opened 2019-05-10
Simplify amr import

jamesalmond:


jamesalmond:
@jamesjefferies this is the same as the Oxford branch but without the setup of the oxford config

PR #611, opened 2019-05-14
Getting a nicer way of using the date picker

jamesjefferies:
Currently you need a lot of mark up and JS to sort the date picker - this provides a Simple Form custom input and requires less JS


app/assets/javascripts/calendars.js
```diff
@@ -29,5 +29,16 @@ $(document).ready(function() {
       allowInputToggle: true,
       date: moment(calendar_event_date_value, 'DD/MM/YYYY')
     });
+
+    $('div.input-group.tempus-dominus-date').each(function() {
```
jamesjefferies: move this in to it's own JS file

app/inputs/tempus_dominus_date_input.rb
```diff
@@ -0,0 +1,36 @@
...
+      template.concat div_button
+    end
+  end
+
+  def input_html_options
+    super.merge(class: 'form-control datetimepicker-input', value: input_value, data: { target: "##{object_name}_#{attribute_name}" })
+  end
+
+  def input_value
+    @builder.object.send(attribute_name).try(:strftime, "%d/%m/%Y") || Time.zone.today.strftime("%d/%m/%Y")
```
jamesalmond: Does this mean the field always gets populated with a value? For my date pickers I need it to be blank by default as they don't have to fill it in.

jamesjefferies: This is now sorted, very good point. If you want a date setting for an empty field, set default_date to be what you want...

PR #614, opened 2019-05-15
Normalise line endings data pipeline

jamesalmond:


jamesalmond:
@jamesjefferies I've run it through the AMR import process on test and it all goes through

PR #616, opened 2019-05-20
Equivalences

jamesalmond:


Gemfile
```diff
@@ -23,7 +23,7 @@ gem 'auto_strip_attributes', '~> 2.5'
 gem 'closed_struct'
 
 # Dashboard analytics
-gem 'energy-sparks_analytics', git: 'https://github.com/PhilipTB/energy-sparks_analytics.git', tag: '0.44.0'
+gem 'energy-sparks_analytics', git: 'https://github.com/PhilipTB/energy-sparks_analytics.git', tag: 'master'
```
jamesjefferies: I usually update the lib/dashboard/version file and tag rather than running against master - just to keep track of what's what...

jamesalmond: Yes, I'll do that before I merge it into test etc.

PR #624, opened 2019-05-22
Only analyse and produce report if alert is valid

jamesjefferies:


spec/services/alerts/adapters/analytics_adapter_spec.rb
```diff
@@ -100,7 +100,9 @@ def self.alert_type(has_variables: true)
 
     context 'when the alert is not valid' do
       it 'does not return a report' do
```
jamesalmond: Spec name not updated

PR #631, opened 2019-06-04
Latest analytics

jamesjefferies:


app/assets/javascripts/common_chart_options.js
```diff
@@ -131,6 +131,9 @@ function barColumnLine(d, c, chartIndex, seriesData, chartType, noZoom) {
       } else if (y2AxisLabel == 'Degree Days') {
         axisTitle = 'Degree days';
         pointFormat = '{point.y:.2f} Degree days';
+      } else if (isAStringAndStartsWith(y2AxisLabel, 'Carbon Intensity of Electricity Grid')) {
```
jamesalmond: I wonder if we should look at a more robust solution for the axis titles at some point? Relying on the specific label name could break if the label name changes slightly.

jamesjefferies:
Superseded by 0.47.0 release

PR #639, opened 2019-06-10
Re-work around separating out schedule data manager

jamesjefferies:


app/services/amr/analytics_unvalidated_meter_collection_factory.rb
```diff
@@ -0,0 +1,61 @@
+require 'dashboard'
+
+module Amr
+  class AnalyticsUnvalidatedMeterCollectionFactory
+    def initialize(active_record_school, analytics_meter_collection_class = MeterCollection)
```
jamesjefferies: just call meter collection class

PR #640, opened 2019-06-11
Fix carbon cron job and update bath lat long

jamesjefferies:


lib/tasks/deployment/20190611094127_update_bath_lat_long.rake
```diff
@@ -0,0 +1,13 @@
+namespace :after_party do
+  desc 'Deployment task: update_bath_lat_long'
+  task update_bath_lat_long: :environment do
+    puts "Running deploy task 'update_bath_lat_long'"
+
+    # Put your task implementation HERE.
+    SolarPvTuosArea.find_by(title: "Bath").update(latitude: 51.3751, longitude: -2.36172)
```
jamesalmond: Using `update!` would let us know if this fails

PR #643, opened 2019-06-11
0.47.0 analytics,more co 2, more alert fixes and derived solar irradiance

jamesjefferies:


app/assets/javascripts/common_chart_options.js
```diff
@@ -131,6 +131,9 @@ function barColumnLine(d, c, chartIndex, seriesData, chartType, noZoom) {
       } else if (y2AxisLabel == 'Degree Days') {
         axisTitle = 'Degree days';
         pointFormat = '{point.y:.2f} Degree days';
+      } else if (isAStringAndStartsWith(y2AxisLabel, 'Carbon Intensity of Electricity Grid')) {
```
jamesjefferies: Agreed on your point before - at some point we need to refactor this stuff and make a nicer way of handling Y2Axis stuff

app/controllers/schools/analysis_controller.rb
```diff
@@ -52,6 +52,10 @@ def storage_heaters
     render_generic_chart_template
   end
 
+  def carbon_emissions
```
jamesjefferies: I wonder whether we should make this tab available for admin users only at the minute, else we will be sitting on this release without being able to push it to prod...

PR #653, opened 2019-06-14
Temperature Form: Auto complete, copy tweaks etc

jamesjefferies:
Invalid form submissions now re-display properly. Previously if you had entered a temperature but not a location, only the temperature field would be shown after you've submitted the form.

Auto complete now works for places, once you've submitted one set of temperatures, auto complete now works for the places - https://leaverou.github.io/awesomplete/

Aim copy removed

Observation spec changed to use build


jamesjefferies:
> 👍 fine with it code-wise. Still not sure about the messages overlapping the previous inputs.

I've added a card to Trello so that we can review this and won't forget about it...

PR #654, opened 2019-06-17
Equivalences carousel

jamesalmond:


app/views/pupils/schools/show.html.erb
```diff
@@ -1,53 +1,71 @@
 <% content_for :page_title do %><%= @school.name %><% end %>
 
-<% if  @equivalence %>
-  <%= render 'equivalence', equivalence_content: @equivalence_content %>
+<% if !@equivalences_content.empty? %>
```
jamesjefferies: @equivalences_content.present? would that also work here?

bin/update
```diff
@@ -12,19 +12,23 @@ FileUtils.chdir APP_ROOT do
   # This script is a way to update your development environment automatically.
   # Add necessary update steps to this file.
```
jamesjefferies: wots this? I don't think I've ever run this!

jamesalmond: I updated it because there was a yarn updated. A script that you can run after pulling changes to get up to date! Been in Rails a while, I think.

jamesalmond: I should have done it in another branch really.

app/views/pupils/schools/show.html.erb
jamesalmond: Updated

PR #655, opened 2019-06-17
Temperature results page

jamesjefferies:


app/views/schools/observations/index.html.erb
```diff
@@ -1,29 +1,59 @@
...
-      <% @observations.each do |observation| %>
-        <% observation.temperature_recordings.each_with_index do |temperature_recording, index| %>
-          <tr scope="row">
-            <td><%= nice_date_times observation.at %></td>
-            <td><%= observation.description %></td>
-            <td><%= temperature_recording.location.name %></td>
-            <td><%= temperature_recording.centigrade %></td>
-          </tr>
-        <% end %>
+      <% @observations.order(at: :desc).each do |observation| %>
```
jamesjefferies: There's a simple pivot going on here, it does the trick for now, but if you've a more elegant way of doing it, then I'm all ears @jamesalmond !

app/helpers/application_helper.rb
```diff
@@ -68,6 +68,16 @@ def class_for_alert_colour(colour)
     end
   end
 
+  def temperature_cell_colour(temperature)
+    if temperature >= 19
+      'table-danger'
+    elsif temperature < 18.0
```
jamesalmond: Does the `.0` do something different to the 19 that doesn't have one?

jamesjefferies: No, what the heck was that there for? Will tweak

PR #660, opened 2019-06-19
Add yaxis title for gas carbon and limit scale to 0.5

jamesalmond:


app/assets/javascripts/common_chart_options.js
```diff
@@ -92,6 +92,10 @@ function barColumnLine(d, c, chartIndex, seriesData, chartType, noZoom) {
   // LINE charts
   if (chartType == 'line') {
     if (y2AxisLabel !== undefined && y2AxisLabel.length) {
+
+      var axisTitle;
```
jamesjefferies: whoops, bit naughty that I'd left those out at some point, thanks for popping them back in!

PR #665, opened 2019-06-25
Chart Configuration - generate and persist

jamesjefferies:
All this does is create the model, with a rake job to populate the dashboard configuration - and at the moment, this does nothing clever, it is setting up a default config of what you would get from the analytics at the moment

The next two steps are:

1) The analysis pages uses this saved configuration rather than from the analytics
2) The saved configuration becomes more intelligent depending on the data available


app/models/schools/configuration.rb
```diff
@@ -0,0 +1,80 @@
...
+        dual_fuel_fuel_type
+      elsif meters_with_validated_readings(:electricity).exists?
+        electricity_fuel_type
+      elsif meters_with_validated_readings(:gas).exists?
+        :gas_only
+      else
+        :none
+      end
+    end
+
```
jamesjefferies: The following two methods are in school at the moment, but once we have some configurations, we can just use these instead of the school ones

app/services/schools/generate_chart_configuration.rb
```diff
@@ -0,0 +1,58 @@
...
+      Rails.logger.info "Generating chart configuration for #{@school.name} - using default values"
+      page_and_chart_config = {}
+      pages.each do |page|
+        page_and_chart_config[page.to_sym] = page_config(page)
+      end
+      @school.configuration.update(dashboard_charts: page_and_chart_config)
+    end
+
+  private
+
```
jamesjefferies: This is in here temporarily for now, doesn't get called will be set up properly once we have some configurations created.

app/models/schools/configuration.rb
```diff
@@ -0,0 +1,80 @@
...
+    # Duplicating school method temporarily
+    def has_solar_pv?
+      meters.detect(&:solar_pv?)
+    end
+
+    # Duplicating school method temporarily
+    def has_storage_heaters?
+      meters.detect(&:storage_heaters?)
+    end
+
```
jamesjefferies: This is a faff as symbols get converted to strings (we currently use symbols for all this config) - maybe we switch it further along, i.e. we get it back from the DB as strings and we convert the values before we pass on to the chart manager...

app/services/schools/generate_chart_configuration.rb
```diff
@@ -0,0 +1,57 @@
+require 'dashboard'
+
+module Schools
+  class GenerateChartConfiguration
+    def initialize(
+        school,
+        aggregated_meter_collection,
```
jamesjefferies: We could aggregate this here rather than passing it in?

.ebextensions/cronjob.config
```diff
@@ -79,6 +79,16 @@ files:
...
 
+    #
+    # Generate chart config daily
+    #
+    "/etc/cron.d/daily-chart-config-generation":
+        mode: "000644"
+        owner: root
+        group: root
+        content: |
+            0 6 * * * root /bin/bash -l -c 'cd /var/app/current && RAILS_ENV=production bundle exec rake configuration:generate_for_schools >> log/jobs.log 2>&1'
```
jamesalmond: Would this be better run just before the alerts?

app/services/schools/generate_chart_configuration.rb
jamesalmond: Prefer to pass one in as it's an expensive process. If you do it inside you can't then re-use it, or pass in a pre-calculated one.

app/services/schools/generate_chart_configuration.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+      @school.configuration.update(dashboard_charts: page_and_chart_config)
+    end
+
+  private
+
+    def _remove?(chart_type, chart_config = { y_axis_units: :kwh })
+      output = ChartData.new(@aggregated_meter_collection, chart_type, false, chart_config).data
+      output.first.series_data.nil?
+    rescue EnergySparksNotEnoughDataException
```
jamesalmond: Need more exceptions here. Possibly all the ones from the equivalences?

app/services/schools/generate_chart_configuration.rb
```diff
@@ -0,0 +1,57 @@
...
+      @dashboard_page_configuration = dashboard_page_configuration
+    end
+
+    def generate
+      Rails.logger.info "Generating chart configuration for #{@school.name} - using default values"
+      page_and_chart_config = {}
+      pages.each do |page|
+        page_and_chart_config[page.to_sym] = page_config(page)
+      end
+      @school.configuration.update(dashboard_charts: page_and_chart_config)
```
jamesalmond: Do we handle failure? Bang method?

.ebextensions/cronjob.config
jamesjefferies: Yep, I'll re-jig

lib/tasks/configuration/generate_for_schools.rake
```diff
@@ -0,0 +1,15 @@
+namespace :configuration do
+  desc 'Run alerts job'
+  task generate_for_schools: [:environment] do
+    puts Time.zone.now
+    School.all.each do |school|
+      if school.meters_with_readings.any?
```
jamesalmond: Can we let the configuration decide this for us? Feels like we're leaking logic out to the rake task. Will it always hold that we don't want to run this for schools without readings?

app/services/schools/generate_chart_configuration.rb
jamesjefferies: Yes, I'll review that properly when we implement this method - made a note in Trello

app/models/schools/configuration.rb
```diff
@@ -0,0 +1,80 @@
...
+    def dashboard_charts_symbols
```
jamesalmond: The use of dashboard is a bit confusing here, as these aren't charts for the dashboard.

.ebextensions/cronjob.config
jamesjefferies: Yes, moved it before the alerts

app/models/schools/configuration.rb
jamesjefferies: Analysis charts - rather than dashboard charts as they are what we now call the pupil and teacher dashboards...

app/services/schools/generate_chart_configuration.rb
jamesjefferies: Gives us some confidence using kwh, but in the future we might want to ensure a chart works properly, if it is a scalable one, against all possible units.

lib/tasks/configuration/generate_for_schools.rake
jamesjefferies: Handle exception in the rake task if something goes wrong in the generation, so it doesn't just stop.

app/models/schools/configuration.rb
```diff
@@ -0,0 +1,80 @@
...
+#  fk_rails_...  (school_id => schools.id) ON DELETE => cascade
+#
+
+module Schools
+  class Configuration < ApplicationRecord
+    belongs_to :school
+
+    delegate :meters, :meters_with_validated_readings, to: :school
+
+    def fuel_types_for_analysis
```
jamesjefferies: Move these in to a separate new service which returns a simple object, we can then use that for asking things like solar_pv?

PR #676, opened 2019-06-28
School configuration part 3

jamesjefferies:
Blocked until the analytics throws it's data exceptions when generating chart data


app/models/chart_data.rb
```diff
@@ -15,6 +17,15 @@ def data
     [values]
   end
 
+  def success?
+    ! data.first.series_data.nil?
+  rescue EnergySparksNotEnoughDataException
```
jamesalmond: Possible we should catch these too? https://github.com/BathHacked/energy-sparks/blob/master/app/services/equivalences/calculator.rb#L27

app/models/chart_data.rb
```diff
@@ -15,6 +17,15 @@ def data
...
+    false
+  rescue => exception
+    Rails.logger.error "Chart generation failed unexpectedly for #{chart_type} and #{@school.name} - #{exception.message}"
```
jamesalmond: Should we Rollbar these?

app/services/schools/generate_chart_configuration.rb
```diff
@@ -25,21 +25,23 @@ def generate
...
-      true
+    def white_listed_page_config(page)
+      page_configuration = page_config(page)
+      list_of_charts = page_configuration[:charts]
+      list_of_charts = list_of_charts.select { |chart| keep?(chart) }
+      { name: page_configuration[:name], charts: list_of_charts }
+    end
+
+    def keep?(chart_type, chart_config = { y_axis_units: :kwh })
+      ChartData.new(@aggregated_meter_collection, chart_type, false, chart_config).success?
```
jamesalmond: Not sure the method name `success?` is quite right. Suggests the chart process has already been run, rather than running it each time its called.

jamesjefferies: ok, I'll have a think

app/models/chart_data.rb
jamesjefferies: Yes, good point

app/models/chart_data.rb
jamesjefferies: Ah yes, we should, especially with the extra check above.

PR #680, opened 2019-07-03
Simple alert type editing and removing has_variables

jamesjefferies:


app/views/admin/alert_types/edit.html.erb
```diff
@@ -0,0 +1,30 @@
+<h1>Edit: <%= @alert_type.title %> alert type</h1>
+
+<%= simple_form_for(@alert_type, url: admin_alert_type_path(@alert_type)) do |f| %>
+
+  <%= render 'shared/errors', subject: @alert_type, subject_name: 'alert type' %>
+
+  <div class="form-group">
```
jamesalmond: Do we need these wrapper divs with simple_form?

app/views/admin/alert_types/edit.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    <%= f.input :title, as: :string, class: 'form-control' %>
```
jamesalmond: Same with these classes

app/views/admin/alert_types/edit.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  <div class="form-group">
+    <%= f.input :description, class: 'form-control' %>
+  </div>
+
+  <div class="form-group">
+    <%= f.input :analysis, class: 'form-control' %>
+  </div>
+
+  <div class="form-group">
+    <% AlertType.frequencies.keys.each do |frequency| %>
```
jamesalmond: What stops us doing something like this here? https://github.com/BathHacked/energy-sparks/blob/master/app/views/onboarding/meters/_form.html.erb#L10

app/views/admin/alert_types/show.html.erb
```diff
@@ -8,14 +8,19 @@
   <dd class="col-md-10"><%= @alert_type.frequency.humanize %></dd>
 
   <dt class="col-md-2">Show ratings</dt>
-  <dd class="col-md-10"><%= @alert_type.show_ratings %></dd>
```
jamesalmond: I don't think we use `show_ratings` for anything. Can we remove it too?

app/views/admin/alert_types/edit.html.erb
jamesjefferies: *cough* no.

app/views/admin/alert_types/edit.html.erb
jamesjefferies: Brilliant, nothing, I'd just forgotten about the label method thingy...

app/views/admin/alert_types/show.html.erb
jamesjefferies: Yes, we can

PR #681, opened 2019-07-03
Add raw download and update report page a bit

jamesjefferies:


jamesjefferies:
Good point, test added.

PR #684, opened 2019-07-09
School configuration part 4

jamesjefferies:


app/controllers/teachers/schools_controller.rb
```diff
@@ -10,13 +10,25 @@ class SchoolsController < ApplicationController
...
       setup_dashboard_alert
       setup_activity_suggestions
     end
 
   private
 
+    def setup_charts
+      @charts = {}
+
+      if @school.meters_for_supply(:electricity).any?
```
jamesalmond: Should this be in the config too? Saves us querying meters again.

app/services/schools/generate_gas_dashboard_chart_configuration.rb
```diff
@@ -0,0 +1,30 @@
...
+    def generate
+      @school.configuration.update!(gas_dashboard_chart_type: gas_dashboard_chart_type)
+    end
+
+  private
+
+    def gas_dashboard_chart_type
+      return Schools::Configuration::NO_CHART unless @fuel_configuration.has_gas
+      chart_config = { y_axis_units: :kwh }
+      working_chart = Schools::Configuration.gas_dashboard_chart_types.keys.reverse.find do |chart_type|
```
jamesalmond: might be worth moving `Schools::Configuration.gas_dashboard_chart_types.keys.reverse` out onto a separate line so it can have a variable name that explains what's going on?

app/controllers/teachers/schools_controller.rb
jamesjefferies: Yep, we can probably remove a lot of these 'meters_for_supply' and other such queries if we persist it - also checking that we have any readings rather than just a meter.

PR #686, opened 2019-07-09
Rich text equivalences

jamesalmond:


jamesalmond:
@jamesjefferies I've pushed some updates that add a form helper for supporting trix fields

PR #691, opened 2019-07-11
Remove merit

jamesalmond:


jamesjefferies:
This is hilarious - has there ever been a PR which has deleted so much code!

jamesalmond:
> This is hilarious - has there ever been a PR which has deleted so much code!

@jamesjefferies and still code climate complains about coverage!

jamesalmond:
@jamesjefferies I've gone for the simplest option of just having points against activities for now. We can split it out to something more generic if we have other things that need more points.

PR #705, opened 2019-07-19
Programmes

jamesjefferies:


app/controllers/programme_types_controller.rb
```diff
@@ -0,0 +1,21 @@
+class ProgrammeTypesController < ApplicationController
```
jamesjefferies: Actually, I think this controller is now longer required and can be deleted

lib/tasks/do_everything.rake
```diff
@@ -0,0 +1,11 @@
+namespace :do do
```
jamesjefferies: Whoops, wondered where this has gone - popped this in to test new analytics releases, obvs. can be expanded on and in the end, pass around the aggregated meter collection...

jamesjefferies:
Ah yes, meant to get rid of that instance variable in the view! Pre-holiday trying to get stuff done...

Happy for you to fix any nonsense which I committed, else to wait until my return! Could merge PR and I’ll refactor later...

jamesalmond:
@jamesjefferies I've already worked through most of them. I was mainly submitting the review for my own notes!

PR #707, opened 2019-07-20
Complete programme activity if appropriate- BASED on PROGRAMMES BRANCH

jamesjefferies:


app/services/activity_creator.rb
```diff
@@ -8,6 +8,32 @@ def process
       @activity.points = @activity.activity_type.score
       @activity.activity_category = @activity.activity_type.activity_category
     end
-    @activity.save
+
+    if @activity.save
+      process_programmes if started_active_programmes.any?
```
jamesalmond: Not sure this check is necessary given we'd be iterating through an empty list later on.

PR #731, opened 2019-08-16
Alert bars teachers pupils part 2

jamesjefferies:
Includes restyling of buttons (in general) and Pupil's dashboard - including podium

### DIFF
--- a/app/assets/stylesheets/headers.scss
+++ b/app/assets/stylesheets/headers.scss
@@ -36,3 +36,37 @@ h4,h5,h6 {
   padding-bottom: 0;
   margin: 0px;
 }
+
+.trix-content {
+  h1 {
+    font-size: 37.3px;
+    padding-top: 20px;
+    padding-bottom: 20px;
+  }
+
+  h2 {
+    font-size: 31.1px;
+  }
+
+  h3 {
+    font-size: 25.9px;
+  }
+
+  h4 {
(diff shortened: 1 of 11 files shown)


app/assets/stylesheets/links.scss
```diff
@@ -27,6 +27,18 @@ a.nav-link.active {
   outline: none;
 }
```
jamesjefferies: There are some links like the change energy usage units which needs fixing - plus work out what to do with sub nav

app/assets/stylesheets/links.scss
```diff
@@ -27,6 +27,41 @@ a.nav-link.active {
...
+.btn-outline-dark {
+  background-image: none !important;
+  border-color: $dark;
+}
+
+.btn-outline-dark:hover {
+  color: $white;
+}
+
+.btn-group-sm {
```
jamesjefferies: Don't need this as we only use this with btn-group and btn

app/assets/stylesheets/headers.scss
```diff
@@ -36,3 +36,37 @@ h4,h5,h6 {
   padding-bottom: 0;
   margin: 0px;
 }
+
+.trix-content {
```
jamesjefferies: CHange the h1 definition to be:

h1, .trix-content h1,

etc instead of this

PR #732, opened 2019-08-20
Public and school manager alerts

jamesalmond:


app/models/dashboard_alert.rb
```diff
@@ -32,5 +32,5 @@ class DashboardAlert < ApplicationRecord
   belongs_to :find_out_more
   belongs_to :content_version, class_name: 'AlertTypeRatingContentVersion', foreign_key: :alert_type_rating_content_version_id
 
-  enum dashboard: [:teacher, :pupil]
+  enum dashboard: [:teacher, :pupil, :public, :management], _suffix: :dashboard
```
jamesalmond: Without the `_suffix`, the enum tries to generate a scope called `DashboardAlert.public`, which isn't valid. Thought it would be easier to adjust the scopes for the enums than pick a new (and inconsistent) name for public.

jamesjefferies: Fair enough

PR #735, opened 2019-08-20
Fix alert failures, create enum for calendar event type

jamesjefferies:


app/services/schedule_data_manager_service.rb
```diff
@@ -14,10 +14,13 @@ def holidays(_area_name)
     @holidays ||= Rails.cache.fetch(cache_key, expires_in: 3.hours) do
       hol_data = HolidayData.new
 
-      Calendar.find(@calendar_id).holidays.order(:start_date).includes(:academic_year).map do |holiday|
-        hol_data.add(holiday.title, holiday.start_date, holiday.end_date)
-      end
+      Calendar.find(@calendar_id).not_term_time.order(:start_date).includes(:academic_year).map do |holiday|
+        # Not really being used at the moment by the analytics code
+        academic_year = nil
+        analytics_holiday = Holiday.new(holiday.analytics_event_type.to_sym, holiday.title || 'No title', holiday.start_date, holiday.end_date, academic_year)
```
jamesjefferies: Public interface - the add bit was working before, why not re-implement that so we use it?

PR #740, opened 2019-08-21
Public dashboard

jamesjefferies:


app/controllers/concerns/dashboard_alerts.rb
```diff
@@ -0,0 +1,16 @@
+module DashboardAlerts
+  extend ActiveSupport::Concern
+
+  def setup_dashboard_alerts
+    @dashboard_alerts = @school.latest_dashboard_alerts.includes(:content_version, :find_out_more).teacher_dashboard.sample(3).map do |dashboard_alert|
```
jamesjefferies: Pass scope through as parameter and school too

jamesjefferies: pass through @school.latest_dashboard_alerts.teacher_dashboard

instead of school

app/controllers/concerns/dashboard_energy_charts.rb
```diff
@@ -0,0 +1,15 @@
+module DashboardEnergyCharts
+  extend ActiveSupport::Concern
+
```
jamesjefferies: pass in school rather than instance variable

app/controllers/concerns/dashboard_timeline.rb
```diff
@@ -0,0 +1,7 @@
+module DashboardTimeline
+  extend ActiveSupport::Concern
+
+  def setup_timeline
```
jamesjefferies: pass in school

app/controllers/concerns/dashboard_alerts.rb
jamesjefferies: return values rather than set instance variables

app/controllers/schools_controller.rb
```diff
@@ -1,23 +1,28 @@
...
     @scoreboards = Scoreboard.includes(:schools).where.not(schools: { id: nil }).order(:name)
     @ungrouped_active_schools = School.active.without_group.order(:name)
     @schools_not_active = School.inactive.order(:name)
   end
 
   # GET /schools/1
-  # GET /schools/1.json
   def show
-    redirect_to teachers_school_path(@school), status: :found
+    redirect_to teachers_school_path(@school), status: :found if current_user && current_user.manages_school?(@school.id)
```
jamesjefferies: For now, redirect if current user has a school and it's this school

jamesjefferies: Also, redirect needs to be at the end, so use the if clause to separate out...

app/views/schools/show.html.erb
```diff
@@ -0,0 +1,25 @@
+<%= content_for :page_title, @school.name %>
+
+<% if ! @school.active? %>
```
jamesjefferies: Remove activate stuff from public dashboard

spec/system/teachers/programme_spec.rb
```diff
@@ -15,6 +15,7 @@
   before do
     sign_in(user)
     visit root_path
+    puts path
```
jamesjefferies: REMOVE JEFFERIES

PR #743, opened 2019-08-22
Create new top level calendars and populate with BH

jamesjefferies:


app/controllers/calendars_controller.rb
```diff
@@ -1,10 +1,13 @@
+# frozen_string_literal: true
```
jamesjefferies: I've no idea where that has appeared from!

app/controllers/calendars_controller.rb
```diff
@@ -20,9 +23,7 @@ def new
 
   # GET /calendars/1/edit
   def edit
-    if @calendar.template?
-      redirect_to calendar_path(@calendar)
-    end
+    redirect_to calendar_path(@calendar) if @calendar.template?
```
jamesalmond: Will this work if it's not a return value for the action?

jamesalmond: Should this be in cancancan as an ability?

app/views/calendars/_calendars.html.erb
```diff
@@ -13,13 +12,12 @@
     <% calendars.each do |calendar| %>
       <tr>
         <td><%= calendar.title %></td>
-        <td><%= calendar.based_on.title unless calendar.based_on.nil? %></td>
-        <td><%= nice_date_times(calendar.created_at) %></td>
+        <td><%= link_to calendar.based_on.title, calendar_path(calendar.based_on) unless calendar.based_on.nil? %></td>
```
jamesalmond: Should this  check be in cancancan as an ability?

app/views/calendars/index.html.erb
```diff
@@ -1,37 +1,17 @@
 <h1>Calendars</h1>
 <p>Click on show to select a calendar. You can then edit any events on the calendar, or add new ones.</p>
-<p><b>Note:</b> - updating a template calendar does not automatically update any dependent calendars.</p>
+<p><b>Note:</b> - updating a template calendar automatically updates any dependent calendars.</p>
```
jamesalmond: We don't use the phrase "template calendar" anywhere else in the views. Change to "Area" or "Top level" (or both)?

lib/tasks/deployment/20190822095112_add_home_nation_calendars.rake
```diff
@@ -0,0 +1,26 @@
...
+    eawca = Calendar.where(calendar_area: england_and_wales_calendar_area, template: true, title: england_and_wales_calendar_area.title).first_or_create!
+    sca = Calendar.where(calendar_area: scotland_calendar_area, template: true, title: scotland_calendar_area.title).first_or_create!
+
+    Calendar.find_by(title: 'Bath and North East Somerset (BANES)').update(based_on_id: eawca.id)
+    Calendar.find_by(title: 'Frome').update(based_on_id: eawca.id)
+    Calendar.find_by(title: 'Sheffield').update(based_on_id: eawca.id)
+    Calendar.find_by(title: 'Oxfordshire').update(based_on_id: eawca.id)
+
+    highland_calendar = Calendar.find_by(title: 'Highland')
+    highland_calendar.update(based_on_id: sca.id) if highland_calendar
```
jamesalmond: Do we need to re-populate the Highland template calendar with the scotland bank holidays? I think it will have the England and Wales ones now.

jamesjefferies:
@jamesalmond made a few tweaks and added the BH refresher after party...

app/controllers/calendars_controller.rb
```diff
@@ -20,51 +23,35 @@ def new
 
   # GET /calendars/1/edit
   def edit
-    if @calendar.template?
-      redirect_to calendar_path(@calendar)
-    end
     build_terms
+    redirect_to calendar_path(@calendar) if @calendar.template?x
```
jamesjefferies: Actually, I don't think this route *actually* gets used!

jamesjefferies:
@jamesalmond have now removed some of the calendar routes, new, edit, update as I don't think we actually use them *at the moment*

lib/tasks/deployment/20190822095112_add_home_nation_calendars.rake
jamesjefferies: There's a job to sort this now

PR #745, opened 2019-08-22
Remove bank holidays and amend tests etc

jamesjefferies:
Bank Holidays come from parent calendars, BH model and table removed


app/services/calendar_factory_from_event_hash.rb
```diff
@@ -3,12 +3,14 @@ def initialize(event_hash, area, template = false)
     @event_hash = event_hash
     @area = area
     @template = template
+    @parent_calendar = parent_calendar
```
jamesjefferies: Bank holidays now come from parent calendars rather than from the BH table

app/services/calendar_factory_from_event_hash.rb
```diff
@@ -24,20 +26,19 @@ def create
 
 private
 
+  def parent_calendar
+    Calendar.find_by(calendar_area: @area, template: true) || Calendar.find_by(calendar_area: @area.parent_area, template: true)
```
jamesjefferies: THis only allows for 3 levels of calendars

spec/support/data_helpers.rb
```diff
@@ -6,6 +6,29 @@ def create_active_school(*args)
       school_creator.process_new_configuration!
     end
   end
+
+  def create_all_calendar_events
```
jamesjefferies: This method was a service in the main application, but it's never going to be used again by the main app, so moved here in to a test support file

app/services/calendar_factory_from_event_hash.rb
jamesalmond: I think this class should be clearer about what it does. We only use it in the code (rather than the specs) to create template area calendars so I think that's all it should do. That way we just need to pass in one area and get its template calendar.

app/services/calendar_factory_from_event_hash.rb
```diff
@@ -24,20 +26,19 @@ def create
...
     HolidayFactory.new(@calendar).create
   end
 
   def create_bank_holidays
-    calendar_event_type = CalendarEventType.bank_holiday.first
-    find_bank_holidays(@area).each do |bh|
-      @calendar.calendar_events.where(title: bh.title, start_date: bh.holiday_date, end_date: bh.holiday_date, calendar_event_type: calendar_event_type).first_or_create!
-    end
-  end
+    bank_holiday = CalendarEventType.bank_holiday.first
```
jamesalmond: If we do what I'm suggesting above then we can copy across all calendar events from the template calendar here. Seems odd that we've got a template but we're only copying across one type of event.

spec/support/data_helpers.rb
```diff
@@ -6,6 +6,29 @@ def create_active_school(*args)
...
+    term_colour = 'rgb(255,172,33)'
+    inset_day_colour = 'rgb(255,69,0)'
+    inset_day_out_colour = 'rgb(181, 108, 226)'
+    bank_holiday_colour = 'rgb(59,192,240)'
+    holiday_colour = 'rgb(92,184,92)'
```
jamesalmond: If it's just going to be used in the test then I don't think we need to specify these

app/services/calendar_factory_from_event_hash.rb
jamesjefferies: This can now just be, copy everything from parent calendar (Which will have the BHs already)

spec/support/data_helpers.rb
jamesjefferies: Agreeeeeeeeeeeeeeed

app/services/calendar_factory_from_event_hash.rb
jamesjefferies: I agree - I think this should be done when we tackle (re-tackle) calendar areas

PR #750, opened 2019-08-28
Fix activity types on test for non-school based logged in users

jamesjefferies:


app/views/activity_types/show.html.erb
```diff
@@ -23,7 +23,7 @@
   </div>
 </div>
 
-<% if @activity_type.active? %>
+<% if @activity_type.active? && current_user && current_user.school %>
```
jamesalmond: `can? :manage, Activity`? I'm really losing track of what we should/can be doing with cancan...

jamesjefferies: Ah, but an admin user can manage Activity, but doesn't have a school set, which is why test was blowing up... Maybe it could be turned into an activity - or, maybe an Admin *can't* create an activity, that would also work

PR #756, opened 2019-08-29
Highlands data feed

jamesjefferies:


app/services/amr/single_read_converter.rb
```diff
@@ -0,0 +1,52 @@
+module Amr
+  class SingleReadConverter
+    def initialize(single_reading_array)
+      @single_reading_array = single_reading_array
+      @results_array = []
+    end
+
+    def perform
```
jamesjefferies: There is an update to this in a later PR as I forgot to update the meter id - also I think there is probably better ways of doing this!

lib/tasks/deployment/20190829130905_add_highlands.rake
```diff
@@ -0,0 +1,25 @@
...
+  task add_highlands: :environment do
+    puts "Running deploy task 'add_highlands'"
+
+    AmrDataFeedConfig.where(
+      description: 'Highlands',
+      s3_folder: 'highlands',
+      s3_archive_folder: 'archive-highlands',
+      local_bucket_path: 'tmp/amr_files_bucket/highlands',
+      access_type: 'Email',
+      date_format: "%e %b %Y %H:%M:%S",
```
jamesjefferies: This also gets updated as the actual records we end up inserting after the conversion have a different date format..

app/services/amr/single_read_converter.rb
jamesjefferies: https://github.com/BathHacked/energy-sparks/pull/774/files

PR #759, opened 2019-09-04
Update teachers chart configuration

jamesjefferies:


app/assets/javascripts/common_chart_options.js
```diff
@@ -103,20 +103,22 @@ function teachersChartOptions(pointFormat) {
     },
     plotOptions: {
       series: {
-        pointWidth: 30, groupPadding: 0.075
+        pointWidth: 30, groupPadding: 0.050
```
jamesjefferies: this wasn't quite to C's liking - there is a further in progress change for the chart columns

PR #760, opened 2019-09-04
Fix scoreboards for current academic year

jamesjefferies:


app/models/scoreboard.rb
```diff
@@ -45,7 +45,7 @@ def active_academic_years(today: Time.zone.today)
     calendar_area.academic_years.where("date_part('year', start_date) >= ? AND start_date <= ?", FIRST_YEAR, today).order(:start_date)
   end
 
-  def scored_schools(recent_boundary: 1.month.ago, academic_year: nil)
+  def scored_schools(recent_boundary: 1.month.ago, academic_year: this_academic_year)
```
jamesjefferies: Scoreboards were updating properly, but the other places with points, like the Podium still had last year's points totals..

Hope I did the right things ..

app/models/scoreboard.rb
```diff
@@ -56,7 +56,7 @@ def scored_schools(recent_boundary: 1.month.ago, academic_year: nil)
     if academic_year
       scored.joins(
         self.class.sanitize_sql_array(
-          ['LEFT JOIN observations ON observations.school_id = schools.id AND observations.at BETWEEN ? AND ?', academic_year.start_date, academic_year.end_date]
+          ['INNER JOIN observations ON observations.school_id = schools.id AND observations.at BETWEEN ? AND ?', academic_year.start_date, academic_year.end_date]
```
jamesalmond: This will have the effect of only showing schools on the scorboard if they have points. Is that intended?

app/models/scoreboard.rb
```diff
@@ -67,4 +67,10 @@ def scored_schools(recent_boundary: 1.month.ago, academic_year: nil)
   def position(school)
     scored_schools.index(school)
   end
+
+  private
+
+  def this_academic_year
```
jamesalmond: Should probably just force this to be passed in at all times so we're always using the correct academic year.

app/models/scoreboard.rb
jamesjefferies: No, it's not intended - hmm, maybe it's worth us having a test to make sure that a school with no points appears - I'll need to remind myself why I changed it too ;)

PR #761, opened 2019-09-04
Remove unused feeds

jamesjefferies:
We're not using Weather Underground any more - we can tidy this stuff up a fair bit (and remove a lot of unused code)


db/migrate/20190904122901_drop_data_feeds.rb
```diff
@@ -0,0 +1,19 @@
...
+  def change
+    remove_column :areas, :data_feed_id, :integer
+    drop_table  :data_feed_readings
+    drop_table  :data_feeds
+  end
+end
+
+
+
+# Deleted rows
```
jamesalmond: Does this need to be in here?

PR #768, opened 2019-09-05
Add gas shower, only show analysis path if school is provided

jamesjefferies:


app/views/pupils/schools/_equivalence.html.erb
```diff
@@ -3,7 +3,9 @@
     <div class="trix-content">
       <%= sanitize equivalence_content.equivalence %>
     </div>
-    <p><%= link_to 'Find out how much energy has been used', school_analysis_path(@school), class: 'btn btn-outline-dark rounded-pill font-weight-bold' %></p>
+    <% if defined?(school) %>
```
jamesalmond: Prefer inspecting the `local_assigns` hash to `defined?` here

jamesjefferies: Ha! Never come across ```local_assigns``` before - I like it!

PR #774, opened 2019-09-05
Patch for single read missing meter ids

jamesjefferies:


app/services/amr/single_read_converter.rb
```diff
@@ -24,7 +24,7 @@ def perform
         else
           readings = Array.new(48)
           readings[reading_index] = reading
-          new_record = { reading_date: reading_day, readings: readings, mpan_mprn: single_reading[:mpan_mprn], amr_data_feed_config_id: single_reading[:amr_data_feed_config_id] }
+          new_record = { reading_date: reading_day, readings: readings, mpan_mprn: single_reading[:mpan_mprn], amr_data_feed_config_id: single_reading[:amr_data_feed_config_id], meter_id: single_reading[:meter_id] }
```
jamesalmond: Think this class might be doing too much? I wonder if it might be better just doing the collection of readings and MPANs to one row per day and not know anything about the database IDs.

jamesjefferies: Hmm yes, I know it's not quite right - we could have another step which adds the meter ids I guess?

PR #775, opened 2019-09-06
Data pipeline patch

jamesjefferies:


data-pipeline/handlers/process_file.rb
```diff
@@ -43,9 +41,21 @@ def process(key:, bucket:)
         { statusCode: 200, body: JSON.generate(response: response) }
       end
 
+      def remove_utf8_invalids(line)
+        line.encode('UTF-8', 'binary', invalid: :replace, undef: :replace, replace: '')
```
jamesalmond: Can this be combined with the other encode? I'm not entirely clear on what the binary flag does so maybe not.

jamesjefferies: I'm not sure to be honest, I'd spent too much time faffing with Ruby versions by then!

PR #781, opened 2019-09-11
Re introduce old school links

jamesjefferies:


app/views/pupils/schools/_explore_old_charts.html.erb
```diff
@@ -0,0 +1,38 @@
+<div class="row dashbards">
```
jamesalmond: Remove this class

app/views/pupils/schools/_explore_old_charts.html.erb
```diff
@@ -0,0 +1,38 @@
...
+  <% if school.meters_for_supply(:electricity).any? %>
```
jamesalmond: Use the school config

jamesalmond: Maybe we should remove all access to these methods from the school?

jamesjefferies: Yep

PR #786, opened 2019-09-12
Refactor calendars part 2

jamesjefferies:


app/controllers/admin/calendars_controller.rb
```diff
@@ -0,0 +1,37 @@
+module Admin
+  class CalendarsController < AdminController
+    load_and_authorize_resource
+
```
jamesjefferies: This has moved in to the admin namespace.. it wasn't before..

app/controllers/admin/calendars_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    def create
+      terms = EnergySparks::CsvLoader.from_text(params.fetch(:terms) {{}}.fetch(:csv) {''})
+      based_on_calendar = Calendar.find(calendar_params[:based_on_id])
+
+      @calendar = CalendarFactory.new(existing_calendar: based_on_calendar, title: calendar_params[:title], calendar_type: Calendar.calendar_types[calendar_params[:calendar_type]]).build
+
+      if @calendar.save
+        # New create terms and holidays from hash
+        CalendarTermFactory.new(@calendar, terms).create_terms
```
jamesjefferies: This should maybe be cleverer to handle duff data, at the moment, if the calendar works, that will be saved, but the terms bit could go bang

app/controllers/calendars_controller.rb
```diff
@@ -3,22 +3,7 @@
...
-  def index
-    @national_calendars = Calendar.national.includes(:schools).order(:title)
-    @regional_calendars = Calendar.regional.includes(:schools).order(:title)
-    @school_calendars = Calendar.school.includes(:schools).order('schools.name')
-  end
-
   # GET /calendars/1
-  # GET /calendars/1.json
   def show
   end
```
jamesjefferies: We had soft deletes of calendars, deleting has been disabled for now until we understand why it might need to happen, who might need to do it and what happens if we do

app/models/academic_year.rb
```diff
@@ -2,24 +2,15 @@
...
-#  index_academic_years_on_calendar_area_id  (calendar_area_id)
-#
-# Foreign Keys
-#
-#  fk_rails_...  (calendar_area_id => calendar_areas.id) ON DELETE => cascade
+#  calendar_id :integer
+#  end_date    :date
+#  id          :bigint(8)        not null, primary key
+#  start_date  :date
 #
```
jamesjefferies: Should add FK for calendar here

app/models/academic_year.rb
```diff
@@ -2,24 +2,15 @@
...
 
 class AcademicYear < ApplicationRecord
   has_many :calendar_events
-  belongs_to :calendar_area
+  belongs_to :calendar, optional: true
```
jamesjefferies: optional true can probably go now

app/models/calendar.rb
```diff
@@ -2,47 +2,42 @@
...
-#  calendar_area_id :bigint(8)        not null
-#  calendar_type    :integer
-#  created_at       :datetime         not null
-#  default          :boolean
-#  deleted          :boolean          default(FALSE)
-#  id               :bigint(8)        not null, primary key
-#  template         :boolean          default(FALSE)
-#  title            :string           not null
-#  updated_at       :datetime         not null
+#  based_on_id   :bigint(8)
```
jamesjefferies: this should be a FK

app/models/calendar.rb
```diff
@@ -70,7 +65,11 @@ def holiday_approaching?(today: Time.zone.today)
     next_after_today.present? && (next_after_today.start_date - today <= 7)
   end
 
-  def academic_year_for(date)
-    calendar_area.academic_year_for(date)
+  private
+
+  def reject_calendar_events(attributes)
+    end_date_date = Date.parse(attributes[:end_date])
+    end_date_default = end_date_date.month == 8 && end_date_date.day == 31
```
jamesjefferies: I've not revisited this at this point!

app/views/admin/calendars/new.html.erb
```diff
@@ -0,0 +1,18 @@
+<h1>New Calendar</h1>
```
jamesjefferies: This replaces the new calendar area functionality - at the moment, the link from the calendars index page specifies **regional** - which is what we are really supporting for new calendars at the moment, but logic is there if required - we could disable completely if we wanted, i.e. removing the input radio buttons

app/views/calendars/show.html.erb
```diff
@@ -1,5 +1,5 @@
 <h1>Calendar: <%= @calendar.title %></h1>
-<p><%= link_to 'All calendars', calendars_path if can? :index, Calendar %></p>
+<p><%= link_to 'All calendars', admin_calendars_path if can? :index, Calendar %></p>
```
jamesjefferies: Only admins can see all calendars

app/controllers/schools/timeline_controller.rb
```diff
@@ -5,11 +5,11 @@ def show
       @academic_year = if params[:academic_year]
                          AcademicYear.find(params[:academic_year])
                        else
-                         @school.calendar_area.academic_year_for(Time.zone.today)
+                         @school.academic_year_for(Time.zone.today)
                        end
       first_observation = @school.observations.order('at ASC').first
       @active_academic_years = if first_observation
-                                 @school.calendar_area.parent_area.academic_years.where('end_date >= ? AND start_date <= ?', first_observation.at, Time.zone.today)
+                                 @school.calendar.based_on.academic_years.where('end_date >= ? AND start_date <= ?', first_observation.at, Time.zone.today)
```
jamesalmond: Are academic years at regional level now? Or at national level? This look like it would be doing the lookup on the regional level but the old implementation did another hop up to the calendar's area's parent area.

app/models/ability.rb
```diff
@@ -40,7 +40,7 @@ def initialize(user)
       else
         school_scope = { id: user.school_id }
         related_school_scope = { school_id: user.school_id }
-        can :crud, Calendar, id: user.school.try(:calendar_id)
+        can [:update, :show], Calendar, id: user.school.try(:calendar_id)
```
jamesalmond: Can this be the same way round as the other one? Looks like there's a difference but there isn't.

app/models/calendar.rb
jamesalmond: Hah, we should leave it in as a weird artefact...

app/models/school.rb
```diff
@@ -80,7 +79,7 @@ class School < ApplicationRecord
 
   belongs_to :calendar, optional: true
   belongs_to :template_calendar, optional: true, class_name: 'Calendar'
```
jamesalmond: Are we planning to remove this now? Or later?

app/views/admin/calendars/new.html.erb
```diff
@@ -0,0 +1,18 @@
...
+<%= simple_form_for @calendar, url: admin_calendars_path(@calendar)  do |form| %>
+
+  <% if @calendar.calendar_type %>
+    <p>Regional Calendar</p>
+    <%= form.hidden_field :calendar_type, value: :regional %>
```
jamesalmond: Can we just use the value that's set here?

jamesalmond: Rather than defaulting to regional

app/views/admin/calendars/new.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  <% else %>
+    <%= form.input :calendar_type, as: :radio_buttons, collection: Calendar.calendar_types.keys, label_method: :humanize %>
+  <% end %>
+
+  <%= form.input :based_on_id, as: :select, collection: Calendar.national, include_blank: false %>
```
jamesalmond: Hmm, if we're forced to pick a national calendar then maybe this should just be a regional only form and we don't need the calendar_type select at all

app/views/admin/school_groups/_form.html.erb
```diff
@@ -24,8 +24,8 @@
         The following defaults are used to pre-populate values when creating a school. Changing them here will not affect areas for schools currently in this group.
       </p>
       <div class="form-group">
-        <%= f.label :default_calendar_area_id, 'Default Calendar Area' %>
-        <%= f.select :default_calendar_area_id, options_from_collection_for_select(CalendarArea.order(:title), 'id', 'title', @school_group.default_calendar_area_id), {include_blank: true}, { class: 'form-control' } %>
+        <%= f.label :default_template_calendar_id, 'Default Template Calendar' %>
+        <%= f.select :default_template_calendar_id, options_from_collection_for_select(Calendar.template.order(:title), 'id', 'title', @school_group.default_template_calendar_id), {include_blank: true}, { class: 'form-control' } %>
```
jamesalmond: Not sure about the `template` scope on the model. I think I'd prefer it to be clear that we're using regional calendars here.

app/controllers/schools/timeline_controller.rb
jamesjefferies: Ah yes, I see what you mean - they are at the national level I think, so, it *should* either be @school.template_calendar.based_on.academic_years or @school.calendar.based_on.based_on.academic_years...

jamesjefferies: I think I was getting confused myself!

app/models/ability.rb
jamesjefferies: Yes, bad James J

app/models/school.rb
jamesjefferies: Later... There's already a lot in this PR!

app/views/admin/calendars/new.html.erb
jamesjefferies: Currently the only link to create a calendar, sets the calendar type to regional, so I've set it up so that if something has already been set, the user doesn't see the options. I don't *think* at the moment, we want a user trying to create a national or school calendar at all.

app/views/admin/calendars/new.html.erb
jamesjefferies: Yes, good point, it's inconsistent at the moment isn't it.

app/views/admin/school_groups/_form.html.erb
jamesjefferies: I agree - that shouldn't be template, should be regional

PR #794, opened 2019-09-19
Sort out UI

jamesjefferies:


app/controllers/schools/low_carbon_hub_installations_controller.rb
```diff
@@ -0,0 +1,42 @@
+module Schools
+  class LowCarbonHubInstallationsController < ApplicationController
+    load_and_authorize_resource :school
+    load_and_authorize_resource through: :school
+
+    def index
+      @low_carbon_hub_installations = @school.low_carbon_hub_installations
```
jamesalmond: Doesn't the above `through` do this already?

app/views/schools/low_carbon_hub_installations/show.html.erb
```diff
@@ -0,0 +1,11 @@
+<h1>Low carbon hub installation <%= @low_carbon_hub_installation.rbee_meter_id %></h1>
+<h2><%= @school.name %></h2>
+
+<p><%= link_to "All low carbon hub installations for #{@school.name}",  school_low_carbon_hub_installations_path(@school) %>
+<% JSON.parse(@low_carbon_hub_installation.information).each do |key, value| %>
```
jamesalmond: Do these need to be in a `<dl>`?

app/controllers/schools/low_carbon_hub_installations_controller.rb
jamesjefferies: yes, affirmative

app/views/schools/low_carbon_hub_installations/show.html.erb
jamesjefferies: Not only that, but I've got the dt and the dd the wrong way around - just been working with HTML since 1995, no biggie

spec/system/low_carbon_hub_installation_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      click_on 'Manage Low carbon hub installations'
+      expect(page).to have_content("There are no Low carbon hub installations at the moment for this school")
+      click_on 'New Low carbon hub installation'
+    end
+
+    it 'I can add a low carbon hub installation' do
+      allow(LowCarbonHubMeterReadings).to receive(:new).and_return(@low_carbon_hub_api_instance)
+
+      expect(@low_carbon_hub_api_instance).to receive(:full_installation_information).with(rbee_meter_id).and_return(information)
```
jamesalmond: Could these be in the shared context too?

jamesjefferies: Yes, but now with expect as the blowing up last test doesn't call these. Still, I'll switch to allow and move them.

PR #807, opened 2019-09-25
Alert prioritisation

jamesalmond:


app/models/management_priority.rb
```diff
@@ -8,6 +8,7 @@
 #  created_at                           :datetime         not null
 #  find_out_more_id                     :bigint(8)
 #  id                                   :bigint(8)        not null, primary key
+#  priority                             :float            default(0.0), not null
```
jamesjefferies: Maybe decimal, although the analytics uses float, this application uses decimal

PR #825, opened 2019-09-30
Enable controls for line charts

jamesalmond:


app/views/schools/charts/show.html.erb
```diff
@@ -41,5 +41,6 @@ with <%= @school.number_of_pupils %> pupils and a floor area of <%= @school.floo
 
 <div id="chart_wrapper_1">
   <h3 class="analysis">Loading: <%= @title %></h3>
-  <%= chart_tag(@school, @chart_type.to_s, index: 1, chart_config: {y_axis_units:  @measurement}) %>
+  <%= render 'shared/analysis_controls' if params[:control] %>
```
jamesalmond: I've added this quick flag for now that just renders the given chart with controls so we can play around with it. First scatter chart I tried didn't "just work".

jamesjefferies: Ok - are we happy fo rnow to have params[:control] referenced here - just wondering if it isn't better for the controller to even have a @show_controls = params[:control] & a @show_advice = !params[:control].present?

jamesalmond: Yes, good idea.

PR #826, opened 2019-09-30
Add delete to event modal

jamesjefferies:


spec/system/calendar_spec.rb
```diff
@@ -6,11 +6,34 @@
...
-    it 'shows the calendar' do
+    it 'shows the calendar, allows an event to be added and deleted' do
+      create(:academic_year, calendar: calendar, start_date: Date.parse("01/01/#{Date.today.year}"))
+
       sign_in(admin)
       visit calendar_path(calendar)
       expect(page.has_content?(calendar.title)).to be true
       expect(page.has_content?('January')).to be true
+
+      find('div.month-container:nth-child(1) > table:nth-child(1) > tbody:nth-child(2) > tr:nth-child(2) > td:nth-child(3) > div:nth-child(1)').click
```
jamesalmond: I think these selectors are a bit complicated. Maybe something like? `fifteenth_jan = find(".calendar .day", text: '15', match: :first)` etc?

app/assets/javascripts/calendar_year_view.js
```diff
@@ -21,6 +21,15 @@ $(document).ready(function() {
       $('#event-modal input[name="calendar_event[start_date]"]').val(startDate.toLocaleDateString("en-GB"));
       $('#event-modal input[name="calendar_event[end_date]"]').val(endDate.toLocaleDateString("en-GB"));
 
+      $('#delete_button').show();
+
+      $('#delete_button').on('click', function(event) {
+        event.preventDefault();
+
+        $('#event-modal input[name="_method"]').val('delete');
```
jamesalmond: Nice solution!

PR #831, opened 2019-10-02
Allow editing of contact email

jamesjefferies:


app/views/admin/school_onboardings/email/edit.html.erb
```diff
@@ -0,0 +1,13 @@
+<h1>Change email address for <%= @school_onboarding.school_name %></h1>
+
+<%= simple_form_for([:admin, @school_onboarding], url: admin_school_onboarding_email_path(@school_onboarding)) do |f| %>
+
+  <%= render 'shared/errors', subject: @school_onboarding, subject_name: 'form' %>
```
jamesalmond: Do we need this with simple form? I don't think we use it elsewere.

app/models/school_onboarding.rb
```diff
@@ -57,6 +57,10 @@ def has_event?(event_name)
     events.where(event: event_name).any?
   end
 
+  def has_only_sent_email?
+    events.count == 1 && has_event?(:email_sent)
```
jamesalmond: I wonder now if this would be better as "only has email sent or reminder sent"?

app/views/admin/school_onboardings/email/edit.html.erb
jamesjefferies: No we don't I turned this from a form_for to a simple_form_for and should have taken out that bit...

PR #835, opened 2019-10-02
Equivalence carousel & Pupil dashboard tweaks

jamesjefferies:


app/views/pupils/schools/_equivalences.html.erb
```diff
@@ -0,0 +1,31 @@
+<div id="pupil_carousel" class="carousel slide pupil-dashboard" data-wrap="false">
+  <div class="carousel-inner">
+    <% equivalences_content.each_with_index do |equivalence_content, index| %>
+      <div class="carousel-item <%= 'active' if index == 0 %>">
+        <div class="row illustration-background">
+          <div class="col-6">
+            <div class="trix-content">
+              <%= sanitize equivalence_content.equivalence %>
+            </div>
+            <% if defined?(school) %>
```
jamesalmond: Could this just be `if school`? Do we ever call it without setting a `school` variable?

jamesjefferies: It's because this was shared across the equivalence preview (where a school was not set) - I've re-introduced the old partial which the preview can use whilst keeping the carousel for the actual page. There is a slight overhead in that if we want to make any layout changes for an equivalence, we'd need to do it in both places, but also keeps it simple for now.

PR #845, opened 2019-10-04
Sorting rebrand of 404,422,500

jamesjefferies:


public/404.html
```diff
@@ -44,66 +72,61 @@ <h2>Page not found (404)</h2>
...
                 </li>
-                <li class="list-inline-item">
-                  <a href="https://twitter.com/energy_sparks" class="btn-social btn-outline"><i class="fa fa-fw fa-twitter"></i></a>
+                <li class="list-inline-item p-2">
+                  <a href="https://www.facebook.com/EnergySparksUK/" class=""><i class="fab fa-facebook-square fa-2x"></i></a>
                 </li>
-                <li class="list-inline-item">
-                  <a href="https://github.com/BathHacked/energy-sparks" class="btn-social btn-outline"><i class="fa fa-fw fa-github"></i></a>
+                <li class="list-inline-item p-2">
+                  <a href="https://github.com/BathHacked/energy-sparks" class=""><i class="fab fa-github fa-2x"></i></a>
```
jamesalmond: These icons don't seem to render on test.

jamesalmond: Possible the old 404 css file is still in the cache? Could that be causing problems?

jamesalmond: Or, I'm getting confused and it's actually my browser that's caching it because these assets don't go through cloudfront...

jamesjefferies: Nah, they're missing for me at home too - the font files haven't loaded...

PR #847, opened 2019-10-04
Add reports - I cracked

jamesjefferies:


jamesalmond:
@jamesjefferies I don't think we have the resources to maintain these as we add to the alerts process and adding them sets a bad precedent.

PR #849, opened 2019-10-08
Redirect nav link to home page, regardless of logged in status

jamesjefferies:


app/controllers/home_controller.rb
```diff
@@ -4,9 +4,10 @@ class HomeController < ApplicationController
   before_action :redirect_if_logged_in, only: :index
 
   def index
-    # This renders using rails magic, the home layout template which
-    # does not include the application container
-    @active_schools = School.active.count
+  end
+
+  def show
```
jamesalmond: Can any route get to this action?

PR #850, opened 2019-10-08
Add privacy consent checkbox

jamesjefferies:


app/controllers/onboarding/consent_controller.rb
```diff
@@ -6,11 +6,24 @@ class ConsentController < BaseController
...
+      @school_onboarding_consent = SchoolOnboardingConsent.new
     end
 
     def create
-      @school_onboarding.events.create!(event: :permission_given)
-      redirect_to new_onboarding_account_path(@school_onboarding)
+      @school_onboarding_consent = SchoolOnboardingConsent.new(school_onboarding_consent_params)
+
+      if @school_onboarding_consent.valid?
+        @school_onboarding.events.create!(event: :permission_given)
```
jamesalmond: I think we should create another event here like "Privacy policy agreed" so we can differentiate between schools who have agreed and those who haven't

jamesjefferies: We could do, but I'm not letting them process *unless* they select the checkbox, so permission given is for data and privacy.

jamesalmond: But we have lots of schools that have already been through the onboarding process that haven't explicitly agreed to the privacy policy. I think we should differentiate between those who have seen the checkbox and clicked on it vs those who have never seen the checkbox but given permission.

jamesjefferies: @jamesalmond that indeed is a fair and well made point :)

jamesjefferies: @jamesalmond and as the events are effectively a list about what has happened, we can just add a new event to SchoolOnboardingEvent and create it when we do the permission given event.

@school_onboarding.events.create!(event: :privacy_policy_agreed)

PR #851, opened 2019-10-08
Migrate alert type description field to rich text

jamesjefferies:
Actually, I'm not sure we needed to do this after all, still, tidies up old stuff, updates notes...

### DIFF
--- a/app/services/template_interpolation.rb
+++ b/app/services/template_interpolation.rb
@@ -31,12 +31,16 @@ class TemplateInterpolation
 private
 
   def process(template, variables)
-    Mustache.render(template, variables)
+    Mustache.render(template_as_string(template), variables)
   end
 
   def get_variables(template)
     mustache = Mustache.new
-    mustache.template = template
+    mustache.template = template_as_string(template)
     mustache.template.tags
+  end
+
+  def template_as_string(template)
+    template.is_a?(ActionText::RichText) ? template.body.to_s : template
   end
 end
--- a/app/services/template_interpolation.rb
+++ b/app/services/template_interpolation.rb
@@ -15,7 +15,12 @@ class TemplateInterpolation
       collection
     end
     templated = fields.inject(with_proxied_objects) do |collection, field|
-      collection[field] = process(@object.send(field) || "", with)
+      template = @object.send(field) || ""
+      collection[field] = if template.is_a?(ActionText::RichText)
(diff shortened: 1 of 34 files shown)


spec/system/admin/alert_type_management_spec.rb
```diff
@@ -248,13 +248,13 @@
       fill_in 'Title', with: ''
       click_on 'Update Alert type'
 
-      expect(page).to have_content("Title can't be blank")
+      expect(page).to have_content("Title *\ncan't be blank")
```
jamesjefferies: Removed shared errors, so content needs to be simple form error content....

jamesjefferies:
Maybe - I think I need to spend some time doing a full audit of our text fields, text areas.

jamesjefferies:
List  for discussion on : https://trello.com/c/uHxmsDav

jamesjefferies:
@jamesalmond have added migrations for further fields including titles - I need to re-review myself when back in the office in case I've made any boo-boos

app/services/template_interpolation.rb
```diff
@@ -31,12 +31,16 @@ def variables(*fields)
...
   def get_variables(template)
     mustache = Mustache.new
-    mustache.template = template
+    mustache.template = template_as_string(template)
     mustache.template.tags
   end
+
+  def template_as_string(template)
+    template.is_a?(ActionText::RichText) ? template.body.to_s : template
+  end
```
jamesjefferies: You ok with this? @jamesalmond it works ok, but feels a bit yuck - ActionText::RichText does not support to_s so you need to get the body (which is ActionText::Content) and then to_s that.

app/views/admin/alert_types/show.html.erb
```diff
@@ -9,11 +9,11 @@
...
   <dt class="col-md-2">Sub category</dt>
   <dd class="col-md-10"><%= @alert_type.sub_category ? @alert_type.sub_category.to_s.capitalize : 'Not set' %></dd>
-
-  <dt class="col-md-2">Description</dt>
-  <dd class="col-md-10"><%= simple_format(@alert_type.description)  %></dd>
 </dl>
 
+<h2>Description</h2>
+<div class="trix-content"><%= @alert_type.description %></div>
+
```
jamesjefferies: Pulled this out of the <dl> into it's own section

PR #854, opened 2019-10-09
New day week charts

jamesalmond:


app/models/chart_data.rb
```diff
@@ -56,6 +56,12 @@ def customised_chart_config(chart_manager)
     if @chart_config_overrides[:mpan_mprn].present?
       chart_config[:meter_definition] = @chart_config_overrides[:mpan_mprn].to_i
     end
+    if @chart_config_overrides[:series_breakdown].present?
+      chart_config[:series_breakdown] = @chart_config_overrides[:series_breakdown].to_sym
+    end
+    if @chart_config_overrides[:date_ranges].present?
+      chart_config[:timescale] = @chart_config_overrides[:date_ranges].map {|range| { daterange: range }}
+    end
```
jamesjefferies: Maybe not with these changes - but I wonder whether at some point we should refactor Chart Data and separate out the config management in to something else which is possibly a bit more testable in isolation

jamesalmond: Yes, there's an obvious split in this file. Hard to test the config stuff when you have calls to the drilldown happening too.

app/views/schools/usage/_within_school_card.html.erb
```diff
@@ -0,0 +1,20 @@
+<div class="card <%= current_supply %>-card" id="within-school" style="display: none;">
+  <div class="card-header"></div>
+  <div class="card-body">
+    <h4 class="card-title">
+      <%= date_option %>
+    </h4>
+    <div class="form-group">
+      <div class="input-group date <%= 'week-view' if week_picker %>" id="datetimepicker3" data-target-input="nearest">
```
jamesjefferies: Do we need to render this in full rather than using the customised date picker simple form thingy?

lib/amr_usage.rb
```diff
@@ -1,46 +1,7 @@
...
-  end
-
-  def hourly_usage_for_date(supply: nil, date: nil, meter: nil, scale: :kwh)
-    multiplier = scale == :kwh ? 1 : 2
-
-    grouped_readings = readings_with_selected_meters(supply, meter).where(reading_date: date.all_day)
-    grouped_readings.pluck(:kwh_data_x48).transpose.map(&:sum).each_with_index.map { |consumption, index| [MINUTES[index], (consumption.to_f * multiplier)] }
-  end
-
+  #
```
jamesjefferies: Do we still use ```last_n_days_with_readings``` in this module?

jamesjefferies:
I'm enjoying the removal of lots of old code here - I was wondering whether it was worth or even possible doing a system test mocking out the analytics calls with some chart data. We don't do this elsewhere and it'd be a faff to set up.

app/views/schools/usage/_within_school_card.html.erb
jamesalmond: I would have to convert the whole (or part of?) the form to simple form to get access to those, I think. Can look at it if you think it's worth it.

lib/amr_usage.rb
jamesalmond: Pushed a change removing the whole thing!

jamesalmond:
@jamesjefferies there really should be more tests on this. But stubbing out analytics for a JS test that renders a chart sounds like a whole world of pain. Maybe we should concentrate on testing the individual components better?

app/models/chart_data.rb
jamesalmond: I've done a first-pass at moving this out (to get the coverage up a bit!).

PR #856, opened 2019-10-10
Show subtitles on pupil analysis chart pages

jamesalmond:


app/views/pupils/analysis/show.html.erb
```diff
@@ -1,12 +1,11 @@
...
-
 <div id="chart_wrapper_1" class="pt-3">
   <div class="d-flex justify-content-between align-items-center">
     <div>
-      <%= link_to 'Back', request.referer.present? ? request.referer : pupils_school_analysis_path(@school), class: 'btn btn-rounded' %>
+      <%= link_to 'Back', request.referer.present? ? request.referer : pupils_school_analysis_path(@school), class: 'btn btn-rounded mr-5' %>
     </div>
     <div>
-      <h3><%= @title %></h3>
+      <h3></h3>
```
jamesjefferies: So we no longer show the title whilst loading? Ah, @title isn't set anyway is it?

jamesalmond: Yes, it isn't set by anything. I did try setting it to `Loading...` but I never actually saw it as the JS executed so quickly.

PR #859, opened 2019-10-11
Allow users to edit, create, destroy locations

jamesjefferies:


app/views/schools/locations/show.html.erb
```diff
@@ -0,0 +1,7 @@
+<h1>Observations for <%= nice_date_times @observation.at %></h1>
```
jamesjefferies: Remove show action, route and template

PR #862, opened 2019-10-11
Show inactive page for users who try and view an inactive school with…

jamesjefferies:
…out admin permissions

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -26,7 +26,7 @@ class Ability
       can :manage, :all
       can :analyse, :test
       can :analyse, :heating_model_fitting
-      cannot :read, :school_menu
+      cannot :read, :my_school_menu
     elsif user.school_admin? || user.group_admin?
       if user.group_admin?
         school_scope = { school_group_id: user.school_group_id }
@@ -38,6 +38,7 @@ class Ability
         can :manage, CalendarEvent do |calendar_event|
           user.school_group.calendars.include?(calendar_event.calendar)
         end
+        can :read, :my_school_menu, School, active: true
       else
         school_scope = { id: user.school_id }
         related_school_scope = { school_id: user.school_id }
@@ -46,7 +47,7 @@ class Ability
         can :manage, SchoolOnboarding do |onboarding|
           onboarding.created_user == user
         end
-        can :read, :school_menu
+        can :read, :my_school_menu, School, active: true
         can :read, :dashboard_menu
       end
       can [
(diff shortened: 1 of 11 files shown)


app/views/shared/_common_nav_bar_menus.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= render 'shared/my_school_menu' if can?(:read, :school_menu) %>
+<%= render 'shared/my_school_menu' if can?(:read, :school_menu) && current_user.school.active? %>
```
jamesjefferies: Trying to think whether we can have this just as an ability rather than checking active school here

app/views/shared/_sub_nav.html.erb
```diff
@@ -8,7 +8,7 @@
       </button>
       <div class="collapse navbar-collapse" id="subnavbarNavDropdown">
         <ul class="navbar-nav">
-          <%= render 'shared/dashboards' if current_user %>
+          <%= render 'shared/dashboards' if can?(:read, :dashboard_menu) && school.active? %>
```
jamesjefferies: and here too, but this school might not be the current user's school

spec/system/calendar_spec.rb
```diff
@@ -25,7 +25,7 @@
       expect { click_on('Save') }.to change { calendar.calendar_events.count }.by(1)
 
       # Wait until ajax call is back
-      assert_selector('td[style]')
+      assert_selector('td[style*="background-color"]')
```
jamesjefferies: Saw this failed on master, always worth trying to make the selector more fussy to see if that fixes it!

app/models/ability.rb
```diff
@@ -22,6 +22,13 @@ def initialize(user)
     can :read, Observation
     can :read, ProgrammeType
 
+    can :manage, Location, school_id: user.school_id
```
jamesalmond: Should these be here?

app/views/shared/_common_nav_bar_menus.html.erb
jamesalmond: `can?(:my_school_menu, current_user.school?)`

app/views/shared/_sub_nav.html.erb
jamesalmond: I don't think it matters if it's their school or not? We just need to check they're the right role and the school is active?

app/views/shared/_common_nav_bar_menus.html.erb
jamesalmond: Hmmm. Yes. Another instance of the cancan approach not quite fitting with our multi-tennant app.

PR #864, opened 2019-10-14
Pupil password case

jamesalmond:


app/models/school.rb
```diff
@@ -205,7 +205,7 @@ def latest_dashboard_alerts
   end
 
   def authenticate_pupil(pupil_password)
-    users.pupil.to_a.find {|user| user.pupil_password == pupil_password}
+    users.pupil.to_a.find {|user| user.pupil_password.casecmp(pupil_password).zero? }
```
jamesjefferies: I *think* you can now use ```casecmp?(pupil_password)``` :)

PR #867, opened 2019-10-16
Fix scopes to not include not

jamesalmond:


app/models/alert.rb
```diff
@@ -44,8 +44,8 @@ class Alert < ApplicationRecord
 
   scope :rating_between, ->(from, to) { where("rating BETWEEN ? AND ?", from, to) }
 
-  enum enough_data: [:enough, :not_enough, :minimum_might_not_be_accurate]
-  enum relevance: [:relevant, :not_relevant, :never_relevant]
+  enum enough_data: [:enough, :not_enough, :minimum_might_not_be_accurate], _prefix: :data
+  enum relevance: [:relevant, :not_relevant, :never_relevant], _prefix: :relevance
```
jamesalmond: Rather than rename these, I've chosen to add a prefix for the scopes. We don't use the scopes and if we change the name of the enum values then we fall out of sync with the analytics return values for these fields.

PR #868, opened 2019-10-16
Chart embeds

jamesalmond:


app/assets/javascripts/analysis.js
```diff
@@ -3,16 +3,14 @@
...
   var $chartDiv = $(chart.renderTo);
   var $chartWrapper = $chartDiv.parent();
-  var $titleH3 = $chartWrapper.find('h3');
 
-  $titleH3.text(title + ' chart');
   $chartWrapper.addClass('alert alert-warning');
-  $chartDiv.remove();
+  $chartWrapper.html(`<h3>${title} chart</h3>`)
 
   $('div#nav-row').before('<div class="alert alert-warning" role="alert">' + title + ' <a href="#' + $chartWrapper.attr('id') + '" class="alert-link">chart</a></div>');
```
jamesjefferies: I think we can get rid of this now as we only display charts which load

app/assets/javascripts/analysis.js
```diff
@@ -3,16 +3,14 @@
...
 }
 
-function chartSuccess(chart_data, chart, noAdvice, noZoom) {
+function chartSuccess(chartConfig, chart_data, chart, noAdvice, noZoom) {
```
jamesjefferies: change chart_data to chartData at some point

PR #873, opened 2019-10-18
Render example variables in alerts view

jamesalmond:


app/controllers/admin/alert_types/ratings_controller.rb
```diff
@@ -28,6 +28,7 @@ def create
       def edit
         @rating = @alert_type.ratings.find(params[:id])
         @content = @rating.current_content
+        @example_variables = load_example_variables(@rating)
```
jamesalmond: We're only doing this on edit/update at the moment as on create we don't know the ratings to get the alert for yet. Asked Claudia what she'd prefer to do in this instance: render example variables regardless of rating or just only render them on edit.

PR #884, opened 2019-10-22
Responsive updates

jamesjefferies:
- [x] Set chart height thing to be dependent on screen width


app/views/teachers/schools/_activities_and_actions.html.erb
```diff
@@ -3,10 +3,14 @@
     <%= render 'activity_types/card', activity_type: activity_type, based_on: 'previous activities'%>
   <% end %>
```
jamesjefferies: This makes the three activity/programme boxes drop into a single column a bit sooner

app/views/teachers/schools/show.html.erb
```diff
@@ -28,7 +28,7 @@
   </div>
   <div class="card border-0">
   </div>
-    <div class="card border-0">
+    <div class="card border-0 d-block d-sm-none d-md-block">
```
jamesjefferies: Hide this when view port shrinks - we switch to three columns and show it further down instead.

PR #889, opened 2019-10-24
Sitemap generator

jamesjefferies:


config/sitemap.rb
```diff
@@ -0,0 +1,42 @@
...
+    add datasets_path
+    add team_path
+    add getting_started_path
+    add scoring_path
+    add privacy_and_cookie_policy_path
+
+    add schools_path
+    add activity_types_path
+
+    ActivityType.find_each do |activity_type|
```
jamesalmond: Only active?

config/app/controllers/robots_txts_controller.rb
```diff
@@ -0,0 +1,15 @@
+class RobotsTxtsController < ApplicationController
```
jamesalmond: Is this a duplicate file?

jamesjefferies: Yes!

config/application.rb
```diff
@@ -3,7 +3,7 @@
 require "rails"
 require "rails/all"
 require "active_storage/engine"
-require_relative "../lib/rack_x_robots_tag"
+require_relative "../lib/rack/x_robots_tag"
```
jamesalmond: @jamesjefferies Something to do with autoloading doesn't like that the file name wasn't related to the class name and meant test kept cycling and wouldn't load.

jamesjefferies: woah, that's a bit weird isn't it? I didn't even look to make sure the test deploy had worked - more fool me... you fixed it then?

jamesjefferies: thanks :)

jamesalmond: Yes, this seems to fix it!

PR #896, opened 2019-11-04
Update analytics including Benchmarking

jamesjefferies:
Things to do

- [x] Check basic principles work - do we get some benchmark variables back to persist in DB?
- [x] Decide whether to store benchmark variables with alert, or in a separate model
- [ ] Work out what new methods actually do and whether we should use them across the board - what are the valid/failure states - can you end up with broken benchmarks, but working priorities for example (see below)
- [ ] Write some tests around the benchmark behaviour
- [ ] Helpful if analytics gave us the benchmark variables with the full name rather than expecting the front end to add short name etc
- [ ] Look at generating the actual benchmark content
- [ ] Variable names should be pre-processed
- [ ] Variable data needs to be strings and numbers, not objects
- [ ] When passing the benchmark data back, do we want to always pass all the data back, even when we are just looking at, say Sheffield?
- [ ] Filtering we would want to pass through the relevant school ids rather than introspect the variables.

Email about generation process first? Marshall handling investigate

### New methods 
New methods in analytics alert framework - calculation worked and make available to users?
Looks like make_available_to_users checks whether relevant and enough data and calculation worked.

Previously we would check for

```
    valid &&
      !enough_data.nil? &&
      !(enough_data == :not_enough) &&
      relevance == :relevant &&
      !rating.nil?

```
New method doesn't check whether rating is set, nor valid


app/services/alerts/adapters/analytics_adapter.rb
```diff
@@ -11,12 +11,13 @@ def report
...
 
-        variables = if pull_variable_data?(analysis_object)
+        variables = if analysis_object.make_available_to_users?
                       {
                         template_data: analysis_object.front_end_template_data,
                         chart_data:    analysis_object.front_end_template_chart_data,
                         table_data:    analysis_object.front_end_template_table_data,
-                        priority_data: analysis_object.priority_template_data
+                        priority_data: analysis_object.priority_template_data,
+                        benchmark_data: (alert_class.benchmark_template_variables.empty? ? {} : analysis_object.benchmark_template_data)
```
jamesjefferies: Do we need this, or was this just to make it work?

app/services/alerts/adapters/analytics_adapter.rb
```diff
@@ -11,12 +11,13 @@ def report
       def produce_report(analysis_object)
         analysis_object.analyse(@analysis_date)
 
-        variables = if pull_variable_data?(analysis_object)
+        variables = if analysis_object.make_available_to_users?
```
jamesjefferies: Is this reliable for all data, or just benchmark data?

If we run one alert - make_available_to_users? is the highest level flag to let us know everything is ok, great, but does this guarantee benchmark, priority, alert, chart data, rating, valid etc is all populated?

PR #898, opened 2019-11-06
Specify exact models for specs to remove ordering test failures

jamesalmond:


spec/services/next_activity_suggester_with_filter_spec.rb
```diff
@@ -109,18 +108,8 @@
 
     context 'where the school has completed the activity' do
       it 'does not use the activity type' do
-        programme.programme_activities[0].update!(activity: create(:activity, activity_type: programme.programme_activities[0].activity_type, school: school))
```
jamesalmond: Hypothesis for failing test: `programme.programme_activities[0]` doesn't always return the same programme activity. The tests already exclude ks2 activity types as the school does not have ks2 as one of its key stages. 95% of the time the programme_activities are returned with ks1 first, which is then updated to be complete and then filtered out. The remaining 5% of the time, `programme.programme_activities[0]` returns the ks2 activity which is already filtered out and so ks1 and ks3 activity types are still returned.

PR #910, opened 2019-11-08
Refactor batch jobs to become one content run

jamesjefferies:


app/services/alerts/generate_and_save_alerts.rb
```diff
@@ -1,8 +1,9 @@
 module Alerts
   class GenerateAndSaveAlerts
-    def initialize(school, alert_framework_adapter = FrameworkAdapter)
+    def initialize(school:, framework_adapter: FrameworkAdapter, aggregate_school: AggregateSchoolService.new(school).aggregate_school)
```
jamesjefferies: Best check all the existing rake tasks to ensure that they all work still

PR #914, opened 2019-11-11
Re-jig alert creation

jamesjefferies:
- [x] More tests around framework adapter and generate and save alerts
- [x] Check for implications around System alerts
- [x] generate method in GenerateAndSaveAlerts is a bit bleurgh
- [ ] Backfill alert errors?


app/services/alerts/generate_and_save_alerts.rb
```diff
@@ -2,39 +2,83 @@ module Alerts
   class GenerateAndSaveAlerts
     def initialize(school:, framework_adapter: FrameworkAdapter, aggregate_school: AggregateSchoolService.new(school).aggregate_school)
       @school = school
-      @alert_framework_adapter = framework_adapter
+      @alert_framework_adapter_class = framework_adapter
       @aggregate_school = aggregate_school
+      @alert_generation_run = AlertGenerationRun.create(school: @school)
```
jamesalmond: `create!`?

jamesalmond: Think I'd prefer this creation on the `perform` method

app/services/alerts/generate_and_save_alerts.rb
```diff
@@ -2,39 +2,83 @@ module Alerts
...
+      @alerts_to_save = []
+      @alert_errors_to_save = []
```
jamesalmond: Not sure these need to be instance variables? Could perform also collect them?

Possibly unlikely to happen, but what would the intended effect of running `perform` twice be? Currently it would produce a single generation run with 3 lots of alerts/errors in. One lot from the first run, and then the lot from the first run again and the lot added from the second run.

app/models/alert_error.rb
```diff
@@ -0,0 +1,25 @@
...
+#
+# Indexes
+#
+#  index_alert_errors_on_alert_generation_run_id  (alert_generation_run_id)
+#  index_alert_errors_on_alert_type_id            (alert_type_id)
+#  index_alert_errors_on_school_id                (school_id)
+#
+
+class AlertError < ApplicationRecord
+  belongs_to :school
```
jamesalmond: No need for school on here as it can be inferred from the run? Same with alerts although there might be more to unpick there....

PR #923, opened 2019-11-13
Create benchmark results and save data

jamesjefferies:


app/services/alerts/alert_attributes_factory.rb
```diff
@@ -0,0 +1,31 @@
+module Alerts
+  class AlertAttributesFactory
```
jamesjefferies: This could be a Builder rather than a Factory - or even could be added to the Alert Class

app/services/alerts/error_attributes_factory.rb
```diff
@@ -0,0 +1,23 @@
+module Alerts
+  class ErrorAttributesFactory
```
jamesjefferies: This could be a Builder rather than a Factory - or even could be added to the AlertError Class

app/services/alerts/alert_attributes_factory.rb
```diff
@@ -0,0 +1,31 @@
...
+        displayable:              @alert_report.displayable?,
+        analytics_valid:          @alert_report.valid,
+        rating:                   @alert_report.rating,
+        enough_data:              @alert_report.enough_data,
+        relevance:                @alert_report.relevance,
+        template_data:            @alert_report.template_data,
+        chart_data:               @alert_report.chart_data,
+        table_data:               @alert_report.table_data,
+        priority_data:            @alert_report.priority_data,
+        created_at:               now,
```
jamesalmond: Rails should handle these for us if we're just using the current time

app/services/alerts/benchmark_attributes_factory.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+
+    def generate
+      now = Time.zone.now
+      {
+        alert_generation_run_id:  @alert_generation_run_id,
+        alert_type_id:            @alert_report.alert_type.id,
+        asof:                     @alert_report.asof_date,
+        data:                     @alert_report.benchmark_data,
+        created_at:               now,
```
jamesalmond: Ditto

app/services/alerts/generate_alert_reports.rb
```diff
@@ -0,0 +1,50 @@
+module Alerts
+  class GenerateAlertReports
+    def initialize(
+        school:,
+        alert_generation_run_id:,
```
jamesalmond: Why pass round the id and not the model?

db/migrate/20191111133812_create_benchmark_results.rb
```diff
@@ -0,0 +1,11 @@
+class CreateBenchmarkResults < ActiveRecord::Migration[6.0]
+  def change
+    create_table :benchmark_results do |t|
+      t.references  :alert_generation_run
+      t.references  :alert_type
+      t.date        :asof
+      t.text        :data
```
jamesalmond: `null: false`?

db/migrate/20191111133812_create_benchmark_results.rb
```diff
@@ -0,0 +1,11 @@
+class CreateBenchmarkResults < ActiveRecord::Migration[6.0]
+  def change
+    create_table :benchmark_results do |t|
+      t.references  :alert_generation_run
```
jamesalmond: Foreign key needed

db/schema.rb
```diff
@@ -230,12 +230,12 @@
     t.decimal "public_dashboard_alert_weighting", default: "5.0"
     t.decimal "teacher_dashboard_alert_weighting", default: "5.0"
     t.decimal "find_out_more_weighting", default: "5.0"
+    t.text "find_out_more_table_variable", default: "none"
```
jamesalmond: Not part of this merge

app/services/alerts/generate_alert_reports.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    def perform
+      alert_reports = []
+      alert_errors_to_save = []
+
+      relevant_alert_types.each do |alert_type|
+        alert_reports << generate_alert_report(alert_type, alert_errors_to_save)
+      end
+
+      AlertError.insert_all!(alert_errors_to_save) unless alert_errors_to_save.empty?
```
jamesalmond: Not sure we should have two places where we save errors

db/schema.rb
jamesjefferies: I'm confused by this! If I run db:migrate on production database on master, this variable moves in the db/schema.rb !

PR #928, opened 2019-11-15
Refactor out benchmark creation

jamesjefferies:


app/models/benchmark_result.rb
```diff
@@ -2,27 +2,27 @@
...
-#  alert_generation_run_id :bigint(8)        not null
-#  alert_type_id           :bigint(8)        not null
-#  asof                    :date             not null
-#  created_at              :datetime         not null
-#  data                    :text
-#  id                      :bigint(8)        not null, primary key
-#  updated_at              :datetime         not null
+#  alert_type_id                      :bigint(8)        not null
+#  asof                               :date             not null
+#  benchmark_result_generation_run_id :bigint(8)
```
jamesalmond: not null?

app/services/alerts/framework_adapter.rb
```diff
@@ -19,6 +19,11 @@ def content
       adapter_instance.content
     end
 
+    def benchmark_dates
+      adapter_class(@alert_type).new(alert_type: @alert_type, school: @school, analysis_date: @analysis_date, aggregate_school: @aggregate_school)
```
jamesalmond: Is this line needed? We're using adapter instance below.

jamesjefferies: No, should be nobbled

PR #942, opened 2019-11-21
Alert production issues

jamesjefferies:
- Run on date not being set as not being saved at alert level
- Default to whatever the analytics adapter decides is the best asof rather than running today - lots of alerts fail if you run them for today...

### DIFF
--- a/app/services/alerts/alert_attributes_factory.rb
+++ b/app/services/alerts/alert_attributes_factory.rb
@@ -1,10 +1,11 @@
 module Alerts
   class AlertAttributesFactory
-    def initialize(school, alert_report, alert_generation_run, alert_type)
+    def initialize(school, alert_report, alert_generation_run, alert_type, asof_date)
       @school = school
       @alert_report = alert_report
       @alert_generation_run = alert_generation_run
       @alert_type = alert_type
+      @asof_date = asof_date
     end
 
     def generate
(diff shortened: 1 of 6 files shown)


app/services/alerts/alert_attributes_factory.rb
```diff
@@ -1,18 +1,19 @@
...
+      @asof_date = asof_date
     end
 
     def generate
       {
         school_id:                @school.id,
         alert_generation_run_id:  @alert_generation_run.id,
         alert_type_id:            @alert_type.id,
-        run_on:                   @alert_report.asof_date,
+        run_on:                   @asof_date,
```
jamesjefferies: This is never set in the alert report (was at one point) therefore pass it in to set the run on date.

app/services/alerts/generate_and_save_alerts.rb
```diff
@@ -37,7 +37,7 @@ def process_alert_type_run_result(alert_type_run_result)
 
     def process_alert_report(alert_type, alert_report, asof_date)
       if alert_report.valid
-        Alert.create(AlertAttributesFactory.new(@school, alert_report, @alert_generation_run, alert_type).generate)
+        Alert.create(AlertAttributesFactory.new(@school, alert_report, @alert_generation_run, alert_type, asof_date).generate)
```
jamesjefferies: as above

app/services/alerts/generate_and_save_benchmarks.rb
```diff
@@ -10,7 +10,7 @@ def initialize(
       @framework_adapter = framework_adapter
     end
 
-    def perform(asof_date = Time.zone.today)
+    def perform(asof_date = nil)
```
jamesjefferies: Don't default to today

PR #947, opened 2019-11-29
Filtering benchmarks and other improvements

jamesjefferies:


app/assets/javascripts/common_chart_options.js
```diff
@@ -12,7 +12,7 @@ Highcharts.setOptions({
 
 function commonChartOptions(clickListener){
   return {
-    colors: ['#97E6FC', '#007EFF', '#FCB43A', '#FF3A3B', '#50E3C2', '#FF8438'],
+    colors: ['#02B8FF', '#007EFF', '#FCB43A', '#FF3A3B', '#50E3C2', '#FF8438'],
```
jamesjefferies: The light blue colour was 'too pale'

PR #949, opened 2019-12-05
Benchmark Reports

jamesjefferies:


app/services/alerts/collate_benchmark_data.rb
```diff
@@ -3,24 +3,28 @@ class CollateBenchmarkData
...
-    def get_benchmarks_for(school, benchmarks)
-      school.latest_benchmark_results.each do |benchmark_result|
-        unless benchmarks.key?(benchmark_result.asof)
-          benchmarks[benchmark_result.asof] = { school.id => {} }
-        end
-        unless benchmarks[benchmark_result.asof].key?(school.id)
-          benchmarks[benchmark_result.asof][school.id] = {}
+    def get_benchmarks_for_latest_run(latest_school_runs, benchmarks)
+      latest_school_runs.each do |benchmark_result_school_generation_run|
+        school = benchmark_result_school_generation_run.school
```
jamesalmond: If we're just using the ID, can we not just get the `school_id` from the benchmark_result_school_generation_run and save the school load?

jamesjefferies: YES!

PR #950, opened 2019-12-05
Refactor the amr data feed config and admin view

jamesjefferies:


app/models/amr_data_feed_config.rb
```diff
@@ -2,15 +2,14 @@
 #
 # Table name: amr_data_feed_configs
 #
-#  access_type             :text             not null
```
jamesjefferies: This was just a string - have converted to an enum now in case of future editing capabilities

app/models/amr_data_feed_config.rb
```diff
@@ -2,15 +2,14 @@
...
 #  column_separator        :text             default(","), not null
 #  created_at              :datetime         not null
 #  date_format             :text             not null
 #  description             :text             not null
 #  handle_off_by_one       :boolean          default(FALSE)
 #  header_example          :text
 #  id                      :bigint(8)        not null, primary key
-#  local_bucket_path       :text             not null
+#  identifier              :text             not null
```
jamesjefferies: This now drives the local and archive paths using an instance method for each rather than persisting them...

app/models/amr_data_feed_config.rb
```diff
@@ -59,4 +65,13 @@ def header_first_thing
...
   end
+
+  def s3_archive_folder
+    "archive-#{identifier}"
+  end
+
+  def local_bucket_path
+    path = ENV['AMR_CONFIG_LOCAL_FILE_BUCKET_PATH'] || 'tmp/amr_files_bucket'
+    "#{path}/#{identifier}"
+  end
```
jamesjefferies: These replace the database columns - with added override if required for the local bucket path

db/migrate/20191205140826_add_new_source_type_remove_access_type.rb
```diff
@@ -0,0 +1,11 @@
+class AddNewSourceTypeRemoveAccessType < ActiveRecord::Migration[6.0]
+  def change
+    add_column :amr_data_feed_configs, :source_type, :integer, default: 0, null: false
+
+    AmrDataFeedConfig.all.each do |config|
```
jamesalmond: Code in migrations?

PR #953, opened 2019-12-06
Switch to upserted for amr data feed readings and nobble highland partial results

jamesjefferies:


app/services/amr/analytics_unvalidated_meter_collection_factory.rb
```diff
@@ -73,7 +73,7 @@ def add_amr_data(active_record_meter)
 
       hash_of_date_formats = AmrDataFeedConfig.pluck(:id, :date_format).to_h
 
-      AmrDataFeedReading.where(meter_id: active_record_meter.id).each do |reading|
+      AmrDataFeedReading.order(created_at: :asc).where(meter_id: active_record_meter.id).each do |reading|
```
jamesjefferies: Forcing this rather than relying on chance means that when we have multiple records for the same day (as we now do) that we pick the *last* created. These get put in a hash keyed on date, so the latest one put in the hash is the one used.

Gemfile
```diff
@@ -18,7 +18,6 @@ gem 'rack-attack'
 # Database/Data
 gem 'pg', '1.0.0' # Use postgresql as the database for Active Record
 gem 'after_party' # load data after deploy
-gem 'upsert', git: 'https://github.com/seamusabshere/upsert.git'
```
jamesjefferies: Have removed the dependency on this to use the Rails 6 native Upsert

app/services/amr/data_feed_upserter.rb
```diff
@@ -1,52 +1,28 @@
...
 
   private
 
-    def upsert_records(array_of_data_feed_reading_hashes)
-      @existing_records = AmrDataFeedReading.count
-      Upsert.batch(AmrDataFeedReading.connection, AmrDataFeedReading.table_name) do |upsert|
-        array_of_data_feed_reading_hashes.each { |row| upsert_record(upsert, row) }
+    def add_import_log_id_and_dates_to_hash
+      created_at = DateTime.now.utc
+      updated_at = DateTime.now.utc
```
jamesjefferies: We still have to add these manually - hopefully made better in a future Rails release

app/services/amr/single_read_converter.rb
```diff
@@ -29,11 +31,15 @@ def perform
...
 
-      @results_array
+      reject_any_low_reading_days
     end
 
     private
 
+    def reject_any_low_reading_days
+      @results_array.reject { |result| result[:readings].count(&:blank?) > BLANK_THRESHOLD }
+    end
```
jamesjefferies: We now reject any where we have more than a single blank record.

spec/factories/dashboard_one_day_amr_readings_factory.rb
```diff
@@ -5,8 +5,8 @@
     transient do
       association :dashboard_meter, factory: :dashboard_gas_meter
       date          { Date.yesterday }
-      kwh_data_x48  { Array.new(48, rand) }
-      one_day_kwh   { kwh_data_x48.sum }
+      kwh_data_x48  { Array.new(48, rand.to_f) }
```
jamesjefferies: This is a dashboard object and therefore should be all floaty

PR #956, opened 2019-12-09
School group summaries

jamesjefferies:


app/controllers/school_groups_controller.rb
```diff
@@ -1,6 +1,9 @@
 class SchoolGroupsController < ApplicationController
   load_and_authorize_resource
 
+  def index
```
jamesalmond: Might be nice to order the school groups

PR #961, opened 2019-12-12
Upload csv functionality

jamesjefferies:


app/models/amr_uploaded_reading.rb
```diff
@@ -0,0 +1,75 @@
...
+#
+#  fk_rails_...  (amr_data_feed_config_id => amr_data_feed_configs.id) ON DELETE => cascade
+#
+
+class AmrUploadedReading < ApplicationRecord
+  belongs_to :amr_data_feed_config
+
+  validates_presence_of :file_name
+
+  validate :validate_reading_data, on: :validate_reading_data
```
jamesjefferies: Only save it if actually valid? You can still show the errors etc...

jamesjefferies: Actually let's have a ReadingData activemodel object which we can then use anywhere....

app/controllers/admin/amr_data_feed_readings_controller.rb
```diff
@@ -0,0 +1,18 @@
+module Admin
+  class AmrDataFeedReadingsController < AdminController
+    def create
```
jamesjefferies: This does the actual insert of the data

app/services/amr/csv_parser.rb
```diff
@@ -2,13 +2,13 @@
...
-    def initialize(config, file_name)
-      @file_name = file_name
+    def initialize(config, path_and_file_name)
+      @path_and_file_name = path_and_file_name
       @config = config
     end
 
     def perform
-      CSV.read("#{@config.local_bucket_path}/#{@file_name}", col_sep: @config.column_separator, row_sep: :auto)
+      CSV.read(@path_and_file_name, col_sep: @config.column_separator, row_sep: :auto)
```
jamesjefferies: This allows us to parse temporary files which are not in the correct local bucket path

app/services/amr/csv_to_amr_reading_data.rb
```diff
@@ -0,0 +1,19 @@
+module Amr
+  class CsvToAmrReadingData
```
jamesjefferies: This replaces CSV to Reading Hash - now returns object

app/controllers/admin/amr_data_feed_readings_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+      amr_upload_reading = AmrUploadedReading.find(params[:amr_uploaded_reading_id])
+
+      amr_data_feed_import_log = AmrDataFeedImportLog.create(
+        amr_data_feed_config_id: amr_upload_reading.amr_data_feed_config_id,
+        file_name: amr_upload_reading.file_name,
+        import_time: DateTime.now.utc
+                                                            )
+      @upserted_record_count = Amr::DataFeedUpserter.new(amr_upload_reading.reading_data, amr_data_feed_import_log.id).perform
```
jamesalmond: Do we need to update the log with the count?

app/services/amr/csv_parser_and_upserter.rb
```diff
@@ -13,24 +13,17 @@ def perform
 
       amr_data_feed_import_log = AmrDataFeedImportLog.create(amr_data_feed_config_id: @config.id, file_name: @file_name, import_time: DateTime.now.utc)
 
-      @upserted_record_count = DataFeedUpserter.new(get_unique_array_of_readings, amr_data_feed_import_log.id).perform
+      amr_reading_data = CsvToAmrReadingData.new(@config, "#{@config.local_bucket_path}/#{@file_name}").perform
```
jamesalmond: Should we check the validity here?

app/controllers/admin/amr_data_feed_readings_controller.rb
jamesjefferies: Yes! Thanks for the reminder

app/services/amr/csv_parser_and_upserter.rb
jamesjefferies: Yes, we now do check the validity and update the Amr Import Log if there's a problem

PR #963, opened 2019-12-13
DB meter attributes

jamesalmond:


app/controllers/admin/schools/school_attributes_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+
+      def show
+        @meter_attribute = @school.meter_attributes.find(params[:id])
+      end
+
+      def edit
+        @meter_attribute = @school.meter_attributes.find(params[:id])
+        @meter_attribute_type = @meter_attribute.meter_attribute_type
+        authorize! :edit, @meter_attribute
```
jamesjefferies: Don't think we need explicit authorisation if inheriting from admin controller?

PR #981, opened 2019-12-30
Switch on whether user can see invisible schools

jamesjefferies:


app/models/ability.rb
```diff
@@ -27,6 +27,7 @@ def initialize(user)
 
     if user.admin?
       can :manage, :all
+      can :show, :all_schools
```
jamesalmond: I think this is inferred by the `:manage, :all` above

jamesjefferies: Doh! Sorted...

PR #985, opened 2020-01-06
Speed up benchmark report page

jamesjefferies:


app/models/benchmark_result.rb
```diff
@@ -22,7 +22,7 @@
 #
 
 class BenchmarkResult < ApplicationRecord
-  belongs_to :benchmark_result_school_generation_run
+  belongs_to :benchmark_result_school_generation_run, counter_cache: :benchmark_result_count
```
jamesjefferies: Are we ok with counter cache in 2020? It could just be a column we update when we finish, but it's a bit neater doing in this way (although almost certainly slower)

lib/tasks/deployment/20200106162015_update_benchmark_counts.rake
```diff
@@ -0,0 +1,14 @@
+namespace :after_party do
+  desc 'Deployment task: update_benchmark_counts'
+  task update_benchmark_counts: :environment do
+    puts "Running deploy task 'update_benchmark_counts'"
+
+    # Put your task implementation HERE.
```
jamesjefferies: Using counter cache makes this a piece of cake

spec/system/school_onboarding_spec.rb
```diff
@@ -218,7 +218,7 @@
 
     it 'lets the user edit inset days, meters, pupils and opening times but does not require them' do
       create :calendar_event_type, title: 'Teacher training', inset_day: true
```
jamesjefferies: Merged this fix in so that at least this PR passes on Travis!

app/models/benchmark_result.rb
jamesalmond: As long as we make all changes to the thing that's being cached via AR then it does the job.

PR #996, opened 2020-01-09
Csv import report page and email

jamesjefferies:


app/views/admin/reports/amr_data_feed_import_logs/index.html.erb
```diff
@@ -0,0 +1,48 @@
...
+        <td><%= log.error_messages %></td>
+        <td><%= log.file_name %></td>
+        <td><%= nice_date_times log.import_time %></td>
+      </tr>
+    <% end %>
+  </tbody>
+</table>
+
+<h2>Successful</h2>
+<%== pagy_bootstrap_nav(@pagy) %>
```
jamesalmond: `==`?

jamesjefferies: it is the eqiuvalent of ```raw```

PR #1012, opened 2020-01-15
Split out group users in admin user management

jamesalmond:


app/controllers/admin/users_controller.rb
```diff
@@ -0,0 +1,55 @@
+module Admin
+  class UsersController < AdminController
```
jamesalmond: This controller has just been moved into the admin namespace. It's not entirely new.

app/controllers/admin/users_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+    load_and_authorize_resource
+
+    def index
+      @school_users = @users.joins(school: :school_group).order('school_groups.name', 'schools.name', :email).group_by {|user| user.school.school_group}
+      @school_group_users = @users.where.not(school_group_id: nil).order('school_groups.name', :email).includes(:school_group)
+      @unattached_users = @users.where(school_id: nil, school_group_id: nil).order(:email)
```
jamesalmond: These lines are new

PR #1016, opened 2020-01-17
Fix reslience issues for generating charts, content batch

jamesjefferies:


app/controllers/benchmarks_controller.rb
```diff
@@ -92,7 +92,7 @@ def benchmark_results
     @benchmark_results = Alerts::CollateBenchmarkData.new.perform(schools)
   end
 
-  def content_manager(date = Time.zone.today)
+  def content_manager(date = BenchmarkResultGenerationRun.latest_run_date)
```
jamesalmond: I think we should somehow be getting access to the benchmark result generation run used above here to get the right date out, otherwise we've got the run selection logic in 2 places.

PR #1033, opened 2020-01-23
0.61.3 - Add structured content methods for Carbon Accordion

jamesjefferies:


app/views/schools/analysis/show.html.erb
```diff
@@ -1,5 +1,29 @@
 <% content_for :page_title, @title %>
-<%= link_to 'Back', school_analysis_index_path(@school), class: 'btn btn-rounded mt-3' %>
-<% @content.each do |content| %>
-  <%= render content[:type].to_s, content: content[:content], school: @school %>
+<%= link_to 'Back', school_analysis_index_path(@school), class: 'btn btn-rounded mt-3 mb-2' %>
+
+<% if @structured_content %>
+  <div class="accordion" id="accordionExample">
```
jamesalmond: Better id?

app/services/alerts/adapters/analytics_adapter.rb
```diff
@@ -20,6 +20,16 @@ def benchmark_dates
         analysis_object.benchmark_dates(@analysis_date)
       end
 
+      def has_structured_content?
```
jamesalmond: Should we add these methods to the system adapter too?

app/views/schools/analysis/show.html.erb
jamesjefferies: Yes, sure

app/services/alerts/adapters/analytics_adapter.rb
jamesjefferies: Yes, I think we should

PR #1056, opened 2020-01-29
Correct units field on GDST schools

jamesalmond:


lib/tasks/deployment/20200129111158_update_gdst_units_field.rake
```diff
@@ -0,0 +1,13 @@
+namespace :after_party do
+  desc 'Deployment task: update_gdst_units_field'
+  task update_gdst_units_field: :environment do
+    puts "Running deploy task 'update_gdst_units_field'"
+
+    AmrDataFeedConfig.where(identifier: ['gdst-historic-electricity','gdst-electricity']).update_all(units_field: 'Data Type')
```
jamesalmond: It was set to 'kWh' on the historic feed

PR #1116, opened 2020-06-11
Developer guide tweaks

jamesjefferies:


config/secrets.yml
```diff
@@ -12,7 +12,7 @@
 
 development:
   secret_key_base: 4ab2fe0e211da787f17c1691aa39753ea90555dd1d99b27793ef30c9fd65663fda6f2292b1269695e1c4fff595c75b3631894ed1927c3926f2945e1bc896705b
-  encrypted_field_salt: <%= ENV["ENCRYPTED_FIELD_SALT"] %>
+  encrypted_field_salt: <%= ENV["ENCRYPTED_FIELD_SALT"] || "vinegar" %>
```
jamesjefferies, line 15: Provide a default here so the application *can* then run without a ```.env``` file

config/environments/development.rb
```diff
@@ -31,7 +31,9 @@
 
   # Store uploaded files on the local file system (see config/storage.yml for options)
   config.active_storage.service = :local
-  config.active_storage.service = :amazon
+
```
jamesjefferies, line 34: Don't default to amazon as you then need to have amazon credentials to run.

PR #1199, opened 2021-01-19
Add credentials for test db

ldodds:
Add credentials for test database.

On ubuntu at least, postgres needs to be configured to allow local connections for postgres user and recommended approach is to set a password for that.

So I needed to specify these for my environment to work.

Expect travis may need the default values populating...?

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -14,6 +14,9 @@ test: &test
   <<: *default
   host: localhost
   database: energy_sparks_test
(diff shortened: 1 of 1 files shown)


jhigman:
Looks good to me - tests run OK for me locally, and Travis has it's own database.travis.yml file anyway, so isn't affected. Happy for it to be merged down.

PR #1200, opened 2021-01-25
Add new Solar for Schools config

ldodds:
New Solar for Schools config

- update meter column
- revise data format
- update reading fields and header example to match sample file.


jhigman:
Looks good 👍

PR #1201, opened 2021-01-25
Add EDF Historic feed, using latest format

ldodds:
Add a new `edf-historic` feed to be used instead of the "Highlands" format.

Uses latest configuration based on sample provided.


jhigman:
Looks good 👍

PR #1202, opened 2021-01-25
Streaming resources via stable URL

ldodds:
Spiked implementation of streaming resources via the app. This is still a WIP for discussion.

* Adds new route `/resources/:serve/:id` to stream resource files from active storage. Use `/resources/inline/:id` for inline display, `/resources/attachment/:id` to force a download
* URLs use id of the resource file which will be stable over time, even if the attached files are changed or updated
* Uploaded file names are preserved as now and these are used for downloads, so you'll get whatever name is provided by the ES admin

Produces urls like: `test.energysparks/resources/attachment/18`

Files should get caught by the CDN I think (although I haven't check the config) and cached.

@jhigman do you think this is a sensible option? It preserves all uploaded content so quick to implement and apply.

I can add a simple route for case studies too.

### DIFF
--- a/app/controllers/case_studies_controller.rb
+++ b/app/controllers/case_studies_controller.rb
@@ -3,4 +3,13 @@ class CaseStudiesController < ApplicationController
   def index
     @case_studies = CaseStudy.order(position: :asc)
   end
+
+  def download
+    resource = CaseStudy.find_by(id: params[:id])
+    if resource.present?
+      serve_from_storage(resource.file, params[:serve])
+    else
+      render file: Rails.public_path.join('404.html'), status: :not_found, layout: false
+    end
+  end
 end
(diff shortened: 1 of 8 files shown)


jhigman:
Looks good. 

Just wondering whether the route should be slightly more "standard", something like /resources/:id/inline and /resources/:id/attachment (or inline and download, even) which is more like what the rails routing usually expects..? (slightly fewer things parameterised in the routes also, which I find avoids issues later). Happy to discuss.

jhigman:
@ldodds Looks like the build is failing on the resources_spec test 

Failure/Error: expect(page.has_link? "", href: "/resources/1/download").to be true
       expected true
            got false

The resource id is dynamic for the other tests (#{@resource.id}) so maybe needs to be for this one, too..?

ldodds:
Yep, missed that one. Will push fix.

ldodds:
Remaining task before this can be pushed to test:

* [ ] Change how case study urls are generated to serve via application
* [ ] Add test for serving of case study page plus download links

ldodds:
@jhigman this has been signed off on test. OK to merge down?

jhigman:
Yes, go ahead.

Cheers, Julian

PR #1203, opened 2021-01-25
Alert types for targeting and tracking

jhigman:


jhigman:
@ldodds This is an after party task that creates the new alert types for targeting and tracking that Philip is introducing - they're already on test, but this needs to be merged to master (and then can be part of the release tomorrow).

ldodds:
Looks OK to me :+1:

ldodds (approved):

PR #1206, opened 2021-01-26
Changed to before :each

ldodds:
@jhigman I was able to get the test to reliably fail locally using an rspec seed. 

This change seems to resolve the issue consistently.


jhigman:
Great 👍

PR #1208, opened 2021-01-27
Meteostat Integration

ldodds:
Initial branch for adding support for Meteostat.

Currently includes:

* [x] new models for `WeatherStation` and `WeatherObservation` (ultimately to replace the dark sky stuff)
* [x] new data loader, using the `MeteoStat` interface in the analytics package
* [x] admin interface to create and manage weather stations
* [x] ability for admin to download CSV files of data for each weather station

Still to do around this bit of the code:

* [x] associate weather stations with schools
* [x] association of weather stations with school groups and onboarding?
* [x] decide when to automatically add stations and associate them with schools
* [x] whether to allow lat/long to be changed, or prefer switching to new station
* [x] switching between Dark Sky and MeteoStat if a schools is associate with an active Weather Station, so right readings are passed to analytics
* [x] decide how to handle/generalise current `Area` stuff (replace in favour of stations?)
* [x] updates to EB config to add new cron job for Meteostat loading, as well as env variables to drive it

Separately need to decide how to handle Dark Sky migration, historical loading, and any improvements to the `MeteoStat` API client.

### DIFF
--- a/.ebextensions/cronjob.config
+++ b/.ebextensions/cronjob.config
@@ -23,6 +23,13 @@ files:
         content: |
             45 3 * * * root /bin/bash -l -c 'cd /var/app/current && RAILS_ENV=production bin/run_as_webapp bundle exec rake data_feeds:dark_sky_temperature_loader >> log/jobs.log 2>&1'
 
+    "/etc/cron.d/load-meteostat-data":
+        mode: "000644"
+        owner: root
+        group: root
+        content: |
+            45 3 * * * root /bin/bash -l -c 'cd /var/app/current && RAILS_ENV=production bin/run_as_webapp bundle exec rake data_feeds:meteostat_loader >> log/jobs.log 2>&1'
+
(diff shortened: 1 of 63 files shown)


ldodds:
@jhigman I think this is ready for review.

The functionality thats not currently included is:

* a crontab entry to run the loading of historical data from Meteostat. That could be added and deployed separately
* attribution of data sources, but this isn't Meteostat specific, so suggest we do as a separate branch
* any changes to the way that the client library is called, pending any refactoring. But again this could be done separately.

Deploying these latest updates to test won't trigger use of Meteostat data in the analytics until the feature flag is set to activate it.

That means I also think its safe, pending any feedback from review, to merge down and release to both test and live. Without any stations configured, and without the feature flag, none of the new code should run.

jhigman:
👍  all looks good

ldodds:
I need to make a change to allow the Dark Sky Areas to be deselected, so will do that before merging down.

PR #1209, opened 2021-01-28
If logged in, schools link goes to group, otherwise full UK

jhigman:


jhigman:
@ldodds this is a change to the navigation, so that the Schools link on the main menu goes to the group for the logged in user (usually the group that their school is in, or - for group admins - the group that they're an admin for). Not logged in users go to the full schools page as before.

ldodds:
@jhigman This looks like a logged in user can't navigate to the overall list of schools, as the link takes them to their group. Don't we want users to be able to see the wider map as well? I couldn't think of another way for them to get there.

jhigman:
@ldodds When you're on the school group view, there's always a "View all schools" button on the page, so you can go to the UK-wide view..

ldodds:
OK. I see, yep. :+1:

ldodds (approved):

PR #1210, opened 2021-01-28
Destroy any owned simulations on delete

jhigman:


jhigman:
@ldodds I think it was only simulations preventing users being deleted (in this instance, anyway)

ldodds:
:+1:

ldodds (approved):

PR #1214, opened 2021-01-29
Automatically generate stations from DSA

ldodds:
AfterParty task to create new Meteostat weather stations for existing Dark Sky Areas.

Doesn't regenerate if there's a station with same title, lat, long already.


jhigman:
👍

PR #1217, opened 2021-02-01
Allow lat/long to be edited

ldodds:
Allow admin to edit lat/long of a weather station. Remove old observations if they do.


jhigman:
👍

PR #1219, opened 2021-02-02
N3rgy data loader

ldodds:
Initial integration of N3rgy client library in the analytics package into the main application.

To test locally you'll need to:

* bundle (to pull in the branch of the analytics library)
* run migrations
* run after party

Currently this has:

* rake task to load data for all meters that are setup in DCC and where consent has been granted. We have no meters of that type currently, as we've not done the consent process yet
* supporting code to invoke the N3rgy API client and upsert any new readings

Things still to do include (but not limited to!):

* [x] implementing some better logic around start/end dates. At present the rake task only passes through what is provided. We should be applying a useful default which is at least the last 24 hours, but might be a date range based on what recent readings we have for the individual meter (if any)
* [x] deciding what should be done with the `missing_readings` that are returned from the API calls. Should these be logged somewhere as warnings?
* [x] tests that run against the sandbox API to do a proper end to end test, to flush out any problems with meter ids, date formats, etc
* [x] exception handling, as the N3RGY client can throw exception if its unhappy with what it gets from the API responses


jhigman:
👍  looks good

ldodds:
Latest updates:

* switch to use the latest interface from analytics
* tweak the start/end date windows to better match what we want to load
* when viewing a meter, now displays DCC related configuration and status code from API
* when adding a new meter, if its supposed to be a DCC meter, we check whether its known. So can't add unknown meters
* when updating a meter, can't activate it as a DCC meter if its not known.

app/controllers/schools/meters_controller.rb
```diff
@@ -15,15 +15,23 @@ def index
     end
 
     def show
+      @n3rgy_status = MeterManagement.new(@meter).check_n3rgy_status
```
jhigman, line 18: There is potential for the status call to raise an error if the API returns something unexpected - could handle in the API, or handle here? (or in the meter management)

PR #1221, opened 2021-02-08
First implementation of user help videos

ldodds:
* Admin functionality to create, edit, delete videos, specifying their order and also whether they are to be featured
* New "user guide videos" link in navbar to show videos
* Home page now displays any of the featured videos in the system
* If no featured videos are available, as the moment it defaults to the current video on the homepage (only)


jhigman:
👍  looks good

PR #1230, opened 2021-02-12
WIP Added delete button to alert rating form

jhigman:


jhigman:
This change needs more thought - there are associated objects in the analysis runs that would also need to be deleted, leaving partial analysis history. Might be better to make alert ratings inactive instead.

PR #1235, opened 2021-02-18
First iteration of code to generate emails from custom RQL reports

ldodds:
We want to produce a more detailed reports about failures to validate API readings, listing individual meters. Doing this will require some changes to how the application generates, captures and log errors around the analytics code.

In order to report on failures we need to keep track of when and where they occur. Rollbar already does this for us. It also allows us to provide additional metadata when we record errors. For example, we currently log school ids, names and other information in a few places. That information can be queried using [the RQL query language](https://docs.rollbar.com/docs/rql).

So, rather than create a new set of tables to manage logging of errors, we can use Rollbar to capture the issues. This will allow us to trap all errors, including unforseen problems, not just those we've anticipated.

This PR has an initial implementation of a daily process that will run a set of pre-configured reports against [the Rollbar API](https://explorer.docs.rollbar.com/#tag/RQL) then generate an email containing the results of each report. 

Each report has a title, description, an RQL query and a set of results (if it was executed successfully). The format of the queries can vary and the template displays all columns and rows. It also introduces links to Rollbar to allow more fine-grained view of errors.

* [x] Implement basic Rollbar client to executing RQL queries
* [x] Integrate the above with a simple service class that runs a set of queries and sends the results as an email
* [x] Tidy up uses of Rollbar across the application to add useful metadata as tags and/or by using [the scope](https://docs.rollbar.com/docs/ruby#the-scope) and [context](https://docs.rollbar.com/docs/ruby#error-contexts) functionality (the latter is already being applied automatically in Rails controllers)
* [x] Add new exception handling into the validation code in the analytics package to log individual meter validation errors, so they can be queried using RQL
* [x] Write initial RQL query to report on meter validation errors 
* [x] Add additional RQL queries to help to surface other problem areas

### DIFF
--- a/lib/rollbar_api/rql_jobs.rb
+++ b/lib/rollbar_api/rql_jobs.rb
@@ -11,7 +11,7 @@ module RollbarAPI
       @client = if client != nil
         client
                 else
-        Faraday.new(API_BASE)
+        Faraday.new(API_BASE, request: { timeout: 20 })
                 end
       raise "ROLLBAR_READ_ACCESS_TOKEN not configured" unless api_token
       @api_token = api_token
--- a/lib/rollbar_api/rql_jobs.rb
+++ b/lib/rollbar_api/rql_jobs.rb
@@ -1,0 +1,64 @@
+require 'json'
+require 'faraday'
+
+module RollbarAPI
+  API_BASE = "https://api.rollbar.com".freeze
+  FAILED_STATES = %w[failed cancelled timed_out].freeze
+  END_STATES = FAILED_STATES + ["success"]
+
+  class RQLJobs
+    def initialize(api_token = ENV['ROLLBAR_READ_ACCESS_TOKEN'], client = nil)
+      @client = if client != nil
+        client
+                else
+        Faraday.new(API_BASE)
+                end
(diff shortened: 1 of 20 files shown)


lib/rollbar_api/rql_jobs.rb
```diff
@@ -0,0 +1,64 @@
+require 'json'
+require 'faraday'
+
+module RollbarAPI
+  API_BASE = "https://api.rollbar.com".freeze
+  FAILED_STATES = %w[failed cancelled timed_out].freeze
+  END_STATES = FAILED_STATES + ["success"]
+
+  class RQLJobs
+    def initialize(api_token = ENV['ROLLBAR_READ_ACCESS_TOKEN'], client = nil)
```
jhigman, line 10: I was wondering if we could move away from the pattern of specifying env vars in initializers.. the RollbarNotifierService may not be able to be instantiated without params, as it calls this initializer, which then looks for env vars..
Maybe we could always pass in those bits of config, and only fetch them from env vars at the very top level?
(I think I've done the same elsewhere with the n3rgy, but planning to revisit that..)

jhigman (approved):

PR #1253, opened 2021-03-01
Send email to multiple users

ldodds:
Send the activation email to all school admins, staff and group users.

Sends a single email with everyone in the to: field


jhigman (approved):
👍

PR #1257, opened 2021-03-02
Correctly assign weather and solar observations when stations not set

ldodds:
Test for presence of MeteoStat and Dark Sky stations.

Also handle missing solar pv area.

Add test for the single page(?) that displays these values if school is not in cache.

### DIFF
--- a/app/controllers/concerns/school_aggregation.rb
+++ b/app/controllers/concerns/school_aggregation.rb
@@ -14,8 +14,8 @@ private
   def setup_loading_stats
     @number_of_meter_readings = @school.amr_validated_readings.count * 48
     @number_of_meters = @school.meters.count
-    @number_of_weather_readings = DataFeeds::DarkSkyTemperatureReading.where(area_id: @school.dark_sky_area.id).count * 48
-    @number_of_solar_pv_readings = DataFeeds::SolarPvTuosReading.where(area_id: @school.solar_pv_tuos_area.id).count * 48
+    @number_of_weather_readings = number_of_weather_readings
+    @number_of_solar_pv_readings = number_of_solar_readings
   end
 
   def aggregate_school_service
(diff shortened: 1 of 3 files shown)


spec/controllers/teachers/school_controller_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  describe "GET #show" do
+
+    context "school not in cache" do
+        before(:each) do
+          allow(AggregateSchoolService).to receive(:caching_off?).and_return(false, true)
+          allow_any_instance_of(AggregateSchoolService).to receive(:aggregate_school).and_return(school)
+          sign_in_user(:school_admin, school.id)
+        end
+
+        it "shows error page" do
```
jhigman, line 16: Does this show an error page, or show the normal page with 0 readings..?

app/controllers/concerns/school_aggregation.rb
```diff
@@ -25,4 +25,21 @@ def aggregate_school_service
...
+
+  def number_of_solar_readings
+    if @school.solar_pv_tuos_area.present?
+      DataFeeds::SolarPvTuosReading.where(area_id: @school.solar_pv_tuos_area.id).count * 48
+    end
+    0
+  end
+
+  def number_of_weather_readings
+    if @school.weather_station.present?
```
jhigman, line 37: If the school has both weather station and dark sky area, should we sum the counts (we'll be using the dark sky data for historic dates)?

jhigman (approved):

spec/controllers/teachers/school_controller_spec.rb
ldodds: This isn't the show page, its a separate template that catches situation when school isn't in cache.

PR #1259, opened 2021-03-02
Add check to see if onboarding user is nil

ldodds:
If a school is setup manually, it won't have an "created user" associated with the onboarding. But may have other users.

This fixes up error when sending the activation email by checking if the user is nil.


jhigman:
👍

PR #1266, opened 2021-03-08
Rake task to create meter attributes

jhigman:


ldodds:
As discussed, suggest adding a task to back out the changes. This means we can also remove and reapply. 

The code also does a `create!` (which might error) rather than a find & then update which means you can't run the script multiple times without it adding more attributes?

ldodds:
:+1:

PR #1269, opened 2021-03-11
User display name when creating onboarding contact

ldodds:
Users don't have to have named, but Contacts to.

During onboarding we create Contacts from Users. So it can fail if a name isn't set (or is empty)

So use `display_name` instead.

Also ensure `display_name` returns a useful value if name is empty.


jhigman (approved):
👍

PR #1274, opened 2021-03-15
Allow admin to search for consent statement by ref or mpxn

ldodds:


app/controllers/admin/consent_grants_controller.rb
```diff
@@ -3,10 +3,30 @@ class ConsentGrantsController < AdminController
...
+        if search["mpxn"].present?
+          meter = Meter.find_by_mpan_mprn(search["mpxn"])
+          if meter.present?
+            return ConsentGrant.where(school: meter.school).by_date
+          else
+            return []
+          end
+        end
+      end
+      return ConsentGrant.by_date
```
jhigman, line 29: "return" statements at end of methods generally unnecessary and frowned upon in idiomatic Ruby..! :-)

app/views/admin/reports/index.html.erb
```diff
@@ -9,6 +9,12 @@
   <li><%= link_to "Meter Report with deprecated meters included", admin_reports_amr_validated_readings_path(all_meters: true) %></li>
 </ul>
 
+<h2>Users</h2>
+
+<ul>
+  <li><%= link_to "Consents Granted", admin_consent_grants_path %></li>
```
jhigman, line 15: We already have this link from the admin "Admin" page - do we need it in both places..?

jhigman: (Maybe remove from other page..?)

jhigman (approved):

PR #1278, opened 2021-03-18
Check api for meters and update status and timestamps

jhigman:


lib/tasks/meters/dcc_meter_checker.rake
```diff
@@ -0,0 +1,7 @@
+namespace :meters do
+  desc 'Check which meters exist in the DCC'
+  task :check_for_dcc => :environment do |_t, args|
+    meters = Meter.where.not(dcc_meter: true).where(dcc_checked_at: nil)
```
ldodds, line 4: Suggest restricting the where clause further to only query real meters (so remove pseudo/sub-meters) and only gas/electricity as I'm not sure they have solar.

ldodds: Also, we have two requirements: one is to check newly setup meters against DCC and another for older meters. So should there be two tasks, e.g. one for meters recently created in last 24 hours or so, and another for older ones?

ldodds:
:+1:

PR #1281, opened 2021-03-23
Call trusted consent api after review and from rake task

jhigman:


app/services/meters/dcc_grant_trusted_consents.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def perform
+      @meters.each do |meter|
+        begin
+          reference = meter.meter_review.consent_grant.guid
+          @n3rgy_api_factory.consent_api(meter).grant_trusted_consent(meter.mpan_mprn, reference)
+          meter.update(consent_granted: true)
+        rescue => e
+          Rails.logger.error("#{e.message} for mpxn #{meter.mpan_mprn}")
+          Rollbar.error(e, job: :dcc_grant_trusted_consents, mpxn: meter.mpan_mprn)
```
ldodds, line 16: Might be worth adding the school in the rollbar error here too?

jhigman: 👍

PR #1282, opened 2021-03-23
1323 allow admin to request school submits an energy bill

ldodds:
Form to allow admin to request bill from school, selecting where to send the email and which meters to mention.

Screenshot from 2021-03-23 16-45-21

### DIFF
--- a/app/services/schools/bill_request_service.rb
+++ b/app/services/schools/bill_request_service.rb
@@ -1,0 +1,18 @@
+module Schools
+  class BillRequestService
+    def initialize(school)
+      @school = school
+    end
+
+    def users
+      users = @school.users.school_admin + @school.users.staff
+      users.sort_by(&:staff_role)
+    end
+
+    def request_documentation!(users, meters = [])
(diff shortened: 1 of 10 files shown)


app/services/schools/bill_request_service.rb
```diff
@@ -0,0 +1,18 @@
...
+      @school = school
+    end
+
+    def users
+      users = @school.users.school_admin + @school.users.staff
+      users.sort_by(&:staff_role)
+    end
+
+    def request_documentation!(users, meters = [])
+      electricity_meters = meters.select {|m| m.meter_type.to_s == "electricity"}
```
jhigman, line 13: I think you can use the enum types as a boolean to select meters, e.g.
electricity_meters = meters.select(&:electricity?)
gas_meters = meters.select(&:gas?)

jhigman (approved):
Looks good. 

Wondering if we should CC the person sending off the request for a bill..? (we don't have any record in the system of the email that's sent - though could possibly retrieve from mailgun..?)

jhigman:
👍

PR #1288, opened 2021-03-25
Don't validate against n3rgy when saving meter but do show status on …

jhigman:
…edit form

### DIFF
--- a/app/views/schools/meters/_form.html.erb
+++ b/app/views/schools/meters/_form.html.erb
@@ -26,14 +26,14 @@
 
   <div class="form-group">
 
-    <p class="alert alert-warning">Experimental features for N3RGY integration</p>
+    <p class="alert alert-warning">Experimental features for n3rgy integration</p>
 
     <% if show_dcc_fields %>
 
       <% if meter_is_dcc %>
-        <p class="alert alert-success">This meter is available in the DCC</p>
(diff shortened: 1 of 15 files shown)


app/views/schools/meters/_form.html.erb
```diff
@@ -23,24 +23,38 @@
...
-    </div>
-    <div class="custom-control custom-checkbox">
-      <%= form.check_box :consent_granted, class: 'custom-control-input'  %>
-      <%= form.label :consent_granted, 'Consent granted', class: "custom-control-label"  %>
-    </div>
+
+    <% if show_dcc_fields %>
+
+      <% if meter_is_dcc %>
+        <p class="alert alert-success">This meter is available in the DCC</p>
```
ldodds, line 34: Suggest a wording tweak here. Based on what we know now, I think "available in DCC" is basically anything where we can do a Read Inventory. This status call is now showing "available via N3RGY". Might make it clearer about about we are showing?

app/views/schools/meters/_form.html.erb
```diff
@@ -23,24 +23,38 @@
...
+      <% else %>
+        <p class="alert alert-warning">This meter is not available in the DCC</p>
```
ldodds, line 36: See previous comment

ldodds:
:+1:

PR #1303, opened 2021-04-09
1365 implement non public schools

ldodds:
* Add rspec for the abilities, as its getting complex. currently just covers dashboard
* Add system specs to check for access to various dashboards
* Add controller and page to prompt people to login when viewing non-public schools


jhigman (approved):
👍

PR #1366, opened 2021-06-22
1521 actual meter attributes

jhigman:


app/views/schools/user_tariffs/index.html.erb
```diff
@@ -37,6 +37,7 @@
             </div>
           <%# end %>
           <div class="padded-row">
+            <%= link_to "View as JSON", school_user_tariff_path(@school, user_tariff, format: :json), class: 'btn', target: '_blank' %>
```
ldodds, line 40: We wouldn't want this here long term. A user doesn't need to see this structure. Could we wrap it in a check to see if current user is an admin? Want to avoid having stuff on the user page which might impact the tests.

Ideally, we'd include this output dump on the existing meter attributes page as we did for the DCC stuff, so it can be seen alongside the others?

PR #1419, opened 2021-07-20
1590 create application page to display targets

jhigman:


ldodds:
@jhigman could you call this controller the Progress Controller, so its accessible via `/schools/:id/progress/:type`?

Then I can use School::TargetsController for the forms and pages for setting, reviewing and editing targets.

ldodds (approved):

PR #1432, opened 2021-08-09
Upgrade to 1.5.0 of did_you_mean

ldodds:
This fixes a long standing bug where attempting to log a NoMethodError triggers the script or server process to hang because its in an infinite loop. Any attempt to log a NoMethodError when calling an object method in my development server triggered the issue. Previously this has been observed on the test server.

Debugging with gdb shows that `did_you_mean` was the cause. Upgrading resolves the issue.

There's also a change to the Travis config to try and [work around this issue](https://github.com/ruby/did_you_mean/issues/117). Explicitly running bundle install with did you mean disabled ensures the right versions of the gems are used for later scripts.

This should all be resolved in Ruby 2.7 later which has Did You Mean 1.5.0 by default I believe.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ addons:
     packages:
     - google-chrome-stable
 sudo: required
+before_install:
+  RUBYOPT='--disable-did_you_mean' bundle install
 before_script:
 - cp config/database.travis.yml config/database.yml
 script:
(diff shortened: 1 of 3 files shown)


ldodds:
I've included the gem just in `development` but it should probably be installed by default. Was just being cautious initially.

PR #1546, opened 2021-10-04
Live data page

jhigman:


ldodds (approved):
Looks good. You have a typo in the _guage.html.erb file name, should be _gauge.html.erb

PR #1550, opened 2021-10-05
1761 implement basic live data gauge

jhigman:


ldodds (approved):
Left some comments in Slack, re: frequency of API calls for login and fast update. But otherwise good to merge as first iteration.

PR #1582, opened 2021-11-01
Cache token for 45 minutes

jhigman:


app/services/cads/live_data_service.rb
```diff
@@ -25,11 +27,13 @@ def power_for_type(powers, type)
...
 
     def api
-      unless @api
-        @api = MeterReadingsFeeds::GeoApi.new(username: ENV['GEO_API_USERNAME'], password: ENV['GEO_API_PASSWORD'])
-        @api.login
+      @api ||= MeterReadingsFeeds::GeoApi.new(token: token)
+    end
+
+    def token
+      Rails.cache.fetch(CACHE_KEY, expires_in: 45.minutes) do
```
ldodds, line 34: Can we make the cache timeout an environment variable here, so its easier to change if they change their API credentials.

app/services/cads/live_data_service.rb
```diff
@@ -25,11 +27,13 @@ def power_for_type(powers, type)
     end
 
     def api
```
ldodds, line 29: We need some error handling in here to report API errors in Rollbar, even if they're not displayed to the user. I'd like to know if we've got authorisation or server errors so they can be caught when we're in production. This will help us debug issues with setup.

ldodds (changes requested):

app/services/cads/live_data_service.rb
ldodds: If we get an auth error, then probably best to also invalidate the Rails cache to at least prompt another login attempt? Then if the token has expired then system will make some attempt to recover?

ldodds: And in the Rollbar logging can we add context so we know which school, device id, etc?

ldodds (approved):

PR #1588, opened 2021-11-02
1826 add data enabled flag

jhigman:


app/models/activity_category.rb
```diff
@@ -6,6 +6,7 @@
 #  description :string
 #  featured    :boolean          default(FALSE)
 #  id          :bigint(8)        not null, primary key
+#  live_data   :boolean          default(FALSE)
```
ldodds, line 9: Needs to be true initially, until we're ready to release the full feature.

db/schema.rb
```diff
@@ -494,7 +494,7 @@
     t.bigint "school_id", null: false
     t.string "name", null: false
     t.string "device_identifier", null: false
-    t.boolean "active", default: false
+    t.boolean "active", default: true
```
ldodds, line 497: Is this an issue with your local schema? A CAD device should be inactive by default I think?

ldodds (approved):

ldodds (approved):

PR #1601, opened 2021-11-09
Dont aggregate school if we're not showing data enabled features

ldodds:
If we're not showing data enabled features, then by-pass the check for whether aggregated school is cached.

This also removes the `DashboardSetup` concern and puts the single function in the `SchoolAggregation` concern. While the function isn't specifically about aggregation, the same basic check is needed in other controllers that do also carry out aggregation. So it felt better to combine them rather than introduce a dependency between two concerns? For other bits of the system, checking the data-enabled flag directly is clearer than importing the concern and then not using most of the functions. What do you think @jhigman?


jhigman (approved):

jhigman:
Yes seems sensible approach.

PR #1630, opened 2021-11-29
Format management summary table from data

jhigman:


ldodds (commented):
Looks good but have added a checklist of changes to Trello, so some more tweaks needed prior to merging.

PR #1721, opened 2022-03-15
Ensure that progress report handles missing data

ldodds:
Adds tests to ensure that if we have some months of missing data for actual consumption or targets, then the progress report still displays. This might happen if we only start getting enough data within the target period -- this might be the case if a school's gas/electricity periods don't align. And we're also planning to allow them to set targets at any time, which means that reports might be incomplete, e.g. in their first year.

So this is mostly adding tests to ensure that the current functionality won't error. And adding some comments to the page to explain why some months in the tables aren't populated.

### DIFF
--- a/app/controllers/schools/progress_controller.rb
+++ b/app/controllers/schools/progress_controller.rb
@@ -29,6 +29,12 @@ module Schools
 
     private
 
+    def summary_table
+      if EnergySparks::FeatureFlags.active?(:use_management_data)
+        Schools::ManagementTableService.new(@school).management_data
+      end
+    end
+
     def index_for(fuel_type)
       @fuel_type = fuel_type
       @current_target = @school.current_target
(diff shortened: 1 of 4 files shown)


app/controllers/schools/progress_controller.rb
```diff
@@ -55,6 +64,17 @@ def show_storage_heater_notes(school, fuel_type)
       fuel_type == :electricity && school.has_storage_heaters?
     end
 
+    def partial_consumption_data?(progress)
+      #this could be more sophisticated, relies on ordering of months array reflecting
+      #actual starting month of the target reporting period
+      first_month = progress.months.first
+      progress.monthly_usage_kwh[first_month].nil? || progress.partial_months[first_month] == true
```
jhigman, line 71: Probably don't need " == true" here? (and makes it harder to parse the evaluation, with " || " earlier in the line..?)

ldodds: Yep, have changed

jhigman (approved):

PR #1743, opened 2022-03-22
Admin allow delete targets estimates

ldodds:


spec/system/schools/estimated_annual_consumptions_spec.rb
```diff
@@ -85,6 +85,10 @@
         latest_estimate = school.latest_annual_estimate
         expect(latest_estimate.electricity).to eq 6000.0
       end
+
+      it 'does not show delete link' do
+        expect(page).to_not have_content("Delete")
```
jhigman, line 90: Would it be more expressive to use "..to_not have_link.." rather than "..to_not have_content.." ?

spec/system/schools/school_targets_spec.rb
```diff
@@ -295,6 +295,11 @@
         expect(school.current_target.storage_heaters).to eql 7.0
       end
 
+      it "doesnt show delete button" do
+        click_on "Revise your target"
+        expect(page).to_not have_content("Delete")
```
jhigman, line 300: Same as above..?

jhigman (approved):

PR #1753, opened 2022-03-25
2029 Create Transport Types Admin Pages

urbanwide:
Admin pages for transport types 
Just putting this here for visibility. 

- [x] index
- [x] show
- [x] edit / update
- [x] new / create
- [x] delete


ldodds (commented):
Could you add a link to the new admin form from the `admin/index.html.erb` view so the page can be accessed by an admin?

urbanwide:
Thank you, I'll add it in now

PR #1781, opened 2022-04-11
EfT Solar data feed config and processing

jhigman:


app/services/amr/data_feed_translator.rb
```diff
@@ -1,43 +1,54 @@
 module Amr
+  class DataFeedException < RuntimeError
```
ldodds, line 2: Duplicate of the new separate class file definition?

jhigman: Good point, fixed

ldodds (approved):

PR #1786, opened 2022-04-19
Calendar event has reference to original event on parent calendar

jhigman:


app/services/calendar_resync_service.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def resync
+    parent_event_ids = @calendar.calendar_events.map(&:id)
+    sync_events = calendar_events_to_sync(@calendar, @from_date)
+    @calendar.calendars.each do |child_calendar|
+      child_calendar.transaction do
+        delete_orphaned_child_events(child_calendar, parent_event_ids)
+        resync_child_events(child_calendar, sync_events)
+      end
+      if child_calendar.calendars.any?
```
ldodds, line 15: Should we only do this if the transaction to apply events to the child calendar was successful? Is there a way to determine that?

PR #1807, opened 2022-05-11
2032 transport survey UI

ldodds:
Draft PR from @urbanwide


ldodds:
@urbanwide I'm happy for this to be merged down in its current form. There's a system test for the core functionality which exercises the basis. The Javascript looks well factored for the moment.

So I suggest we merge and then do further smaller PRs for additional improvements.

One thing I did notice is that the JS is using alerts quite a bit for feedback. I think it *might* be better to add messages to the modal dialog or the page, rather than as an alert, as then they can be styled and might be simpler for the user. Otherwise they have to dismiss those messages and worth with the modal. That could be done separately.

urbanwide:
Great, thanks for reviewing @ldodds !

I think you have a good point re: the alerts - I'll pop it in a separate trello card.

PR #1816, opened 2022-05-17
Transport surveys UI - change passenger labels

urbanwide:
Another small pull request - this time to update the passenger labels in the transport surveys app.

Change passenger label to "Me + 1 more" etc here, rather than "Just Me + 1 more."

The build is currently failing on a test that looks like is from elsewhere.


ldodds (approved):

ldodds:
We've had some flickering tests in the past, but I don't recall seeing this one. If it occurs again we can investigate further.

PR #1820, opened 2022-05-20
Transport Survey Alerts

urbanwide:
- Rather than using js alerts, use bootstrap ones. Unless alerts are fatal / blocking, they fade after a period. See what you think, the alternative would be to allow them to be dismissed rather than fading.
- Spruce up the 'remove and save unsaved responses' styling and functionality on the main page. Have re-styled the buttons - couldn't find any working button css that did primary / secondary styling, so I improvised with a tiny bit of custom css and along with other standard bootstrap styles.
- Also remove the "reset and clear" functionality from the app (https://trello.com/c/zr26HkFa/2181-transport-survey-ui-remove-ability-to-reset-and-clear-results) as it relied on changes in this PR.


ldodds (approved):
Changes look good. I've not reviewed the UX, might be best to do that after we've completed the latest round of changes based on Claudia's feedback. But good to merge.

PR #1823, opened 2022-05-23
Add category to transport type

urbanwide:
Quick question before this is merged. I have put "motorbike" in the "car" category but this doesn't seem totally right. Or do we need to rename / broaden the "car" category to "motorised" or similar? I have put "park and stride" in the "car" category too?

Summary of changes:
* Add new enum field to transport types
* Make changes to admin interface to support new field + associated rspec
* This includes an after party task to assign a category to each transport type
* This also changes the image for the electric car transport type to be "🔌🚘" as requested by Claudia
* While I was doing this, I noticed an extra column had sneaked into the transport_survey_responses because of a mistake in a previous migration, so there is a migration to remove this too.

### DIFF
--- a/app/controllers/admin/transport_types_controller.rb
+++ b/app/controllers/admin/transport_types_controller.rb
@@ -28,7 +28,7 @@ module Admin
     private
 
     def transport_type_params
-      params.require(:transport_type).permit(:name, :image, :kg_co2e_per_km, :speed_km_per_hour, :can_share, :park_and_stride, :note)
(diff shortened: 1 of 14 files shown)


urbanwide:
Thanks Leigh, that makes sense - I'll move park and stride to active travel.

For motorbike, if it isn't going to have a category, I'll need to change the db schema & code to allow null for category but this shouldn't be an issue.

urbanwide:
Thanks, that makes sense! I'll make the changes

ldodds (approved):

urbanwide:
Thanks Leigh, I'll merge this now then.

PR #1839, opened 2022-05-30
Added i18n and i18n-js with config files and test js

jhigman:


jhigman:
The latest version of the i18n-js gem and i18n NPM package work together - the gem exports language files, the NPM package uses them to support the JS translations.

The setup is more complex than with previous versions of the gem, and there's a possible concern around the process for generating the json languages files when content is added or changed in the yaml files.

The NPM package requires Webpacker to compile which may add complexity in the tool chain (though we're already using it for transport stuff).

Future versions of the gem and package may change again - this is an alpha version of the gem.

jhigman:
Closing until required

PR #1841, opened 2022-05-31
Write translated text into template and interpolate in js

jhigman:


jhigman:
Alternative approach, writing out the translated strings that will be used in the template.

Possible issues around interpolation - in the example, the 'power' value needs to be written out dynamically from the javascript, but that requires interpolation into the translated string.

The interpolation in JS uses '${}' fields, where the Ruby interpolation uses '%{}', so the delimiters have to be substituted.

The simplistic Javascript interpolation works for basic fields, but wouldn't handle the full translation feature set (e.g. zero/one/many counts etc). That would require a callback to the server.

This example is a bit artificial - probably a better way to achieve translation here would be to simply format the text string on the server and send it with the rest of the results of the Ajax call..

PR #1872, opened 2022-06-22
Update kg_co2e_per_km for transport types from the latest gov.uk figures

urbanwide:
Update the transport type kg_co2e_per_km values based on the latest data from gov.uk

I have a query on this: For Taxi, in the file downloaded from gov.uk, a "Regular taxi" has a value of 0.20826, however, it looks like that previously, the same value as Car (new value 0.17148) has been used instead. Should we use the value for "Car" or "Regular Taxi" for Taxi?

Thank you!


ldodds:
Looking at the methodology there's some differences between Cars and Regular Taxis due to different assumptions around how they are used, average occupancy, etc. So I'd use Regular Taxi rather than Car for our Taxi.

I think its also clearer then about where we source the figures.

ldodds (approved):

PR #1911, opened 2022-07-19
Add locale to preview urls

jhigman:


app/helpers/application_helper.rb
```diff
@@ -360,4 +360,17 @@ def file_type_icon(type)
   def calendar_event_status(calendar_event)
     calendar_event.based_on ? 'inherited' : ''
   end
+
+  def current_locale(locale)
```
ldodds, line 364: minor comment: should this be `current_locale?` as its returning a boolean?

ldodds (approved):

app/helpers/application_helper.rb
jhigman: yes good idea

PR #1922, opened 2022-07-25
Admin activity type reports

ldodds:
This PR adds a couple of new admin reports to allow us to better review and manage the activity types.

The main report is a summary of all activity types, along with some stats on how often they've been recorded and referenced from other parts of the system.

Then for each activity type there is a report that summarises which schools have recorded that activity, along with dates, etc.

I've also tidied up a couple of related pages to add some consistent navigation, table formatting, etc.

The tests are basic: just exercising that the reports display with the expected content and some navigation checks.

* [x] activity types report
* [x] individual activity types report
* [x] add some basic tests

### DIFF
--- a/app/controllers/admin/activity_types_controller.rb
+++ b/app/controllers/admin/activity_types_controller.rb
@@ -7,6 +7,7 @@ module Admin
     before_action :load_chart_list, except: :index
 
     def index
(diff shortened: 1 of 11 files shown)


iant (approved):
couple of comments but looks good to me @ldodds 👍

PR #1923, opened 2022-07-26
Synchronisation of equivalence templates

jhigman:


ldodds (changes requested):
I think we need to override the `tx_resources` method to make it something like:

```
  def self.tx_resources
    EquivalenceType.all.map(&:current_content)
  end
```

So that we're only synchronising the current `EquivalenceTypeContentVersion` for each type. There's around 102 content versions, but only 29 current ones.

ldodds (approved):

PR #1929, opened 2022-07-29
Github actions spike

iant:
This is a spike to see if we can use Github actions to run tests.

### DIFF
--- a/.github/workflows/rubyonrails.yml
+++ b/.github/workflows/rubyonrails.yml
@@ -1,0 +1,38 @@
+name: "Ruby on Rails CI"
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:11-alpine
+        ports:
+          - "5432:5432"
+        env:
+          POSTGRES_DB: rails_test
+          POSTGRES_USER: rails
+          POSTGRES_PASSWORD: password
+    env:
+      RAILS_ENV: test
+      DATABASE_URL: "postgres://rails:password@localhost:5432/rails_test"
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


iant:
@ldodds this is working and comparable in speed to travis (as far as i can tell).

.github/workflows/rubyonrails.yml
```diff
@@ -0,0 +1,42 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:11-alpine
```
iant, line 12: postgres 10

.github/workflows/rubyonrails.yml
```diff
@@ -0,0 +1,42 @@
+name: "Ruby on Rails CI"
+on:
+  push:
+    branches: [ master ]
```
ldodds, line 4: Add in `aws-eb-test` and `aws-eb-master` so that we trigger a build when we push to these branches ahead of a deploy.

.github/workflows/rubyonrails.yml
```diff
@@ -0,0 +1,42 @@
...
+  pull_request:
+    branches: [ master ]
```
ldodds, line 6: This means only PRs that will be merged into `master` will be built. I've occasionally had to do a couple of PRs that patch the live server (we weren't ready to release `master` at that time). So as an extra safety net, I'd add `aws-eb-master` alongside `master` here.

ldodds (changes requested):

.github/workflows/rubyonrails.yml
iant: :+1:

.github/workflows/rubyonrails.yml
iant: :+1:

ldodds (approved):

PR #1933, opened 2022-08-01
2330 localise pupil dashboard templates to use i18n

jhigman:
Extracting text strings from the templates for pupil dashboard and analysis pages.

A few points:

- some strings have embedded html tags (mostly "strong") and so have "_html" suffixes on keys
- those strings may be used in title tags, in which case they need to have "strip_tags" applies
- some strings have ellipses, which are encoded as &hellip; , in which case they need "sanitize" to be applied
- some strings have interpolated values which are themselves translated
- the values for interpolation are converted to i18n keys (these are basically energy types)
- "Storage Heaters" gets converted to 'storage_heaters'
- "Electricity+Solar PV" gets converted to 'electricity_and_solar_pv' (original string is used as part of url path, hence "+" character)
- some common strings for links (e.g. "Back") are in common.labels
- ordinals (e.g. on podium) will be handled by custom ordinal i18n code elsewhere
- some of the strings in the analysis pages could almost certainly be parameterised to reduce number

### DIFF
--- a/config/locales/views/pupils/pupils.yml
+++ b/config/locales/views/pupils/pupils.yml
@@ -1,36 +1,36 @@
 ---
 en:
   pupils:
-    schools:
-      show:
-        title: Pupil dashboard for %{school_name}
-        look_at_the_energy_use: Let's look at the energy use data for your school
-        activities_completed: You've completed %{count} activities
-        no_activities_completed: No activities completed, make a start!
-        complete_an_activity: Complete an activity to score points on Energy Sparks
-        try_this_first: Try this one first
-        try_this_next: Try this one next
-        choose_another_activity: Choose another activity
-        see_scoreboard: See the scoreboard
-        points: points
-        admin_view: Admin view
-        user_view: User view
-        setting_up: We're setting up this school's energy data and will update this page when it is ready to explore
-        find_out_more: Find out more&hellip;
-        things_to_do: Things to do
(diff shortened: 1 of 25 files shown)


config/locales/views/pupils/pupils.yml
```diff
@@ -0,0 +1,67 @@
...
+      find_how_much_storage_heaters_cost_html: find out <strong>how much</strong> the storage heaters cost
+      find_out_about_solar_html: find out about <strong>solar</strong>
+      find_storage_heaters_comparison: find out how my school's storage heater use compares with other schools
+      find_storage_heaters_use_last_year: find out how much electricity was used by storage heaters last year
+      find_storage_heaters_use_long_term: find out how storage heater electricity use has changed long term
+      find_when_electricity_used_html: find out <strong>when</strong> the electricity was used
+      find_when_gas_used_html: find out <strong>when</strong> the gas was used
+      find_when_storage_heaters_used_html: find out <strong>when</strong> the storage heaters used electricity
+      gas_data_html: Here's the <strong>gas</strong> data for your school
+      i_want_to: I want to&hellip;
```
ldodds, line 36: I'm wondering whether we need to use the HTML entity here rather than just "…".

If we switched to using the UTF-8 symbol then we can remove calling sanitize on this when included. https://techoverflow.net/2019/08/05/what-is-hellip-in-html/

ldodds (changes requested):

ldodds (approved):

PR #1936, opened 2022-08-02
Added data pipeline build command

jhigman:


iant:
hi @jhigman @ldodds looks like this is failing because of the different ruby versions.  I'll pick this up as part of https://trello.com/c/QIv3UDX5/2343-extract-data-pipeline-into-separate-github-project (with actions/tests running on a separate repo) so will close this for now...

PR #1937, opened 2022-08-02
Change how we load environment variables

ldodds:
In the upgrade from Amazon Linux 1 to Amazon Linux 2 we had to change how we load environment variables. There were some helper scripts and files created in AL1 which were no longer available.

In AL2 I started using this to load the variables:

```
export $(cat /opt/elasticbeanstalk/deployment/env | xargs)
```

This works fine. However, when attempting to clone an existing environment I found that the file `/opt/elasticbeanstalk/deployment/env` does not seem to be created until right at the end of a successful deployment **after** the after party tasks are run.

This means I was getting an error in the log from the script as it could not load the environment variables. 

I think this also means that the first deployment of code to a new platform version has previously been failing to run those scripts.

There are various solutions but the simplest option was to use:

```
export $(/opt/elasticbeanstalk/bin/get-config --output YAML environment | sed -r 's/: /=/' | xargs)
```

This uses an AWS utility to show the environment variables as YAML. Sed is used to turn them into the expected format which is `NAME=VALUE` rather than `NAME: VALUE`.

I've run this command on the test server to confirm it works OK.

### DIFF
--- a/.platform/hooks/predeploy/99_after_party.sh
+++ b/.platform/hooks/predeploy/99_after_party.sh
@@ -5,7 +5,7 @@ EB_APP_STAGING_DIR=$(/opt/elasticbeanstalk/bin/get-config platformconfig -k AppS
 EB_APP_USER=$(/opt/elasticbeanstalk/bin/get-config platformconfig -k AppUser)
 
 set +x
-export $(cat /opt/elasticbeanstalk/deployment/env | xargs)
+export $(/opt/elasticbeanstalk/bin/get-config --output YAML environment | sed -r 's/: /=/' | xargs)
 set -x
 
 cd $EB_APP_STAGING_DIR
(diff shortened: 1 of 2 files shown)


jhigman (approved):

ldodds:
I had to make some changes to this. It turned out that `/opt/elasticbeanstalk/deployment/env` includes:

* a different `PATH`
* `RBENV_VERSION`
* `RBENV_ROOT`

The different `PATH` was causing `bundler` to not be found. 

I've updated both scripts to set these based on what ElasticBeanstalk is doing. I've manually tested these on the server and have done a release of this branch to test and confirmed that I can run a console.

Looking at the EBS documentation there's nothing specifically mentioned about the values of these environment variables. And other approaches to surfacing the environment variables in the shell (e.g. when accessing over `ssh` or for scripts) assume everything is running as the EC2 user.

iant (approved):

PR #1940, opened 2022-08-03
Add simplified search for translated activity and intervention types

iant:
This PR is to add search for translated activity types

- [x] switch to custom Postgres queries rather than pg_search gem
- [x] ensure that queries are searching both the name and the description for activity type (name involves joining to mobility table with right field and locale; description involves joining to action text with right field and locale). For welsh, use the postgres simple configuration / dictionary rather than English - switch the dictionary in the query based on current locale.
- [x] ensure that queries are searching both the name and the description for intervention type (name involves joining to mobility table with right field and locale; description involves joining to action text with right field and locale. For welsh, use the postgres simple configuration / dictionary rather than English - switch the dictionary in the query based on current locale.


ldodds:
Is this ready for review?

app/models/intervention_type.rb
```diff
@@ -24,22 +24,14 @@
...
   extend Mobility
   include TransifexSerialisable
+  include Searchable
+
   translates :name, type: :string, fallbacks: { cy: :en }
   translates :summary, type: :string, fallbacks: { cy: :en }
   translates :description, backend: :action_text
   translates :download_links, backend: :action_text
 
   include PgSearch::Model
```
ldodds, line 34: Do we still need to include this if we're not using PgSearch on this model? Same on the ActivityType

app/models/concerns/searchable.rb
```diff
@@ -0,0 +1,79 @@
...
+              (
+                to_tsquery(:dictionary, ''' ' || :query || ' ''')
+              ), 0
+            )
+          ) AS rank FROM "#{table}"
+
+          LEFT OUTER JOIN (
+            SELECT "#{table}"."id" AS id, "action_text_rich_texts"."body"::text AS action_text_rich_texts_body
+            FROM "#{table}"
+            INNER JOIN "action_text_rich_texts" ON "action_text_rich_texts"."record_type" = '#{name}'
```
ldodds, line 42: In some cases we're passing in named parameter, e.g. `:locale` and `:dictionary` and in others using interpolated strings e.g. `name` and `table`. Is there a reason for that?

spec/models/activity_type_spec.rb
```diff
@@ -26,12 +26,12 @@
 
     it 'updates original name so search still works' do
```
ldodds, line 27: This is testing that we were still backfilling the original `name` attribute via `copy_searchable_attributes` because the previous version used pg_search. 

We don't really need this behaviour any more. So I think we should either remove that behaviour in this PR or do a follow-on PR to remove it.

spec/models/intervention_type_spec.rb
```diff
@@ -34,15 +34,15 @@
       intervention_type_1 = create(:intervention_type, name: 'foo')
       intervention_type_2 = create(:intervention_type, name: 'bar')
 
-      expect(InterventionType.search('foo')).to eq([intervention_type_1])
-      expect(InterventionType.search('bar')).to eq([intervention_type_2])
+      expect(InterventionType.search(query: 'foo', locale: 'en')).to eq([intervention_type_1])
```
ldodds, line 37: I think we should improve the tests to add both english and welsh attribute names and confirm that a search in English does not find Welsh results and vice versa. This should be fairly simple to add by setting the appropriate attributes and adding some variants of the searches.

The tests currently only assets that searching by name works, but can you also add tests that show that matching is being done on all of the indexed fields, e.g. name, summary and description. That could be done for a single locale.

ldodds (changes requested):

app/models/intervention_type.rb
iant: :+1:  I've removed now

spec/models/intervention_type_spec.rb
iant: :+1: done

spec/models/activity_type_spec.rb
iant: @ldodds fyi i've removed the non null constraint from intervention types in this pr (see above) so we can write the search content test.

ldodds (approved):

PR #1942, opened 2022-08-04
Remove default dates on activity and action forms

ldodds:
We encourage schools to record activities and actions, they can record what they do each day but more frequently they log activities in a series of updates, specifying when they did them.

Currently we default the dates on the forms to the current date. This isn't encouraging users to pick the actual date they record an activity.

We've decided to remove the defaults for now

### DIFF
--- a/app/controllers/interventions_controller.rb
+++ b/app/controllers/interventions_controller.rb
@@ -2,7 +2,7 @@ class InterventionsController < ApplicationController
   def new
     @intervention_type = InterventionType.find(params[:intervention_type_id])
     @intervention_type_group = @intervention_type.intervention_type_group
-    @observation = current_user.school.observations.new(intervention_type_id: @intervention_type.id, at: Time.zone.now)
+    @observation = current_user.school.observations.new(intervention_type_id: @intervention_type.id)
     authorize! :create, @observation
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/interventions_controller.rb
```diff
@@ -2,7 +2,7 @@ class InterventionsController < ApplicationController
   def new
     @intervention_type = InterventionType.find(params[:intervention_type_id])
     @intervention_type_group = @intervention_type.intervention_type_group
-    @observation = current_user.school.observations.new(intervention_type_id: @intervention_type.id, at: Time.zone.now)
+    @observation = current_user.school.observations.new(intervention_type_id: @intervention_type.id)
```
iant, line 5: hi @ldodds  just to double check (as i may well be looking in the wrong place) but looks like there's a couple of validations for `:at` on `observation.rb`
```
  ...
  validates_presence_of :at, :school
  validate :at_date_cannot_be_in_the_future
  ...
```
will these need removing?

iant (approved):
looks good to me.  one comment on a possible validation that may need removing.

app/controllers/interventions_controller.rb
ldodds: No, they need to provide one. This just ensures they pick one rather than submitting the form with the default.

iant: 👍

PR #1954, opened 2022-08-11
Update TX config

ldodds:
Ran:

```
bundle exec rake i18n:copy_analytics_yaml
bundle exec rake i18n:generate_tx_config
```

To generate an updated config for the transifex command-line tool. The tool is used to push YAML files to transifex. 

Contents of this config has been pushed. Checking in update so we can pull down YAML when translated.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -36,9 +36,58 @@ source_lang  = en
 type         = YML
 minimum_perc = 0
 
+[o:energy-sparks:p:energy-sparks:r:en-yml]
+file_filter  = config/locales/<lang>/en.yml
+source_file  = config/locales/en.yml
+source_lang  = en
+type         = YML
+minimum_perc = 0
+
+[o:energy-sparks:p:energy-sparks:r:helpers-application_helper-yml]
+file_filter  = config/locales/<lang>/helpers/application_helper.yml
+source_file  = config/locales/helpers/application_helper.yml
+source_lang  = en
+type         = YML
+minimum_perc = 0
+
+[o:energy-sparks:p:energy-sparks:r:helpers-timeline_helper-yml]
+file_filter  = config/locales/<lang>/helpers/timeline_helper.yml
+source_file  = config/locales/helpers/timeline_helper.yml
+source_lang  = en
+type         = YML
+minimum_perc = 0
+
 [o:energy-sparks:p:energy-sparks:r:mailers-mailer-yml]
 file_filter  = config/locales/<lang>/mailers/mailer.yml
 source_file  = config/locales/mailers/mailer.yml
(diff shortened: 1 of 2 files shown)


.tx/config
```diff
@@ -36,38 +36,255 @@ source_lang  = en
...
+source_lang  = en
+type         = YML
+minimum_perc = 0
+
 [o:energy-sparks:p:energy-sparks:r:mailers-mailer-yml]
 file_filter  = config/locales/<lang>/mailers/mailer.yml
 source_file  = config/locales/mailers/mailer.yml
 source_lang  = en
 type         = YML
 minimum_perc = 0
```
iant, line 66: does there need to be a config for the `config/locales/analytics` dir?  only one file in there called `common.yml`

.tx/config
```diff
@@ -36,38 +36,255 @@ source_lang  = en
...
+source_lang  = en
+type         = YML
+minimum_perc = 0
+
 [o:energy-sparks:p:energy-sparks:r:views-home-home-yml]
 file_filter  = config/locales/<lang>/views/home/home.yml
 source_file  = config/locales/views/home/home.yml
 source_lang  = en
 type         = YML
 minimum_perc = 0
```
iant, line 192: do we also need a config for `config/locales/views/home/privacy_and_cookie_policy.yml` here?

.tx/config
```diff
@@ -36,38 +36,255 @@ source_lang  = en
...
 
+[o:energy-sparks:p:energy-sparks:r:views-terms_and_conditions-terms_and_conditions-yml]
+file_filter  = config/locales/<lang>/views/terms_and_conditions/terms_and_conditions.yml
+source_file  = config/locales/views/terms_and_conditions/terms_and_conditions.yml
+source_lang  = en
+type         = YML
+minimum_perc = 0
+
```
iant, line 290: are configs needed for `activerecord.yml`, `common.yml`, `default.yml` and `devise.yml`? (maybe not as in the locale dir - but figured best to double check)

iant (approved):
@ldodds looks good. a few comments on possible files missing?

.tx/config
ldodds: The config should have an entry for every YAML file, except anything in the `cy` directory. I've checked and these are all present, as expected.

.tx/config
ldodds: You can't configure a directory, only specific files. So for now there's just the common.yml. There will be others as we extract text from the analytics.

.tx/config
ldodds: Yes, this was missing. Dodgy regex in the rake task.

PR #1958, opened 2022-08-15
Fix finder to not use meter name

ldodds:
The Solar Edge and Low Carbon Hub loaders were using `first_or_create!` with a set of attributes that included the meter `name`.

But meters can be renamed by admins. This caused a bug where a meter wasn't being properly found after being renamed, but then data failed to insert as the meter did exist.

Switch to use `find_or_create_by!` and passed in a block to set attributes on create.


ldodds:
Merging this without internal review so that I can release changes to fix problem reported by user. Existing tests covered all the basic behaviour, have expanded to check that import works correctly when meters already exist.

PR #1971, opened 2022-08-24
Localise chart series data

iant:


iant:
hi @ldodds would welcome an early review on this pr if you get a mo please 🙏  
particularly on the `label_point_and_max_for` method (there are some expectation tests defined to give you an idea of what's being returned for given string values)

ldodds (approved):
Looks good to me. The important thing is that the series labels, etc fall back to defaults. This should limit any issues missing translations, rather than errors.

There's tests for the other refactorings which is :+1: 

Suggest we merge and then mop up any issues that arise.

PR #1979, opened 2022-08-26
Infinite increases on school comparison charts

iant:
Some school comparison charts return "infinite" values, e.g. percent change in gas usage between last week and this week, where last week there was no gas consumption i.e. this week divided by last week = Infinity. 

The analytics has been updated to add configuration specifying min/max values for charts. See code changed in: https://github.com/Energy-Sparks/energy-sparks_analytics/pull/292

This pr fixes the front-end application needs by applying the max x axis value to highcharts

from
Screenshot 2022-08-26 at 14 35 15

to
Screenshot 2022-08-26 at 15 16 11

### DIFF
--- a/app/models/chart_data_values.rb
+++ b/app/models/chart_data_values.rb
@@ -1,6 +1,6 @@
 class ChartDataValues
   attr_reader :anaylsis_type, :title, :subtitle, :chart1_type, :chart1_subtype,
-              :y_axis_label, :x_axis_label, :x_axis_categories, :x_max_value,
+              :y_axis_label, :x_axis_label, :x_axis_categories, :x_max_value, :x_min_value,
               :advice_header, :advice_footer, :y2_axis_label, :y2_point_format, :y2_max, :x_axis_ranges, :annotations,
               :transformations, :allowed_operations, :drilldown_available, :parent_timescale_description,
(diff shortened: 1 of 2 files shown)


app/models/chart_data_values.rb
```diff
@@ -31,6 +31,7 @@ def initialize(chart, chart_type, transformations: [], allowed_operations: {}, d
       @subtitle           = chart[:subtitle]
       @x_axis_categories  = translate_categories_for(chart[:x_axis])
       @x_axis_ranges      = chart[:x_axis_ranges] # Not actually used but range of actual dates
+      @x_max_value        = chart[:x_max_value]
```
iant, line 34: @ldodds do we need an `x_min_value` too (e.g. `@x_min_value  = -chart[:x_max_value]`) ? it's not defined in Philips code changes.

iant: @ldodds update to the above. as per Philips message on https://trello.com/c/EivNH176/326-update-scatter-charts-to-plot-regression-as-a-line I've set in the .js the min to equal the max if max is present (see above)

iant:
@ldodds as discussed i've added the `x_min_value`.  this is ready for review now.

ldodds (approved):

PR #1983, opened 2022-08-31
Remove null constraint on user of consent grant and test that user de…

jhigman:
…lete works

### DIFF
--- a/app/controllers/schools/consents_controller.rb
+++ b/app/controllers/schools/consents_controller.rb
@@ -8,7 +8,7 @@ module Schools
     end
 
     def create
-      @consent_grant = ConsentGrant.new(school_consent_params.merge(user: current_user, school: @school, ip_address: request.remote_ip))
+      @consent_grant = ConsentGrant.new(school_consent_params.merge(user: current_user, school: @school, ip_address: current_ip_address))
       if @consent_grant.save
         ConsentGrantMailer.with(consent_grant: @consent_grant).email_consent.deliver_now
         redirect_to root_path(@school)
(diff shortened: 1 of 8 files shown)


app/controllers/schools/consents_controller.rb
```diff
@@ -8,7 +8,7 @@ def show
     end
 
     def create
-      @consent_grant = ConsentGrant.new(school_consent_params.merge(user: current_user, school: @school))
+      @consent_grant = ConsentGrant.new(school_consent_params.merge(user: current_user, school: @school, ip_address: request.remote_ip))
```
ldodds, line 11: I think this will be incorrect on the test/live servers because we're behind an Elastic Load Balancer, needs to use X-Forwarded-For I think to be the actual user IP.

ldodds (changes requested):
Think this needs a change to the ip address capture to avoid logging the ELB ip address.

ldodds:
This is the default nginx log config on the servers:

```

    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                      '$status $body_bytes_sent "$http_referer" '
                      '"$http_user_agent" "$http_x_forwarded_for"';
```

And example log entries:

```
172.31.24.130 - - [31/Aug/2022:11:35:38 +0000] "GET / HTTP/1.1" 200 25718 "-" "Mozilla/5.0 (Windows NT 6.2; WOW64) AppleWebKit/537.4 (KHTML, like Gecko) Chrome/98 Safari/537.4 (StatusCake)" "139.59.190.241"
172.31.32.168 - - [31/Aug/2022:11:37:58 +0000] "GET / HTTP/1.1" 301 215 "-" "-" "171.22.30.130"
```
The first entries are the load balancer (we see the same small set of IPs repeatedly), the last entries appear to be the users IP address.

One option would be to log X-Forwarded-For if available, otherwise default to remote address.

ldodds (approved):

PR #1995, opened 2022-09-02
Fix trendline display on charts to use only 2 points

iant:


app/assets/javascripts/common_chart_options.js
```diff
@@ -254,13 +254,29 @@ function scatter(chartData, highchartsChart, seriesData) {
...
+
       highchartsChart.addSeries(
         {
           type: 'line',
           name: seriesData[key].name,
-          data: seriesData[key].data.filter(function( obj ) {
-            return obj[1] !== null; // Remove nulls so there's no gap in the line
-          })
+          data: trendlineData,
+          marker: {
```
urbanwide, line 272: @iant code looks good - I'm not sure of the wider impact of the changes though. I assume the marker stuff is no longer required as part of this too?

iant: hi @urbanwide I think we're ok to remove but I'll make a note to discuss with Leigh when he's back.

urbanwide (approved):
Looks good 👍

PR #2001, opened 2022-09-07
Add I18N to the Transport Survey ERB templates

urbanwide:
This covers i18n for the non-javascript parts of the transport survey app.


iant (approved):
:+1: all looks good to me @urbanwide

PR #2002, opened 2022-09-08
Ensure correct month name and abbr month name array formatting

iant:
`month_names` and `abbr_month_names` are still being pulled from TX with a missing empty 0th element (needed as there's no zero month see https://github.com/rails/rails/blob/main/activesupport/lib/active_support/locale/en.yml).  

This pr adds a a test to ensure `month_names` and `abbr_month_names` for both `en` and `cy` locales are in the correct format in case the missing empty 0th element is accidentally checked in.

### DIFF
--- a/config/webpacker.yml
+++ b/config/webpacker.yml
@@ -17,7 +17,7 @@ default: &default
   cache_manifest: false
 
   # Extract and emit a css file
-  extract_css: false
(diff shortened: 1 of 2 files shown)


urbanwide (approved):
@iant 👍

PR #2006, opened 2022-09-08
Reviewed translations from Transifex (2)

iant:
This PR brings in more newly (100%) reviewed Welsh translations from Transifex

via ../tx pull --force --mode 'onlyreviewed' -l cy


urbanwide (approved):
Looks good @iant

PR #2008, opened 2022-09-09
Make date time picker under usage charts a bit wider

iant:
This pr makes the date time picker slightly wider so year isn't hidden when the selected locale is cy

from
Screenshot 2022-09-09 at 16 03 25

to
Screenshot 2022-09-09 at 16 04 14


urbanwide (approved):
Looks good @iant!

PR #2009, opened 2022-09-09
Add missing translation for activity

iant:


config/locales/views/schools/schools.yml
```diff
@@ -73,6 +73,11 @@ en:
         action:
           one: "%{count} action"
           other: "%{count} actions"
+        activity:
+          are_you_sure: Are you sure?
```
urbanwide, line 77: In common.yml, there is "confirm" which has the "Are you sure?" text in, along with some other common labels such as delete (but not edit as yet, but this could be added). Wondered if these should be used instead? Just a thought.

iant: :+1: good point @urbanwide i've pushed an update to use the common labels.

urbanwide: Nice one 👍 Looks good!

urbanwide (approved):

PR #2011, opened 2022-09-12
i18n for pagy

urbanwide:
This PR covers i18n for pagy, which is used by transport surveys but also site-wide.

When doing this, I found that the pagy.rb, the pagy config file, was not being picked up by the app and so moved it to config/initializers where pagy expects it to be.

Also, I have added config/locales/pagy.yml, and config/locales/cy/pagy.yml. The Welsh version is actually English at the moment, but since I had to configure the app to use a locale file and did not have the welsh version yet, I thought this would be the best thing to do for now. Also technically, we do not need an English yml file for pagy as the translations are inbuilt but since we need to put one there for transifex, I thought it worth linking to from pagy.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,7 @@ class ApplicationController < ActionController::Base
   around_action :switch_locale
   before_action :authenticate_user!
   before_action :analytics_code
+  before_action :pagy_locale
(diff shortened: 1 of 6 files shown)


iant (approved):
looks good to me @urbanwide

urbanwide:
Thank you @iant!

PR #2012, opened 2022-09-13
Fix trendlines for series name starting with Trendline

iant:
This pr fixes the trend line plotting to work for series names starting 'Trendline ...' as well as 'trendline_' it also replaces the first and last point code with `connectNulls: true` to conserve data points (the line was previously breaking due to missing (null) values).


urbanwide (approved):
@iant Looks good and a more simple solution too

PR #2013, opened 2022-09-13
Add transport survey link to pupil dashboard

urbanwide:
Also fixes a button that was using the incorrect translation

### DIFF
--- a/app/views/pupils/schools/show.html.erb
+++ b/app/views/pupils/schools/show.html.erb
@@ -44,16 +44,16 @@
 <% if @show_temperature_observations %>
   <% if @temperature_observations.empty? %>
     <%= render 'schools/dashboard/info_bar',
-          colour: 'bg-positive',
+          colour: 'bg-neutral',
           icon: 'temperature-high fa-3x',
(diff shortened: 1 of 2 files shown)


app/views/management/schools/show.html.erb
```diff
@@ -49,6 +49,8 @@
 
 <%= render 'management/schools/record_intervention', school: @school %>
 
+<%= render 'schools/dashboard/transport_surveys', school: @school %>
```
ldodds, line 52: I think for the moment we were only proposing to add this to the pupil dashboard, and not the management dashboard. So can you remove this?

ldodds (changes requested):
Can you remove from the management dashboard, otherwise rest of changes looks good.

I think you can also remove the temporary link that we added to the "Manage school" menu as well. This means the feature is discoverable via the pupil dashboard and later via the activity that Paula will create.

urbanwide:
Hi Leigh - Claudia asked me to also add the link to the adult dashboard (see my comment in the PR description above and the trello card). Shall I still remove it?

This PR does remove the manage school link too.

ldodds (approved):

PR #2014, opened 2022-09-13
Improve display of switch school menu

ldodds:
When a user is associated with a large number of schools, the list of schools in the "Switch schools" menu gets very long.

This PR:

* Ensures the list is sorted by name (previously it was in an arbitrary order)
* Sets a max height (arbitrary value) for the menu to make it scrollable. ([Example](https://www.codeply.com/p/pHbaRmjRgn)

### DIFF
--- a/app/assets/stylesheets/nav.scss
+++ b/app/assets/stylesheets/nav.scss
@@ -48,6 +48,11 @@
     left: auto;
   }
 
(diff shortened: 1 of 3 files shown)


iant (approved):
:+1: looks good to me @ldodds

PR #2015, opened 2022-09-13
Clean up DashboardAlert records to remove all unused public & teacher alerts

iant:
We no longer generate DashboardAlert records for teacher or public dashboard types.

This pr covers writing an AfterParty tasks to delete the older records for these dashboards.


urbanwide (approved):
Code looks good, assuming we no longer need these type of alerts anymore anyway 👍

ldodds (approved):

PR #2016, opened 2022-09-13
Add missing translations to timeline

urbanwide:
Just a couple of translations missing. Looks like other untranslated items have not yet come down from Transifex.


iant (approved):
:+1:

PR #2025, opened 2022-09-14
Make dashboard info button full width and height

iant:
"Record a pupil activity" text, as shown in dashboard prompt, is too long when shown in Welsh. This pr makes the button size full width and height so no text breaks out of the pill button. 

from
Screenshot 2022-09-14 at 14 18 59

to

Screenshot 2022-09-14 at 14 15 02

Screenshot 2022-09-14 at 14 15 19

Screenshot 2022-09-14 at 14 15 28

### DIFF
--- a/app/views/schools/dashboard/_info_bar.html.erb
+++ b/app/views/schools/dashboard/_info_bar.html.erb
@@ -9,7 +9,7 @@
     </div>
     <% buttons.each do |title, path| %>
       <div class="col-md-3">
-        <%= link_to (sanitize title), path, class: "btn btn-light btn rounded-pill font-weight-bold" %>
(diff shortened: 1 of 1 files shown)


iant:
@urbanwide will the change in this pr break updates you're doing in https://github.com/Energy-Sparks/energy-sparks/pull/2013/files ?

urbanwide:
@iant I don't think your code will have any merge issues with my PR, however, have you considered what might happen if there are two buttons present? The temperature observations (and also now transport surveys from this PR) can both have two buttons...

ldodds:
@iant I'd prefer to keep the button on the right, rather than full width/height. Are there other options?

ldodds:
Could try `height: auto` or `height: fit-content` for these buttons within the infobars.

ldodds (approved):

PR #2026, opened 2022-09-14
Allow welsh case study files to be downloaded

urbanwide:
This is how multi-locale upload looks in the admin/edit form:
Screenshot 2022-09-19 at 19 37 29
This is how the multi-locale downloads can be accessed from the admin/index page (it just had a single "download" link previously). If there is no link, the file has not yet been uploaded.
Screenshot 2022-09-19 at 19 38 57
This is how the case study files can be accessed in the front-end. Not sure about using the Welsh word for download for the Welsh download link - what do you think?
Screenshot 2022-09-19 at 19 39 27

### DIFF
--- a/app/controllers/admin/case_studies_controller.rb
+++ b/app/controllers/admin/case_studies_controller.rb
@@ -40,8 +40,8 @@ module Admin
     private
 
     def case_study_params
-      translated_params = t_params(CaseStudy.mobility_attributes)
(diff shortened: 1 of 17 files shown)


ldodds (approved):
Looks good. Although one question do we need the "mirror" locale name in the `default.yml` file?

urbanwide:
Thanks @ldodds. Mirror was added after there was an issue with rendering in development because the mirror locale did not have a label. Hope this is ok.

PR #2031, opened 2022-09-15
Simplify adult dashboards

ldodds:
We currently have:

- pupil dashboard
- school dashboard (for non-school users, public views, etc)
- management dashboard (for pupils, school users, etc viewing school)

We've been slowly ensuring that the school and management dashboards are aligned so they have similar content, where appropriate. This PR completes the process of removing the management dashboard, so we just have a single Adult dashboard which selectively displays some content based on the user.

There's quite a few files changed here, so here's a summary of changes:

# Controller and Views

Across the site we generally link to the schools controller (`schools_path`), then filter certain users to specific dashboards. E.g. staff, admins, etc went to the mgt dashboard, pupils to the pupils dashboard, non-logged in users to a default public dashboard. The bulk of the changes are to remove these redirects so that everyone ends up at the `schools_path` and the controller + views just change what is displayed based on user permissions.

The older `management_school_path` route has been retained, with a permanent redirect to deal with any bookmarks or other links that might exist. I've updated the code so we don't link to this any more.

One small complication is that without a separate management dashboard I needed to add a special case for pupil users to allow them to visit the adult dashboard. In this new model, there's a `switch` parameter that is added only for links to the adult dashboard for pupil users. This removes the default redirection behaviour. This is also now used in some of the system specs to allow us to test Adult Dashboard behaviour with `pupil` users.

Some redirection behaviour still exists depending on whether the school is not visible, or doesn't have public data or is not yet data-enabled. I've refactored the controller logic for this, but the behaviour was already covered by existing system tests.

# Specs

I've significantly refactored the specs that test all of the above behaviour. While there were already tests for all of the core logic, we were only testing this for a limited selection of accounts: mostly just guest users with most tests limited to `staff` users.

I've broken down the `system/school_spec.rb` to create separate feature specific tests for each of the dashboard features: charts, management table, management priorities, alerts, timeline and navigational elements. These are now found in `spec/system/schools/dashboard`

These specs all have a similar pattern:

- a set of Rspec `shared_examples` that setup the test data and test the common functionality
- Rspec system specs that visits the dashboard as a guest, `pupil`, `staff`, `school_admin`,  `group_admin` and `admin` users and running the shared examples. This allows us to easily test the common functionality across all user types
- Additional system specs that test extra or different behaviour for specific roles

This greatly improves the test coverage across all of those roles as previously we weren't checking all of this behaviour for every type of user.

Checklist of changes:

* [x] Refactor the SchoolsController to retain necessary redirects, but otherwise allow school users to access this path
* [x] Copy functionality from Management::SchoolsController to SchoolsController plus templates, to align them
* [x] Change how the sub nav selectively links to the dashboard, so its always targeting the school path
* [x] Ensure that Management::SchoolsController functionality is behind test for ability to `show_management_dash` to ensure access control remains unchanged for management features
* [x] Revise management controller tests, to use `school_path` instead to confirm functionality is working as originally designed
* [x] Change Management::SchoolsController so that it just does a permanent redirect
* [x] Remove any direct links to `management_school_path` in code, templates and specs using `school_path` instead
* [x] Tidy up controller
* [x] Tidy up templates and partials
* [x] Tidy up tests
* [x] Add in any missing tests as needed
* [x] Fix printable report


app/controllers/schools_controller.rb
```diff
@@ -101,8 +110,40 @@ def destroy
...
+
   def setup_default_features
     @observations = setup_timeline(@school.observations)
+    #Setup management dashboard features if users has permission
+    #to do that
+    if can?(:show_management_dash, @school)
+      @add_contacts = site_settings.message_for_no_contacts && @school.contacts.empty? && can?(:manage, Contact)
+      @add_pupils = site_settings.message_for_no_pupil_accounts && @school.users.pupil.empty? && can?(:manage_users, @school)
+      @prompt_training = !@show_data_enabled_features || current_user.confirmed_at < 60.days.ago
+      @prompt_for_bill = @school.bill_requested && can?(:index, ConsentDocument)
```
iant, line 145: perhaps "prompt training" and `prompt for bill` could be methods on the user model? something like
```
PROMPT_TRAINING_DAY_THRESHOLD = 60
...
def prompt_training?
  confirmed_at && (confirmed_at < PROMPT_TRAINING_DAY_THRESHOLD.days.ago)
end
```

```
def prompt_for_bill?
  return unless can?(:index, ConsentDocument)
  school.bill_requested
end
```

app/views/schools/show.html.erb
```diff
@@ -19,9 +19,57 @@
...
 
+<% if @suggest_estimates_for_fuel_types.present? && @suggest_estimates_for_fuel_types.any? %>
+  <%= render 'management/schools/prompt_for_estimate', school: @school, fuel_types: @suggest_estimates_for_fuel_types %>
+<% end %>
+
+<% if @prompt_for_bill %>
+  <%= render 'management/schools/prompt_for_bill', school: @school %>
+<% end %>
+
+<% if @prompt_training %>
```
iant, line 30: `if current_user.prompt_for_bill?`  (if moved to the user model)?
`if current_user.prompt_training? || !@show_data_enabled_features`  (if moved to the user model)?

iant (approved):
this looks good @ldodds couple of comments with suggestions on possibly extracting bits out to methods, but totally ok as is.  scanned the tests but they look good too.

PR #2039, opened 2022-09-20
Ensure statistics for amr_data_feed_readings table are always up to date

urbanwide:
Enables vacuum service for the following:
- [x] import_all.rake (updates existing code)
- [x] after doing the N3rgy readings inserts, see `import_n3rgy_readings.rake`
- [x] after importing solar data, see the tasks in `lib/tasks/solar`
- [x] at the end of the `ManualDataLoadRunJob`

See note against import_all.rake about moving the Vacuum Service call out of the loop. I've not particularly been able to manually test the rake tasks, so suggest this needs doing on the test server (as you'd expect really I guess!)?

Worth noting for future implementation that a VACUUM can't be run inside a transaction block - this tripped me up for a while trying to work out why my rspec wasn't working.

### DIFF
--- a/lib/tasks/amr_importer/import_all.rake
+++ b/lib/tasks/amr_importer/import_all.rake
@@ -6,7 +6,7 @@ namespace :amr do
       begin
         FileUtils.mkdir_p config.local_bucket_path
         Amr::Importer.new(config).import_all
-        Database::VacuumService.new([:amr_data_feed_readings])
+        Database::VacuumService.new([:amr_data_feed_readings]).perform
       rescue => e
         puts "Exception: running import_all for #{config.description}: #{e.class} #{e.message}"
         puts e.backtrace.join("\n")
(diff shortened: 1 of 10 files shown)


lib/tasks/amr_importer/import_all.rake
```diff
@@ -6,7 +6,6 @@ namespace :amr do
       begin
         FileUtils.mkdir_p config.local_bucket_path
         Amr::Importer.new(config).import_all
-        Database::VacuumService.new([:amr_data_feed_readings])
```
urbanwide, line 9: The vacuum service would have been called multiple times for this job on the same table if left here (once for each S3_folder), so, I moved it out of here and to the end, so that it is only called once for this job. Hope this is the right thing to do!

urbanwide:
Ok that makes sense what you said, thanks. Hopefully the changes I've made are ok.

ldodds (approved):

PR #2043, opened 2022-09-21
Allow editing of pseudo meters

iant:
The Meter Management page does not provide a link to edit pseudo meters. But there is a link to edit them via the detail page.

The editing was originally disabled due to a limitation in the data loader, which has now been removed.

This pr:
- [x] allows the edit button to appear on the Meter Management (meters/index) page like other meters
- [x] adds a warning, for admins on the form for pseudo meters to warn about changing the type or mpan
- [x] adds validations to stop changing the type or mpan if its a pseudo meter (only)

### DIFF
--- a/app/views/schools/meters/_active_meters.html.erb
+++ b/app/views/schools/meters/_active_meters.html.erb
@@ -31,12 +31,10 @@
       </div>
     </td>
     <td>
-      <% if ! meter.pseudo %>
-        <div class="btn-group btn-group-sm" role="group">
-          <%= link_to 'Edit', edit_school_meter_path(@school, meter), class: 'btn btn-sm' if can?(:edit, meter) %>
-          <%= link_to 'Deactivate', deactivate_school_meter_path(@school, meter), method: :put, class: 'btn btn-sm' if can?(:deactivate, meter) %>
-        </div>
-      <% end %>
+      <div class="btn-group btn-group-sm" role="group">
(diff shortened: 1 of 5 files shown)


app/views/schools/meters/_active_meters.html.erb
```diff
@@ -31,12 +31,10 @@
...
     <td>
-      <% if ! meter.pseudo %>
-        <div class="btn-group btn-group-sm" role="group">
-          <%= link_to 'Edit', edit_school_meter_path(@school, meter), class: 'btn btn-sm' if can?(:edit, meter) %>
-          <%= link_to 'Deactivate', deactivate_school_meter_path(@school, meter), method: :put, class: 'btn btn-sm' if can?(:deactivate, meter) %>
-        </div>
-      <% end %>
+      <div class="btn-group btn-group-sm" role="group">
+        <%= link_to 'Edit', edit_school_meter_path(@school, meter), class: 'btn btn-sm' if can?(:edit, meter) %>
+        <%= link_to 'Deactivate', deactivate_school_meter_path(@school, meter), method: :put, class: 'btn btn-sm' if can?(:deactivate, meter) %>
```
iant, line 36: should the deactivate link remain hidden here?

ldodds: No, I think you should be able to activate/deactivate the meters too.

iant: 👍 ta!

PR #2045, opened 2022-09-21
Do not include school groups with no schools on benchmark page

iant:
The list of school groups on the Benchmark page, e.g. [here](https://energysparks.uk/benchmark?benchmark%5Bschool_types%5D%5B%5D=0&benchmark%5Bschool_types%5D%5B%5D=1&benchmark%5Bschool_types%5D%5B%5D=2&benchmark%5Bschool_types%5D%5B%5D=3&benchmark%5Bschool_types%5D%5B%5D=4&benchmark%5Bschool_types%5D%5B%5D=5&benchmark%5Bschool_types%5D%5B%5D=6&benchmark_type=annual_energy_costs_per_pupil) is produced by the `ComparisonService`

It's possible for a school group that doesn't yet have any visible schools to appear on this list. This is occasionally causing an analytics error.  This PR removes this from happening to ensure that any Rollbar issues are actually with the benchmarking code.

- [x] for an admin user, only include school groups that have schools
- [x] for other users, only include school groups that are public and which have at least one visible school (the users own group should always been in the list).

### DIFF
--- a/app/models/school_group.rb
+++ b/app/models/school_group.rb
@@ -86,4 +86,8 @@ class SchoolGroup < ApplicationRecord
   def page_anchor
     name.parameterize
   end
+
+  def self.with_active_schools
+    joins(:schools).where('schools.active = true').distinct
+  end
 end
--- a/app/services/comparison_service.rb
+++ b/app/services/comparison_service.rb
@@ -16,9 +16,9 @@ class ComparisonService
 
   def list_school_groups
     if @user.present? && @user.admin?
-      return SchoolGroup.order(:name).to_a
+      return SchoolGroup.with_active_schools.order(:name).to_a
     end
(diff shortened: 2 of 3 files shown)


app/models/school_group.rb
```diff
@@ -86,4 +86,8 @@ def update_school_partner_positions!(position_attributes)
   def page_anchor
     name.parameterize
   end
+
+  def self.with_active_schools
+    joins(:schools).where('schools.active = true').distinct
```
iant, line 91: `SELECT DISTINCT "school_groups".* FROM "school_groups" INNER JOIN "schools" ON "schools"."school_group_id" = "school_groups"."id" WHERE (schools.active = true);`

ldodds (approved):

PR #2048, opened 2022-09-22
Missing chart_type param

iant:
We are getting a flurry of Rollbar errors due to a missing chart_type parameter:

https://rollbar.com/energysparks/EnergySparksProduction/items/1238/?item_page=0&#instances

Its not clear why this is happening, but looking at the error, these all(?) seem to have a From header of "googlebot(at)googlebot.com".

So this might be a search engine error, e.g. following a URL in the response but without adding a parameter (which might have been injected via JS?)

Need to:

- [x] check whether we can end up having any direct links to chart.json without a chart_type
- [x] ensure we return a 400 Bad Request error for the missing param, not a 500. This might discourage repeated requests
- [x] perhaps add a search engine exclusion for that url to robots.txt

Returns 
`{"error":"param is missing or the value is empty: chart_type","status":400}`

### DIFF
--- a/app/controllers/schools/charts_controller.rb
+++ b/app/controllers/schools/charts_controller.rb
@@ -10,38 +10,42 @@ class Schools::ChartsController < ApplicationController
   before_action :check_aggregated_school_in_cache
 
   def show
+    respond_to do |format|
+      format.html { render_html }
+      format.json { render_json }
+    end
+  end
+
+private
+
+  def render_html
+    @chart_type = params.require(:chart_type).to_sym
+    set_measurement_options
+    @measurement = measurement_unit(params[:measurement])
+    @title = @chart_type.to_s.humanize
+  end
+
+  def render_json
     @chart_type ||= begin
                       params.require(:chart_type).to_sym
                     rescue => error
                       render json: { error: error, status: 400 }.to_json and return
                     end
 
-    respond_to do |format|
-      format.html do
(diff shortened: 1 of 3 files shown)


app/controllers/schools/charts_controller.rb
```diff
@@ -10,34 +10,43 @@ class Schools::ChartsController < ApplicationController
...
-      end
+      format.html { render_html }
+      format.json { render_json }
     end
   end
 
 private
 
+  def render_html
+    @chart_type = params.require(:chart_type).to_sym
```
iant, line 22: really this should use strong params throughout the controller but i'm a bit reluctant to do a big refactor here without proper test coverage.  e.g.
```
  def chart_params
    params.require(:chart_type).permit(:date_ranges, :transformations, :school_id, :y_axis_units, :chart_y_axis_units, :series_breakdown, :mpan_mprn, :measurement)
  end
```

```
    chart_config = {
      mpan_mprn: chart_params.mpan_mprn,
      series_breakdown: chart_params.series_breakdown,
      date_ranges: get_date_ranges
    }
```

ldodds (approved):
@iant I think this looks OK, the main thing is to return a better error for JSON responses, which this is now doing I think.

I'm still unclear why we're getting those, but can only conclude its search engines just following links. Lets merge this ready for review on test server.

PR #2051, opened 2022-09-26
Repeat manual load

urbanwide:
A few questions for this (otherwise straightforward) task:

- We want to remove the old manual upload button - are we happy to remove the functionality in the controller / routes too?
- Do we want to rename the "Insert this data (NEW)" button to "Insert this data"?
- I've added a button to "Upload another file" to the bottom of the page after completing a run. Hopefully this is the right place and do we only want this to appear if the job is complete? Or are we happy with it there when the job is still pending?


urbanwide:
Hi @ldodds I've popped a few questions in the description of this task if you'd be able to take a look when you're feeling better? Thanks, Deb

ldodds:
@urbanwide yes, we should remove the old upload routes/controller. We'll only be using this new option now. Renaming the button as you suggested, and only showing option to submit another load once the current one has completed sounds fine to me.

urbanwide:
Thanks @ldodds. One last question, there is a link on the complete run page to the configuration (Banes in this example below):
Screenshot 2022-09-27 at 13 30 57

However, when I click on the link on my local machine, it goes to the following page, which doesn't seem right to me:

Screenshot 2022-09-27 at 13 31 27

Shall I change it, and if so to which page? This one?

Screenshot 2022-09-27 at 14 01 00

Thanks!

ldodds (approved):

PR #2052, opened 2022-09-27
Allow old content managed alert ratings to be deleted

iant:


app/controllers/admin/alert_types/ratings_controller.rb
```diff
@@ -46,6 +46,20 @@ def update
         end
       end
 
+      def destroy
```
ldodds, line 49: Check for ContentManaged alert class here too, throw an error if not

ldodds (changes requested):

app/controllers/admin/alert_types/ratings_controller.rb
iant: :+1: done now.  this is good for another review when you get a mo...

ldodds (approved):

PR #2058, opened 2022-09-28
Integrate Delayed Job Web

iant:
To help us view and manage the job queues we will need an interface onto DelayedJob.  This PR adds [Delayed Job Web](https://github.com/ejschmitt/delayed_job_web) to the Admin pages.  This interface needs to be accessible only to EnergySparks admin users, so it also ensures the routes are authenticated.

- [x] add delayed job web
- [x] add authentication
- [x] add a link to the ui from the admin dashboard


config/routes.rb
```diff
@@ -426,6 +426,9 @@
 
     post 'amr_data_feed_readings/:amr_uploaded_reading_id', to: 'amr_data_feed_readings#create', as: :create_amr_data_feed_readings
 
+    authenticated :user, -> user { user.admin? }  do
```
ldodds, line 429: Could you add a simple system spec to confirm that the authentication of the route is working as expected? Maybe one for a guest/school user and one for admin.

ldodds (changes requested):

config/routes.rb
iant: :+1:   have just pushed an update with some tests.

ldodds (approved):

PR #2069, opened 2022-09-29
Transifex pull 2022-09-29

ldodds:


config/locales/cy/default.yml
```diff
@@ -8,8 +8,9 @@ cy:
       update: Diweddaru %{model}
   languages:
     cy: Cymraeg
-    en: English
-  name: Cymraeg
+    en: Saesneg
```
iant, line 11: I think this needs to stay as "English" as it's used by the locale switcher.

iant (approved):

PR #2070, opened 2022-09-30
Configure cronjob for DelayedJob metrics

iant:


.ebextensions/cronjob.config
```diff
@@ -248,6 +248,28 @@ files:
...
+
+    #
+    # Send Delayed Job process metrics to AWS Cloudwatch
+    #
+    "/etc/cron.d/send-delayed-job-process-metrics-to-cloudwatch":
+        mode: "000644"
+        owner: root
+        group: root
+        content: |
+          */5 * * * * root aws cloudwatch put-metric-data --metric-name ProcessCount --namespace DelayedJob --value ./script/delayed_job_number_of_work_process.sh --unit Count --dimensions ec2-metadata --instance-id
```
ldodds, line 270: The parameters are different to the other job. `--value` is assigned using different syntax, there's no region, and there instance id doesn't look like its set correctly? Is this a mistake or is there a reason for the differences?

iant: @ldodds yep, that was a mistake, sorry about that.  fixed now in the pr and tested on the `test-3-4-9`.  this should be ok to go now.

PR #2073, opened 2022-09-30
Replace validation and content batch with delayed jobs

iant:


lib/tasks/amr_importer/validate_amr_readings.rake
```diff
@@ -1,27 +1,8 @@
...
-        Rails.logger.error "Exception: running validation for #{each_school.name}: #{e.class} #{e.message}"
-        Rails.logger.error e.backtrace.join("\n")
-        Rollbar.error(e, job: :validate_amr_readings, school_id: each_school.id, school: each_school.name)
-      end
+    School.process_data.each do |school|
+      ValidateAndAggregateAmrReadingsJob.perform_later(school)
     end
-
-    total_amr_readings_after = AmrValidatedReading.count
-    puts "Total AMR Readings after: #{total_amr_readings_after} - inserted: #{total_amr_readings_after - total_amr_readings_before}"
```
iant, line 24: @ldodds I'm thinking here we publish `total_amr_readings_before` and `total_amr_readings_after` to cloudwatch as hooks on the first and last job, respectively.  what do you think?

iant: (in fact maybe we just need to publish total_amr_readings_after as we'd already have a count from the previous run)

config/initializers/wisper_event_subscriptions.rb
```diff
@@ -1,8 +1,5 @@
 # Initializer to register event listeners via the Wisper (https://github.com/krisleech/wisper) gem
 
-#Monitor when school has been regenerated so we can show school target progress report
-Wisper.subscribe(Targets::ContentGenerationListener.new, :ContentBatch)
-
```
iant, line 5: need to do this as part of content batch job when all school jobs have finished

school_target = school.most_recent_target
school_target.update(report_last_generated: Time.zone.now) if school_target.present?

PR #2078, opened 2022-10-03
Ensure confirmation emails sent for admin created accounts

ldodds:
Send confirmation emails for all accounts created by an admin, to ensure that user sets password and agrees to the terms and conditions.

Tweak formatting of table to use Bootstrap 4 table class, not older Bootstrap 3 class.


urbanwide (approved):
👍

PR #2085, opened 2022-10-04
Replace content batch with good job

iant:


lib/tasks/content_batch.rake
```diff
@@ -1,9 +1,7 @@
 namespace :content do
   desc "Content batch"
   task batch: :environment do
-    puts "#{DateTime.now.utc} Content batch start"
     Rails.cache.clear
```
iant, line 4: can we move this (optionally?) to the job?

PR #2090, opened 2022-10-05
Queue configuration for Good Job

iant:
We will have the following jobs executing via Good Job:

- `DccCheckerJob`, checks meter status after it has been setup. Runs intermittently
- `GrantTrustedConsentsJob`, runs when admin completes review of a meter. So intermittently
- `ManualDataLoadRunJob` runs when admin uploads CSV file. Done regularly during the day
- `SchoolBatchRunJob` runs when admin regenerates a school. Done regularly during the day
- "DailyRegenerationJob" a new job (wip) to run validation and content generation, which will schedule several hundred jobs (one per school) every morning

So we want to ensure that we process the daily regeneration jobs as quickly as possible, so it is faster than the current batch processes. Whilst also ensuring that admin jobs are run as soon as possible, and that the daily regeneration jobs don't swamp the server.

Initially we will configure two queues so we need to update the GoodJob configuration to use:

- [x] `regeneration` queue, configured to use 2 workers. The new "DailyRegenerationJob" will submit jobs to this queue
- [x] `default` queue, configured to use 5 workers. All other jobs submitted to this queue
- [x] When `ManualDataLoadRunJob` and `SchoolBatchRunJob` are submitted give then a priority of `10` and the `DccCheckerJob` and `GrantTrustedConsentsJob` jobs a priority of `5`. This will allow admin jobs to take priority over the other infrequent jobs.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
 web:  bundle exec puma -C config/puma.rb
-worker: bundle exec good_job start
+worker: bundle exec good_job --queues="regeneration:2;*;"
--- a/app/jobs/dcc_checker_job.rb
+++ b/app/jobs/dcc_checker_job.rb
@@ -1,6 +1,9 @@
 class DccCheckerJob < ApplicationJob
-  self.queue_adapter = :good_job
   queue_as :default
+
+  def priority
+    5
+  end
 
   def perform(meter)
     Meters::DccChecker.new([meter]).perform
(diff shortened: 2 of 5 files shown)


ldodds (approved):

iant:
Screenshot 2022-10-05 at 12 20 46

iant:
@ldodds I've added some job prioritisation bits - could you give this another quick review before i merge please

ldodds (approved):

PR #2091, opened 2022-10-05
Create Daily Regeneration Job to replace validation and content batch…

iant:
… processes

### DIFF
--- a/app/jobs/daily_regeneration_job.rb
+++ b/app/jobs/daily_regeneration_job.rb
@@ -1,0 +1,19 @@
+class DailyRegenerationJob < ApplicationJob
+  queue_as :regeneration
+
+  def perform(school:)
+    total_amr_readings_before = AmrValidatedReading.count
+
+    GoodJob.logger.info("Validating and persisting #{school.name}...")
+    Amr::ValidateAndPersistReadingsService.new(school).perform
+
+    GoodJob.logger.info("Clearing cache for #{school.name}...")
+    AggregateSchoolService.new(school).invalidate_cache
+
+    total_amr_readings_after = AmrValidatedReading.count
+    GoodJob.logger.info("Total AMR Readings after #{school.name}: #{total_amr_readings_after} - inserted: #{total_amr_readings_after - total_amr_readings_before}")
+
+    ContentBatch.new([school], GoodJob.logger).regenerate
+    GoodJob.logger.info("Finished daily regeneration for #{school.name}")
+  end
+end
--- a/lib/tasks/schools/daily_regeneration.rake
+++ b/lib/tasks/schools/daily_regeneration.rake
@@ -1,0 +1,8 @@
+namespace :school do
+  desc "Schools daily regeneration"
+  task daily_regeneration: :environment do
+    School.process_data.each do |school|
+      DailyRegenerationJob.perform_later(school: school)
+    end
+  end
+end


app/jobs/daily_regeneration_job.rb
```diff
@@ -0,0 +1,19 @@
+class DailyRegenerationJob < ApplicationJob
+  queue_as :regeneration
+
+  def perform(school:)
+    total_amr_readings_before = AmrValidatedReading.count
```
ldodds, line 5: We don't really need this level of logging. We have this detail in the school job run by admins, to just give them some feedback about what has happened. But within the nightly job we typically just log that we've run each school, with no detail.

The `.count` can also be slow if the db stats are out of date.

Can you just simplify the logging to just record the start and end time for each job, e.g. puts "#{DateTime.now.utc} Regeneration for school #{school.name} start" and one for the end.

lib/tasks/schools/daily_regeneration.rake
```diff
@@ -0,0 +1,8 @@
+namespace :school do
+  desc "Schools daily regeneration"
+  task daily_regeneration: :environment do
```
ldodds, line 3: I've recently updated all of the other rake tasks that are run by cron so that they:

- log a message with timestamp at start
- log a message with timestamp at end
- include a rescue within loops, so that a failure processing one school does not cause the whole job to fail
- within that rescue log the exception to the log file and Rollbar
- provide the name of the batch process in the Rollbar call

In this case a failure might only be a failure to submit the job for processing, but would be good to catch anything.

Can you have a look at the jobs, e.g. `validate_amr_readings.rake` and follow that pattern?

ldodds (changes requested):

app/jobs/daily_regeneration_job.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def perform(school:)
+    GoodJob.logger.info("#{DateTime.now.utc} Regeneration for school #{school.name} start")
+    run_validate_and_persist_readings_service_for(school)
+    run_aggregate_school_service_for(school)
+    run_content_batch_for(school)
+    GoodJob.logger.info("#{DateTime.now.utc} Regeneration for school #{school.name} end")
+  end
+
+  def run_aggregate_school_service_for(school)
```
ldodds, line 12: maybe rename this and the log message, its not aggregating the school here, just invalidating the rails cache. I've never seen it thrown an error, but no harm in having the exception handler

app/jobs/daily_regeneration_job.rb
```diff
@@ -0,0 +1,19 @@
...
+    GoodJob.logger.info("Validating and persisting #{school.name}...")
+    Amr::ValidateAndPersistReadingsService.new(school).perform
+
+    GoodJob.logger.info("Clearing cache for #{school.name}...")
+    AggregateSchoolService.new(school).invalidate_cache
+
+    total_amr_readings_after = AmrValidatedReading.count
+    GoodJob.logger.info("Total AMR Readings after #{school.name}: #{total_amr_readings_after} - inserted: #{total_amr_readings_after - total_amr_readings_before}")
+
+    ContentBatch.new([school], GoodJob.logger).regenerate
```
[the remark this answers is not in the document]
ldodds: This still needs to be changed to generate

ldodds (changes requested):

ldodds (approved):

PR #2096, opened 2022-10-06
Update daily regeneration tasks to create new benchmarks

iant:


lib/tasks/schools/daily_regeneration.rake
```diff
@@ -5,6 +5,7 @@ namespace :school do
     School.process_data.each do |school|
       puts "Run daily regeneration job for #{school.name}"
       begin
+        BenchmarkResultGenerationRun.create!
         DailyRegenerationJob.perform_later(school: school)
       rescue => e
         puts "Exception: running validation for #{school.name}: #{e.class} #{e.message}"
```
ldodds, line 11: We want one BenchmarkResultGenerationRun per day. This is still creating one per school. So the `.create!` call needs to be outside the school loop.

ldodds (changes requested):

ldodds (approved):

PR #2099, opened 2022-10-06
Add list of schools tabs to manage school group admin page

urbanwide:
~~Hi @ldodds - Just wanted to check I have the right schools listed in the "active" tab. The schools I have in there at the moment are: @school_group.schools.active - should this be @school_group.schools.visible (which would include schools that are active *and* visible?). The reason why I ask is because the label "Active" is used on other areas of the same page (e.g. active school counts and pupils in active schools) and are using the @school_group.schools.visible (active and visible schools) scope.~~

~~Also - you mentioned that you're likely to want to remove the admin/school_setup page at some point - do we also want to add the "Download as CSV" functionality to the onboarding schools section of this page?~~
Addressed in comments below. Thanks!


ldodds:
@urbanwide ah yes. It should be active and visible. So that the labelling is consistent.

re: the download, I think we should check with admins. Am not sure how often its used. Suggest we leave that decision to when we decide to remove the school setup page.

app/views/admin/school_groups/_active_schools.html.erb
```diff
@@ -0,0 +1,13 @@
+<table class="table table-striped">
+  <% school_group.schools.active.by_name.each do |school| %>
```
ldodds, line 2: this should be visible, to be consistent with the counts that we're showing elsewhere.

active and visible schools should have completed their onboarding, so will be in this list.

ldodds (changes requested):

app/views/admin/school_groups/_active_schools.html.erb
urbanwide: Thanks @ldodds - have pushed active to visible here now!

ldodds (approved):

PR #2100, opened 2022-10-06
Remove generation of old analysis pages

iant:
There is an Expert Analysis page linked from the Manage School menu for each school (`/admin/schools/:school/analysis`).
which shows a tabbed view of analysis charts for the school. This is the old analysis pages and is now only visible to admins. Only the Heating Model Fitting page is used, to help review the thermostatic models. The page is driven by a list of page sections and charts that are stored in the `analysis_charts` in the `school.configuration`. This value is populated every time we run the content batch, via the `Schools::GenerateAnalysisChartConfiguration` service. The service reads a configuration for the analytics and runs every chart in that configuration, to determine which ones can be successfully run. If so then its added to the config. This was originally done so that we wouldn't display broken charts to users.  But this isn't needed any more as users don't see this page. This means we're unnecessarily running lots of charts every day. E.g. for Freshford Church School its ~18 charts.

This pr covers a first set of steps to remove this functionality:

- [x] Remove the call to `analysis_charts = analysis_chart_configuration.generate` in `GenerateConfiguration` and just set the value of the `analysis_charts` to `{}` (leave other calls in place, as we still need to run the pupil charts
- [x] Change the `Admin:AnalysisController` to always redirect to `admin_school_analysis_tab_path` even if the `analysis_charts` variable is empty
- [x] Update `admin/analysis/_nav.html` to leave only the "Heating Model Fitting" fit, so we can still view these charts

### DIFF
--- a/app/controllers/admin/analysis_controller.rb
+++ b/app/controllers/admin/analysis_controller.rb
@@ -9,14 +9,11 @@ module Admin
 
     before_action :check_aggregated_school_in_cache
     before_action :build_aggregate_school, except: [:analysis]
-    before_action :set_nav
     before_action :set_measurement_options
 
     def analysis
-      if @school.analysis?
-        # Redirect to correct dashboard
-        redirect_to admin_school_analysis_tab_path(school: @school, tab: pages.keys.first)
-      end
+      heat_meter = @aggregate_school.all_heat_meters.first
+      redirect_to admin_school_analysis_tab_path(@school, tab: :heating_model_fitting, mpan_mprn: heat_meter.mpan_mprn)
     end
 
     def show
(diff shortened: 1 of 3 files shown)


app/controllers/admin/analysis_controller.rb
```diff
@@ -9,14 +9,11 @@ class AnalysisController < AdminController
...
-    before_action :set_nav
     before_action :set_measurement_options
 
     def analysis
-      if @school.analysis?
-        # Redirect to correct dashboard
-        redirect_to admin_school_analysis_tab_path(school: @school, tab: pages.keys.first)
-      end
+      heat_meter = @aggregate_school.all_heat_meters.first
+      redirect_to admin_school_analysis_tab_path(@school, tab: :heating_model_fitting, mpan_mprn: heat_meter.mpan_mprn)
```
iant, line 16: this was redirecting to the overview page by default (now removed) so I've updated to redirect to the first meter found (and have subsequently removed all now unused nav and page code from the controller too)

PR #2101, opened 2022-10-06
New rake task to reload solar pv areas

ldodds:
Simple rake tasks that loops through all active Solar PV areas and reloads the data, from full first year of current date.

This will let us reload the Sheffield data following their updates


iant (approved):
looks good to me.  one nitpick comment on using minimum if possible.

PR #2105, opened 2022-10-06
Reduce memory footprint of schedule data

iant:


app/services/schedule_data_manager_service.rb
```diff
@@ -53,31 +88,46 @@ def temperatures
...
+          academic_year
+        )
+
+        hol_data.add(analytics_holiday)
       end
-      uk_grid_carbon_intensity_data
+      Holidays.new(hol_data)
     end
+
+    # Only use holiday data within datetime bounds of school meter readings
```
ldodds, line 119: Suggest we leave all holiday data for the moment. Its a relatively small amount of data compared to temperature, solar, etc.

app/models/school.rb
```diff
@@ -181,6 +181,31 @@ class ProcessDataError < StandardError; end
   # https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Dirty.html#method-i-will_save_change_to_attribute-3F
   after_save :add_joining_observation, if: proc { saved_change_to_activation_date?(from: nil) }
 
+  def reading_date_bounds
```
ldodds, line 184: I think we'll need all of the readings, so the upper bound should be just whatever data we have.

For example: we're loading schedule data to validate 2 months of readings for a school. If we only load the temperature data for the existing readings, then we won't have temperatures for that new data.

So suggest only clipping the lower bound

app/services/schedule_data_manager_service.rb
iant: :+1:

app/models/school.rb
iant: :+1:  have updated the school method

iant:
This pr is complete but awaiting a new analytics tag release to include https://github.com/Energy-Sparks/energy-sparks_analytics/pull/342 (also so tests will pass).  will update once tag is ready.

app/services/schedule_data_manager_service.rb
```diff
@@ -17,29 +18,63 @@ def self.calendar_cache_key(calendar)
...
+  end
 
-        hol_data.add(analytics_holiday)
+  def uk_grid_carbon_intensity
+    @uk_grid_carbon_intensity ||= find_uk_grid_carbon_intensity
+  end
+
+  private
+
+  def school_reading_date_bounds_present?
```
ldodds, line 38: This doesn't look like its called?

app/services/schedule_data_manager_service.rb
```diff
@@ -17,29 +18,63 @@ def self.calendar_cache_key(calendar)
...
+      DataFeeds::CarbonIntensityReading.all.pluck(:reading_date, :carbon_intensity_x48).each do |date, values|
+        uk_grid_carbon_intensity_data.add(date, values.map(&:to_f))
+      end
+      uk_grid_carbon_intensity_data
+    end
+
+    # Only use uk grid carbon intensity data within lower datetime bounds of school meter readings
+    return cached_uk_grid_carbon_intensity unless @school.minimum_reading_date.present?
+
+    cached_uk_grid_carbon_intensity.select { |date, _v| date >= @school.minimum_reading_date }
```
ldodds, line 72: This will return a new Hash I believe, rather than an instance of GridCarbonIntensity. The analytics will be expecting an instance of the latter, so I think this logic should be similar to the other methods where its deleting entries from the object?

spec/services/schedule_data_manager_service_spec.rb
```diff
@@ -21,13 +21,125 @@
...
 
     it 'assigns school date periods for the analytics code' do
+      allow(school).to receive(:minimum_reading_date).and_return(nil)
       results = ScheduleDataManagerService.new(school).holidays
       school_date_period = results.find_holiday(date_version_of_holiday_date_from_calendar)
       expect(school_date_period.start_date).to eq date_version_of_holiday_date_from_calendar
       expect(school_date_period.type).to_not be_nil
     end
+
+    it 'loads holiday data' do
```
ldodds, line 34: Can you expand the tests to check the return type of the objects. Part of the contract with the analytics is that we provide instances of its object, rather than just hashes.

iant:
@ldodds this one's ready for another review when you get a mo...

ldodds (approved):

PR #2106, opened 2022-10-06
Remove check for whether school is in cache on old controller

ldodds:
This controller should only be hit if a user has a bookmarked link, but its possible there are some older links on the site.

I've removed the check for whether the school is cached as this will happen after the redirect.

### DIFF
--- a/app/controllers/management/schools_controller.rb
+++ b/app/controllers/management/schools_controller.rb
@@ -2,10 +2,6 @@ module Management
   class SchoolsController < ApplicationController
     load_and_authorize_resource
 
-    include SchoolAggregation
-
-    before_action :check_aggregated_school_in_cache
-
     def show
       redirect_to school_path(@school), status: :moved_permanently
     end


app/views/schools/show.html.erb
```diff
@@ -1,6 +1,8 @@
 <%= content_for :page_title, t('schools.show.dashboard_title', school_name: @school.name) %>
 
-<%= render 'shared/dashboard_title', school: @school, co2_pages: @co2_pages, public: true, show_data_enabled_features: @show_data_enabled_features %>
+<%= render 'shared/dashboard_title2', school: @school, co2_pages: @co2_pages, public: true, show_data_enabled_features: @show_data_enabled_features %>
+
+<%#= render 'shared/dashboard_title', school: @school, co2_pages: @co2_pages, public: true, show_data_enabled_features: @show_data_enabled_features %>
```
iant, line 5: can this commented line be removed?

iant (approved):

PR #2109, opened 2022-10-07
Add csv export to good jobs report

iant:


app/views/admin/reports/good_jobs/export.csv.erb
```diff
@@ -0,0 +1,14 @@
...
+<%= CSV.generate_line headers %>
+<%- @jobs.each do |job| -%>
+<%= CSV.generate_line(
+  [
+    job['queue_name'],
+    job['job_class'],
+    job['job_id'],
+    job['performed_at'],
+    job['finished_at'],
+    job['time_to_completion_in_seconds']
```
urbanwide, line 11: Assume this is not supposed to be time_to_completion as per the header?

urbanwide: Otherwise I would suggest using job.values_at(*headers) here. Sorry if I misunderstood

iant: :+1:

urbanwide (approved):
Looks good @iant 👍

PR #2121, opened 2022-10-12
Simple admin mode

ldodds:
Adds a simple admin mode option to the website. This is to allow us to put the site into maintenance mode whilst doing database upgrades.

Setting `FEATURE_FLAG_ADMIN_MODE` will display a maintenance page on all routes, unless the user is an admin.

Allows access to the login page, to allow admins to login.

The maintenance page has a simple message and no navigation.


urbanwide (approved):
Looks good @ldodds 👍

PR #2125, opened 2022-10-13
Configure school group wide dashboard messages

urbanwide:
This PR allows admins to configure school group wide dashboard messages


app/models/dashboard_message.rb
```diff
@@ -0,0 +1,13 @@
...
+#
+# Table name: dashboard_messages
+#
+#  created_at :datetime         not null
+#  id         :bigint(8)        not null, primary key
+#  message    :text
+#  updated_at :datetime         not null
+#
+class DashboardMessage < ApplicationRecord
+  has_one :school_group, dependent: :nullify
```
ldodds, line 11: As we'll later be adding school specific messages, should this be a polymorphic association, so it could hold references to either a school or school group?

This can obviously be changed later, but wondering whether its worth doing now to avoid having to migrate any data?

ldodds (approved):

PR #2127, opened 2022-10-14
Set country as well as lat lng when geocoding

jhigman:


ldodds (changes requested):
Rather than have an after party task to geocode the schools again, I'd prefer to have that as a stand alone rake task. Avoids potential issues with long running tasks during a release.

jhigman:
Moved the code from the after party task to a rake task:

    bundle exec rails school:set_countries

ldodds (approved):

PR #2129, opened 2022-10-18
Improve school time validation

ldodds:
Recently discovered a few invalid school time records in the database. Turned out there were some issues with how the current validation rules were being applied.

This PR refactors how the rules around overlapping time periods is done, to make sure it works for both new and stored data.

Will separately tidy up existing data.

### DIFF
--- a/app/controllers/onboarding/school_times_controller.rb
+++ b/app/controllers/onboarding/school_times_controller.rb
@@ -7,7 +7,13 @@ module Onboarding
     def update
       @school = @school_onboarding.school
       @school.attributes = school_params
-      if @school.save(context: :school_times_update)
+      saved = false
+      @school.school_times.each do |st|
+        saved |= st.save
+        puts "SAVED?"
+        puts saved
+      end
+      if saved #@school.save(context: :school_time_update)
         redirect_to new_onboarding_completion_path(@school_onboarding, anchor: 'opening-times')
       else
         render :edit
--- a/app/controllers/onboarding/school_times_controller.rb
+++ b/app/controllers/onboarding/school_times_controller.rb
@@ -7,13 +7,7 @@ module Onboarding
     def update
       @school = @school_onboarding.school
       @school.attributes = school_params
-      saved = false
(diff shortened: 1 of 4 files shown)


app/controllers/onboarding/school_times_controller.rb
```diff
@@ -7,7 +7,13 @@ def edit
...
       @school = @school_onboarding.school
       @school.attributes = school_params
-      if @school.save(context: :school_times_update)
+      saved = false
+      @school.school_times.each do |st|
+        saved |= st.save
+        puts "SAVED?"
+        puts saved
+      end
+      if saved #@school.save(context: :school_time_update)
```
iant, line 16: does this commented code need removing?

app/models/school_time.rb
```diff
@@ -97,33 +94,79 @@ def to_analytics
...
+    calendar_period = overlapping_calendar_periods
+    overlapping = false
+    school.school_times.each do |other|
+      if other != self &&
+         usage_type == other.usage_type &&
+         day.include?(other.day) &&
+         calendar_period.include?(other.calendar_period) &&
+         overlapping_times?(other)
+        overlapping = true
+      end
```
iant, line 114: small nitpick but wondering if this might be easier to read as
```
overlapping = true if other != self &&
                      usage_type == other.usage_type &&
                      day.include?(other.day) &&
                      calendar_period.include?(other.calendar_period) &&
                      overlapping_times?(other)
 ```

iant: maybe also break early once overlapping is set true?

iant (approved):
looks good to me @ldodds :+1:  (couple of small nitpick comments)

app/models/school_time.rb
ldodds: Have updated, can you have another look?

iant (approved):

app/models/school_time.rb
iant: :+1: looks good to me.

PR #2130, opened 2022-10-18
Display school group wide messages

urbanwide:
This displays the school group message on the dashboard - hopefully in the right place @ldodds - let me know if it needs changing! 

In the admin / school groups index, I also added a little info symbol against school groups with a dashboard message set - just to give quick visibility on those with messages set (otherwise you have to click through to each school group to check which ones had messages). Hope that's ok, if not, I can take it out.

### DIFF
--- a/app/views/admin/school_groups/index.html.erb
+++ b/app/views/admin/school_groups/index.html.erb
@@ -18,7 +18,11 @@
     <tbody>
       <% @school_groups.each do |school_group| %>
         <tr>
-          <td><%= school_group.name %></td>
+          <td><%= school_group.name %>
+            <% if school_group.dashboard_message %>
+              <span class="badge badge-info" title="Dashboard message is shown for schools in this group: <%= school_group.dashboard_message.message %>"><%= fa_icon(:info) %></span>
+            <% end %>
+          </td>
           <td><%= school_group.school_onboardings.select(&:incomplete?).count %></td>
           <td><%= school_group.schools.visible.count %></td>
           <td><%= school_group.schools.visible.data_enabled.count %></td>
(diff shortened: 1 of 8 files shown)


app/views/admin/school_groups/index.html.erb
```diff
@@ -18,7 +18,11 @@
     <tbody>
       <% @school_groups.each do |school_group| %>
         <tr>
-          <td><%= school_group.name %></td>
+          <td><%= school_group.name %>
+            <% if school_group.dashboard_message %>
+              <span class="badge badge-info" title="Dashboard message is shown for schools in this group: <%= school_group.dashboard_message.message %>"><%= fa_icon(:info) %></span>
```
ldodds, line 23: Nice touch :)

app/views/schools/show.html.erb
```diff
@@ -44,6 +44,8 @@
 <% end %>
 
 <% if @show_standard_prompts %>
+  <%= render 'management/schools/dashboard_message', school: @school %>
```
ldodds, line 47: I think this needs to be more prominent. This list "standard prompts" is a set of reminders that are more or less always shown. As this message is presumably more urgent, then I think it needs to be higher up the page.

I'd suggest putting it immediately after the @dashboard_alerts section, and ensure its only shown to school users. Should be safe to just wrap it in the test for @show_standard_prompts.

ldodds (changes requested):

app/views/admin/school_groups/index.html.erb
urbanwide: 👍

app/views/schools/show.html.erb
urbanwide: Thanks @ldodds for the tips. Have moved it up as suggested and wrapped in a @show_standard_prompts test. I'll merge when the tests have finished running 👍

ldodds (approved):

PR #2132, opened 2022-10-19
Handle escaped parameters in rewrite

ldodds:
Turns out that during round-trip to database and converting to string, the links in the HTML description may have escaping applied. This means that link rewriting is failing for some links which have parameters.

A small change (and restructured tests) to escape the URLs before we do the replacement.


iant (approved):
:+1: looks good to me @ldodds

PR #2133, opened 2022-10-19
Add validators to school model

ldodds:
* Percentage free school meals should be between 0..100
* Postcodes should match a known format. Added case insensitive regex for handling this


iant (approved):
looks good to me @ldodds  - the SO link is pretty thorough! :D

PR #2136, opened 2022-10-19
Welsh translation snagging list 2

ldodds:
Adds translations for:

- Carbon emissions link on dashboard
- Fuel types in management table
- School timeline partials which didn't have translated timeline entries and button text.


iant (approved):
:+1: looks good @ldodds

PR #2137, opened 2022-10-19
Export detail for school groups

urbanwide:
This PR allows detail for school groups to be exported.

While I was there, I switched out all use of select(&:incomplete?) and select(&:complete?) on school onboardings for new .incomplete and .complete scopes that should be much more efficient.

I put the school export code in it's own service as it didn't feel right to fill up the model or controller with it. Hope this is ok!

Also be aware that when breaking down onboarding counts by school type, that most onboardings don't have an associated school and therefore don't have a school type either - so for example, a school group may show 3 schools onboarding in total but when being broken down by school type, have 0 for each (see example here): [school-groups-2022-10-21t15-58-45z.csv](https://github.com/Energy-Sparks/energy-sparks/files/9840657/school-groups-2022-10-21t15-58-45z.csv)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -156,4 +156,5 @@ group :test do
   gem 'webdrivers'
   gem 'simplecov', :require => false, :group => :test
   gem 'shoulda-matchers'
+  gem 'timecop'
 end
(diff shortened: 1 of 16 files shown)


ldodds:
Looks good @urbanwide. Having a separate CSV seems sensible to me and :+1: on the new scope :)

PR #2138, opened 2022-10-19
Optimise the Alert and Benchmark Generation Service

iant:
At the moment in the Content Batch / SchoolBatchRunJob we have separate steps for:

- GenerateAndSaveAlerts
- GenerateAndSaveBenchmarks

Both of these services:

- loop through a list of alerts
- run them for a specific school
- insert new records into the database

This PR creates a new service that combines these steps into one.

### DIFF
--- a/app/services/alerts/generate_and_save_alerts_and_benchmarks.rb
+++ b/app/services/alerts/generate_and_save_alerts_and_benchmarks.rb
@@ -1,28 +1,22 @@
 module Alerts
   class GenerateAndSaveAlertsAndBenchmarks
-    def initialize(
-          school:,
-          aggregate_school: AggregateSchoolService.new(school).aggregate_school,
-          benchmark_result_generation_run: BenchmarkResultGenerationRun.latest
-        )
+    def initialize(school, aggregate_school, benchmark_result_generation_run)
       @school = school
-      @aggregate_school = aggregate_school
-      @benchmark_result_generation_run = benchmark_result_generation_run
+      @aggregate_school = aggregate_school || AggregateSchoolService.new(school).aggregate_school
+      @benchmark_result_generation_run = benchmark_result_generation_run || BenchmarkResultGenerationRun.latest
     end
 
-    def perform(asof_date = Time.zone.today)
+    def perform
       ActiveRecord::Base.transaction do
-        @alert_generation_run = AlertGenerationRun.create!(school: @school)
-        @benchmark_result_school_generation_run = BenchmarkResultSchoolGenerationRun.create!(school: @school, benchmark_result_generation_run: @benchmark_result_generation_run)
+        build_alerts_and_alert_types
+      end
+    end
 
-        relevant_alert_types.each do |alert_type|
-          service = alert_type_run_result = GenerateAlertTypeRunResult.new(school: @school, aggregate_school: @aggregate_school, alert_type: alert_type, use_max_meter_date_if_less_than_asof_date: true).perform
+    def build_alerts_and_alert_types
+      # Run though all of the schools alerts
+      @alert_generation_run = AlertGenerationRun.create!(school: @school)
 
-          service.benchmark_dates(asof_date).each do |benchmark_date|
-            alert_type_run_result = service.perform(benchmark_date)
-            process_alert_type_run_result(alert_type_run_result)
-          end
-        end
+      relevant_alert_types.each do |alert_type|
       end
     end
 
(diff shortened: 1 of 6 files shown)


app/services/alerts/generate_and_save_alerts_and_benchmarks.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    def perform(asof_date = Time.zone.today)
+      ActiveRecord::Base.transaction do
+        @alert_generation_run = AlertGenerationRun.create!(school: @school)
+        @benchmark_result_school_generation_run = BenchmarkResultSchoolGenerationRun.create!(school: @school, benchmark_result_generation_run: @benchmark_result_generation_run)
+
+        relevant_alert_types.each do |alert_type|
+          if alert_type.benchmark == true
+            service = GenerateAlertTypeRunResult.new(school: @school, aggregate_school: @aggregate_school, alert_type: alert_type, use_max_meter_date_if_less_than_asof_date: true)
+            service.benchmark_dates(asof_date).each do |benchmark_date|
```
ldodds, line 18: What we need to check is whether the benchmark dates that are being used are different to whatever the default is when `perform` is called without a date. This will require checking the various alert type classes to see if there's any major differences.

app/services/alerts/generate_and_save_alerts_and_benchmarks.rb
```diff
@@ -0,0 +1,81 @@
...
+              alert_type_run_result = service.perform(benchmark_date)
+              process_alert_type_run_result(alert_type_run_result)
```
ldodds, line 20: This is creating multiple alert run results, but at the moment we just have one per run. So this isn't quite right.

What I'm hoping is that we call analyse on the alert type class once, storing the alert and benchmark data and then if its a benchmark we might call it additional times to generate the benchmark results for those extra dates.

But we need to be clear on what the benchmark dates are.

ldodds (changes requested):

ldodds:
Additional changes:

- Update `AggregateSchoolService.analysis_date` to return `meter_collection.aggregated_electricity_meters.amr_data.end_date` for `:storage_heater`, otherwise unchanged
- Set `use_max_meter_date_if_less_than_asof_date` to `false` for alerts that have no `fuel_type`

In the analytics:

- Create new sub-class of `AlertImpendingHoliday`, e.g. `AlertImpendingHolidayBenchmark` and ensure it always uses the last holiday date in its calculation

A PR that 

- adds a new AlertType record for `AlertImpendingHolidayBenchmark`, using same config as the existing one
- updates the existing entry for `AlertImpendingHoliday` to set benchmark to `false`

app/services/alerts/generate_and_save_alerts_and_benchmarks.rb
```diff
@@ -0,0 +1,79 @@
...
+      @benchmark_result_generation_run = benchmark_result_generation_run || BenchmarkResultGenerationRun.create!
+      @framework_adapter = framework_adapter
+    end
+
+    def perform
+      ActiveRecord::Base.transaction do
+        @alert_generation_run = AlertGenerationRun.create!(school: @school)
+        @benchmark_result_school_generation_run = BenchmarkResultSchoolGenerationRun.create!(school: @school, benchmark_result_generation_run: @benchmark_result_generation_run)
+
+        relevant_alert_types.each do |alert_type|
```
ldodds, line 15: Add special case for AlertImpendingHoliday to run it separately as an alert and a benchmark with different flags.

ldodds: if alert_type.class_name = AlertImpendingHoliday
          service = GenerateAlertTypeRunResult.new(
            school: @school,
            aggregate_school: @aggregate_school,
            alert_type: alert_type,
            use_max_meter_date_if_less_than_asof_date: alert_type.fuel_type.present? ? true : false
          )
          alert_type_run_result = service.perform(Time.zone.today)
          process_alert_type_run_result(alert_type_run_result)

          service = GenerateAlertTypeRunResult.new(
            school: @school,
            aggregate_school: @aggregate_school,
            alert_type: alert_type,
            use_max_meter_date_if_less_than_asof_date: true
          )
          alert_type_run_result = service.perform(Time.zone.today)
          process_benchmark_type_run_result(alert_type_run_result) if alert_type.benchmark == true
else
 ...current code
end

ldodds (changes requested):

ldodds (changes requested):

app/services/content_batch.rb
```diff
@@ -33,15 +33,16 @@ def generate_content(benchmark_result_generation_run)
...
-      # Generate alerts
-      suppress_output { Alerts::GenerateAndSaveAlerts.new(school: school, aggregate_school: aggregate_school).perform }
-
-      @logger.info "Generated alerts"
-
-      # Generate benchmarks
-      suppress_output { Alerts::GenerateAndSaveBenchmarks.new(school: school, aggregate_school: aggregate_school, benchmark_result_generation_run: benchmark_result_generation_run).perform }
-
-      @logger.info "Generated benchmarks"
+      # Generate alerts & benchmarks
```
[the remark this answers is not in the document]
iant: :+1: done

app/services/alerts/generate_and_save_alerts_and_benchmarks.rb
iant: :+1: done - tho needs double checking i have the correct logic for the separate processes for AlertImpendingHoliday

ldodds (approved):

PR #1999, opened 2022-09-07
Tidy up old Content & Alert run database content

iant:

PR #2141, opened 2022-10-21
Add rake task for database content deletion

iant:
The changes made in https://github.com/Energy-Sparks/energy-sparks/pull/1999 will work fine when we're running the database cleanup on a regular basis.

However, for the initial run, when we have ~3 years of data to delete, it's very slow and could potentially interfere with the live system.

This PR implements a separate rake task (and any other code changes) to allow us to run the database cleanup for 6/12 months at a time, so we can do this progressively and in smaller chunks. E.g. first deleting 2019 data.

We can run the rake task manually for each year and it will delete all content older than the date passed
```
bundle exec rake 'database:content_deletion['2019-12-31']'
bundle exec rake 'database:content_deletion['2020-12-31']'
bundle exec rake 'database:content_deletion['2021-12-31']'
bundle exec rake 'database:content_deletion['2022-07-01']'
```
(the last date being beginning of the month, 3 months ago as per the default in `Alerts::ContentDeletionService` - the rake task also warns and requires confirmation if a date passed is earlier than `Alerts::ContentDeletionService::DEFAULT_OLDER_THAN`)

### DIFF
--- a/lib/tasks/alerts/delete_benchmark_runs.rake
+++ b/lib/tasks/alerts/delete_benchmark_runs.rake
@@ -2,7 +2,9 @@ namespace :alerts do
   desc 'Delete benchmark runs'
   task delete_benchmark_runs: [:environment] do
     puts "#{DateTime.now.utc} Delete benchmark runs start"
-    Alerts::DeleteBenchmarkRunService.new.delete!
+    ActiveRecord::Base.transaction do
+      Alerts::DeleteBenchmarkRunService.new.delete!
+    end
     puts "#{DateTime.now.utc} Delete benchmark runs end"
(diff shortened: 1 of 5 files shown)


ldodds (approved):

lib/tasks/alerts/delete_benchmark_runs.rake
```diff
@@ -2,7 +2,9 @@ namespace :alerts do
   desc 'Delete benchmark runs'
   task delete_benchmark_runs: [:environment] do
```
iant, line 3: @ldodds before i merge this - is this rake task still needed?

PR #2148, opened 2022-10-27
Disable advice generation on charts

ldodds:
The ChartManager is responsible for generating some of the advice that is added displayed above and below the charts on the advice pages. This includes both static and dynamic text, with the latter including content generated by running additional analysis code.  By default it is always generated, but there is an option to disable it.

This wasn't being used by the application, possibly because we used to have a mixture of pages that did rely on adding some of this content dynamically to the web page. This was done in the old "Expert Analysis" pages which have now largely been removed. None of the other pages on the site now rely on fetching the `advice_header` and `advice_footer` content from the chart JSON.

This PR changes how we run the charts so by default we no longer generate the advice. A new URL parameter has been added to allow the advice to be generated on request.

The javascript has been updated to set this parameter using the existing `no-advice` setting.

The `chart_tag` has also been updated to change the `show_advice` parameter to be `false` by default. I've also updated the templates that use it to remove the parameter when not needed.

The net result is:

- only the remaining expert analysis pages now request the advice
- all other requests for charts no longer request advice by default, saving both some execution time per chart and reducing the size of the JSON returned

I've done manual testing to confirm that charts disable correctly and that the expert analysis pages still include the advice.

### DIFF
--- a/app/assets/javascripts/analysis.js
+++ b/app/assets/javascripts/analysis.js
@@ -132,6 +132,7 @@ function processAnalysisChartAjax(chartId, chartConfig, highchartsChart) {
   var dateRanges = chartConfig.date_ranges;
   var dataPath = chartConfig.jsonUrl;
   var transformations = chartConfig.transformations;
+  var noAdvice = chartConfig.no_advice;
   var requestData = {
     chart_type: chartType,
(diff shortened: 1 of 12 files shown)


iant (approved):
:+1: looks good to me

PR #2151, opened 2022-10-28
2507 allow admins group admins to record activities and actions for schools

jhigman:


app/controllers/activity_types_controller.rb
```diff
@@ -23,16 +23,23 @@ def show
...
+  def for_school
+    school = School.find(params[:school_id])
+    redirect_to new_school_activity_path(school, activity_type_id: @activity_type.id)
   end
 
   private
 
-  def can_be_completed(activity_type, school)
-    ActivityTypeFilter.new(school: school).activity_types.include?(activity_type)
+  def can_be_completed_for_schools(activity_type, schools)
```
ldodds, line 39: We could change this so that we avoid the performance overhead for admins by letting them record any activity for any school. I think that's reasonable as its a super user privilege. But for group admins we'd want to enforce the usual rules.

spec/system/activities_spec.rb
```diff
@@ -233,6 +233,77 @@
     end
   end
 
+  context 'as a group admin' do
+    let!(:group_admin)    { create(:group_admin, school: school)}
```
ldodds, line 237: This is setting up a group admin that is associated with both a school group and a school. I don't think that will happen in practice, so the tests might well not be testing the revised behaviour especially around rights to record activities?

ldodds (changes requested):

ldodds (approved):

PR #2152, opened 2022-10-28
Capture information about alternate heating sources used at schools

iant:


ldodds (approved):
Looks good, let's get this merged and onto the test server so it can be reviewed

PR #2153, opened 2022-10-28
2507 interventions recorded by admins

jhigman:


spec/system/interventions_spec.rb
```diff
@@ -69,6 +69,77 @@
     end
   end
 
+  context 'as a group admin' do
+    let!(:group_admin)    { create(:group_admin, school: school)}
```
ldodds, line 73: Same comment as on the other PR. This isn't a usual setup so can you rework the test so the group admin is never directly associated with a school, so we can be sure any permissions, school lookups, etc are workign correctly.

ldodds (changes requested):

ldodds (approved):

PR #2154, opened 2022-10-31
2480 allow schools to have issues and notes

urbanwide:
Basic model outline for notes (and issues). Has a very light use of STI to represent Issue being a subclass of Note. Wasn't sure if you meant for me to do this @ldodds when you outlined this on Trello. Happy not to use STI if you'd rather not!


urbanwide:
Just noticed fuel type can be for issues or notes, not just issues, so going to change that now

db/migrate/20221031115646_create_notes.rb
```diff
@@ -0,0 +1,16 @@
+class CreateNotes < ActiveRecord::Migration[6.0]
+  def change
+    create_table :notes do |t|
+      t.boolean :issue, default: false
+      t.string :title, null: false
+      t.text :description, null: false
```
ldodds, line 6: I think this should be a rich text field, similar to what we're using for e.g. activity types, etc.

db/migrate/20221031115646_create_notes.rb
```diff
@@ -0,0 +1,16 @@
+class CreateNotes < ActiveRecord::Migration[6.0]
+  def change
+    create_table :notes do |t|
+      t.boolean :issue, default: false
```
ldodds, line 4: For future proofing, it might be better to have an enum for the type, so easier to add new types later?

ldodds (changes requested):

ldodds (approved):

PR #2160, opened 2022-11-02
Fix temperature range error for targets

iant:
This PR adds a minimum 4 day offset for all loaded temperature schedule data.  This is related to changes in https://github.com/Energy-Sparks/energy-sparks_analytics/pull/351

### DIFF
--- a/app/services/schedule_data_manager_service.rb
+++ b/app/services/schedule_data_manager_service.rb
@@ -115,6 +115,9 @@ class ScheduleDataManagerService
   end
 
   def temperature_days_offset
+    # The projected future temperature is calculated from an average of the past temperatures at the same time
+    # of year in previous years, the number needs to be smooth and not too noisy, '4 days either side' provides
+    # this averaging, otherwise you get a much more volatile temperature adjustment.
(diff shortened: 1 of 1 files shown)


app/services/schedule_data_manager_service.rb
```diff
@@ -103,10 +105,26 @@ def find_temperatures
...
     return cached_temperatures unless use_date_bounded_schedule_data?
 
-    dates_to_remove = cached_temperatures.keys.select { |date| date < @school.minimum_reading_date }
+    school_minimum_reading_date = school_minimum_reading_date_with_temperature_days_offset
+    dates_to_remove = cached_temperatures.keys.select { |date| date < school_minimum_reading_date }
     cached_temperatures.remove_dates!(*dates_to_remove)
   end
 
+  def school_minimum_reading_date_with_temperature_days_offset
+    @school.minimum_reading_date - temperature_days_offset.days
```
ldodds, line 114: Note: we still need to change this to fix the other error I noted. That could be done in a separate PR if you prefer.

ldodds (approved):

PR #2165, opened 2022-11-02
Allow mailchimp subscription without school, using school group name …

jhigman:
…if available


iant (approved):

PR #2166, opened 2022-11-02
Schedule deletion of older alert content

ldodds:
* Add a cron job to run the content deletion service
* Remove the old cron job that run the benchmark deletion service (the above now does that too)
* Remove the old rake task for the benchmark deletion


iant (approved):
👍 looks good to me

PR #2169, opened 2022-11-03
Revise dashboard for expired targets

ldodds:
We have some schools that are coming to the end of their energy saving target period.

This is first of some improvements to application views, prompts and other functionality for expired targets.

This PR has:

* new methods on School and School Target to identify expired vs current targets, with tests
* revised logic for school dashboard so that we don't show progress reports for expired targets
* adds a new dashboard prompt to set a new target if the current one has expired


iant (approved):
looks good to me @ldodds - couple of minor comments (1 pos missing method and maybe moving a some ruby bits to the db)

PR #2170, opened 2022-11-03
New model and admin interface for storing system wide prices

iant:
The BenchmarkMetrics class in the analytics includes some hard-coded prices:

```
  ELECTRICITY_PRICE = 0.15
  SOLAR_EXPORT_PRICE = 0.05
  GAS_PRICE = 0.03
  OIL_PRICE = 0.05
```

We want to move away from hard-coded prices and instead allow these to be managed by an EnergySparks admin via the application.  This PR
- [x] adds database fields to the existing settings model to manage a list of prices.
- [x] updates the existing settings admin interface to support changing values in the application UI
- [x] Add the following text as hints
Solar Price
used to calculate £ value of exporrted solar, with existing panels for the advice page
used to calculate economic benefit of installing solar for advice page and alert
Electricity Price
used to calculate cost of a kettle, smartphone, tv, console, for equivalences
used as default electricity tariff when school has gas only, when calculating cost of point of use water heaters
Gas price
use to calculate cost of a shower for equivalences

Supporting some level of caching (so as to avoid repeated database calls) and updating the analytics code to use saved values (or defaults as a fallback) will be handled in separate PR's.


app/models/site_settings.rb
```diff
@@ -13,11 +13,29 @@
...
     order('created_at DESC').first || new
   end
 
   def temperature_recording_month_numbers
     temperature_recording_months.reject(&:blank?).map(&:to_i)
   end
+
+  def self.current_prices
+    Rails.cache.fetch(CURRENT_PRICES_CACHE_KEY) do
+      OpenStruct.new(current.prices)
```
iant, line 32: perhaps we need to default to constant values if store_accessor values are empty

iant:
See also https://github.com/Energy-Sparks/energy-sparks_analytics/pull/356 - this would need to be be merged first.

app/models/site_settings.rb
ldodds: I'd consider these values to be "seeds" so if we were starting from clean db then they'd need to be populated.

The migration adds defaults, so we should never be without the values.

app/models/site_settings.rb
```diff
@@ -13,11 +13,29 @@
 #
 
 class SiteSettings < ApplicationRecord
+  store_accessor :prices, :electricity_price, :solar_export_price, :gas_price, :oil_price
+  validates :electricity_price, :solar_export_price, :gas_price, :oil_price, numericality: { only_float: true, allow_blank: true }
```
ldodds, line 17: We're allowing blank values here. Shouldn't we make sure there's always a price set?

ldodds (changes requested):

app/models/site_settings.rb
```diff
@@ -13,11 +13,29 @@
...
   def self.current
     order('created_at DESC').first || new
   end
 
   def temperature_recording_month_numbers
     temperature_recording_months.reject(&:blank?).map(&:to_i)
   end
+
+  def self.current_prices
+    Rails.cache.fetch(CURRENT_PRICES_CACHE_KEY) do
```
ldodds, line 31: What's the default cache time here? Should we be invalidating the cache when the SiteSettings are updated (if the prices have changed)

ldodds (changes requested):

app/models/site_settings.rb
iant: :+1:

app/models/site_settings.rb
iant: no default cache time - we'll invalidate the cache after save (so runs both on create and update (just to be safe) - see `delete_current_prices_cache` l18)

app/models/site_settings.rb
ldodds: So I think this is still outstanding. There must always be a non-zero price?

ldodds (changes requested):

iant:
@ldodds yep sorry - just pushed an update for non null validation and updated the test.

iant:
^^ and fyi added a feature flag check to the analytics pr https://github.com/Energy-Sparks/energy-sparks_analytics/pull/356/files#diff-e4dd4e5bd5af7604cefb6b6617e0eb0eb93e0d56a7eefb3cf4b91de1cd441e53R24

iant:
Putting a blocked label on this one for now while @PhilipTB takes a look at what needs doing in the analytics library to use other tariffs.

iant:
@ldodds  is this ok to close now?  and ok close the related analytics pr https://github.com/Energy-Sparks/energy-sparks_analytics/pull/356 ?

db/migrate/20221103112116_add_prices_to_site_settings.rb
```diff
@@ -0,0 +1,11 @@
+class AddPricesToSiteSettings < ActiveRecord::Migration[6.0]
+  def change
+    add_column(:site_settings, :prices, :jsonb)
+    SiteSettings.current.update!(
+      electricity_price: BenchmarkMetrics::ELECTRICITY_PRICE,
```
ldodds, line 5: If this is merged before the analytics, then there won't be a price here?

iant: indeed.  i'll hard code these values instead as they'll be over-writable anyway.  :+1:

iant:
Screenshot 2023-05-09 at 14 16 46

app/views/schools/advice/solar_pv/_analysis_no_solar_table.html.erb
```diff
@@ -50,6 +53,7 @@
         <td class="text-right"><%= format_unit(scenario.solar_pv_output_co2, :co2) %></td>
         <td class="text-right"><%= format_unit(scenario.capital_cost_£, :£) %></td>
         <td class="text-right"><%= scenario.payback_years.round %> years</td>
+        <td class="text-right"><%= format_unit(scenario.export_income_£, :£) %></td>
```
iant, line 56: fyi @ldodds Export income is missing from this table (so no changes shown in the table with any changes in the site settings solar_export_price.  let me know if this needs moving/removing
Screenshot 2023-05-11 at 09 18 02

iant:
@ldodds i think this ones ready for another review when you get a mo

ldodds (approved):

PR #2185, opened 2022-11-08
Remove bullet charts from school target overview

ldodds:
Feedback has been that its not clear what the bullet charts are showing on the school targets page.

So for now we'll remove then in favour of just showing their % progress.

### DIFF
--- a/app/views/schools/school_targets/_limited_data.html.erb
+++ b/app/views/schools/school_targets/_limited_data.html.erb
@@ -4,13 +4,13 @@
   </div>
   <div class="col-md-8 align-self-center work_week d-none d-sm-block" style="padding-right: 25px">
     <div class="d-flex justify-content-center mt-2">
-      <span class="text-center">
+      <h4 class="text-center">
         <% if overview_data.present? && overview_data.work_week(fuel_type).has_data %>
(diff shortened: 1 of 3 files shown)


app/views/schools/school_targets/_limited_data.html.erb
```diff
@@ -4,13 +4,13 @@
   </div>
   <div class="col-md-8 align-self-center work_week d-none d-sm-block" style="padding-right: 25px">
     <div class="d-flex justify-content-center mt-2">
-      <span class="text-center">
+      <h4 class="text-center">
         <% if overview_data.present? && overview_data.work_week(fuel_type).has_data %>
-              <%= up_downify(overview_data.work_week(fuel_type).change) %> last week against target of <%= target %>%
+              <%= up_downify(overview_data.work_week(fuel_type).change) %> last week against target of -<%= target %>%
```
iant, line 9: spacing looks off here by a couple if chars?  (and a nitpick - maybe use `<%=-target%>` as above for consistency?)

iant (approved):
all looks good.  one comment on a bit where it looks like the spacing might be off by a couple of chars?

PR #2189, opened 2022-11-09
Button tags being removed from charts during interpolation

jhigman:


ldodds (approved):

ldodds (changes requested):
As noted in Slack, needs a bit more work/testing

config/application.rb
```diff
@@ -46,6 +46,18 @@ class Application < Rails::Application
       ActionText::ContentHelper.allowed_attributes.add 'data-chart-config'
     end
 
+    config.action_view.sanitized_allowed_attributes = Rails::Html::SafeListSanitizer.allowed_attributes.add('type')
```
ldodds, line 49: Maybe add some comments in here, around why we've added these?

PR #2191, opened 2022-11-09
Capture information about alternate heating sources used at schools

iant:


iant:
@ldodds i think this is ready for a review.  will remove all the old alt heating source code from https://github.com/Energy-Sparks/energy-sparks/pull/2152/files in a separate pr.

app/controllers/schools/your_school_estates_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    load_and_authorize_resource :school
+    before_action :return_to_school_unless_feature_enabled
+
+    def edit
+    end
+
+    def update
+      respond_to do |format|
+        if @school.update(school_params)
+          # Schools::SchoolUpdater.new(@school).after_update!
```
iant, line 12: @ldodds does this need to be enabled here?

app/models/school.rb
```diff
@@ -177,6 +177,8 @@ class ProcessDataError < StandardError; end
 
   validates_associated :school_times, on: :school_time_update
 
+  validates :alternative_heating_oil_percent, :alternative_heating_lpg_percent, :alternative_heating_biomass_percent, :alternative_heating_district_heating_percent, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 100, allow_blank: true }
```
iant, line 180: rails 6.1 has support for `validates ..., numericality: { in: 0..100 }` https://github.com/rails/rails/pull/41022

ldodds (approved):

app/views/shared/_manage_school.html.erb
```diff
@@ -3,6 +3,7 @@
   <div class="dropdown-menu" aria-labelledby="manage_school" id="manage_school_menu">
     <%= link_to t('manage_school_menu.edit_school_details'), edit_school_path(school), class: 'dropdown-item' %>
     <%= link_to t('manage_school_menu.edit_school_times'), edit_school_times_path(school), class: 'dropdown-item' if can? :manage_school_times, school %>
+    <%= link_to t('manage_school_menu.your_school_estate'), edit_school_your_school_estate_path(school), class: 'dropdown-item' if EnergySparks::FeatureFlags.active?(:your_school_estates) && can?(:manage_alternative_heating_sources, school) %>
```
ldodds, line 6: You have a new ability here `manage_alternative_heating_sources` but do we need that if the heating source information is part of the school model? The check should just be whether the user can manage the school?

If an ability is used but not assigned then only Energy Sparks admins would ever see this feature.

ldodds (changes requested):

app/controllers/schools/your_school_estates_controller.rb
iant: :+1: will remove

app/views/shared/_manage_school.html.erb
iant: :+1:  have removed to match edit school (l4 - `edit_school_path(school)`)

iant:
hi @ldodds think this one's ready for another review now...

ldodds (approved):

PR #2194, opened 2022-11-10
Allow admins to view all open issues for a school group

urbanwide:


app/views/layouts/application.html.erb
```diff
@@ -11,7 +11,7 @@
       <%= render 'shared/nav' %>
     <% end %>
 
-    <div class="application container <%= 'extra-padding' unless show_sub_nav?(@school, @hide_subnav) %>">
+    <div class="application container<%= ' extra-padding' unless show_sub_nav?(@school, @hide_subnav) %><%= ' header-fix' if header_fix_enabled? %>">
```
iant, line 14: perhaps this could be extracted to a helper? something like 
```
<div class="application container <%= conditional_application_container_classes %>">
```
```
def conditional_application_container_classes
  classes = ''
  classes += ' extra-padding' unless show_sub_nav?(@school, @hide_subnav)
  classes += ' header-fix' if header_fix_enabled?
  classes
end
```

app/helpers/nav_helper.rb
```diff
@@ -54,6 +54,10 @@ def show_sub_nav?(school, hide_subnav)
     school.present? && school.id && hide_subnav.nil?
   end
 
+  def header_fix_enabled?
+    !!@header_fix_enabled
```
iant, line 58: would `@header_fix_enabled == true` work here?

iant (approved):
This looks good @urbanwide :+1:

app/helpers/nav_helper.rb
urbanwide: Yes I think it would, thank you. (My thinking was that @header_fix_enabled might not be defined so doing it this way might cause an error. Just tested this theory though and this would be only the case for local variables and not instance variables. So I'll change it, thank you :-) )

PR #2196, opened 2022-11-10
2579 optimise s3 upload jobs

jhigman:


jhigman:
The upload of aggregated school data to S3 will be paused for the moment, and the upload of unvalidated data will be tweaked (in another PR) to lessen the load.

PR #2203, opened 2022-11-14
Assign Issues to Users

urbanwide:
When an issue is recorded or updated it should be possible to assign it to an EnergySparks admin who will be responsible for resolving it. This card covers:
- [x]  allowing an Energy Sparks admin to be assigned as the default owner of all issues for a School Group. This will be a property of the group and will be set by editing the group
- [x] updating the issue model to associate an issue with an admin user
- [x] update the forms to allow an issue to be assigned/reassigned
- [x] show the list of energy sparks admins in the options of who can be assigned an issue when it is being added/edited
- [x]  if there's a admin assigned to a school group, then default to this user when creating a new issue for the group

### DIFF
--- a/app/controllers/admin/school_groups_controller.rb
+++ b/app/controllers/admin/school_groups_controller.rb
@@ -52,6 +52,7 @@ module Admin
         :default_solar_pv_tuos_area_id,
         :default_weather_station_id,
         :default_chart_preference,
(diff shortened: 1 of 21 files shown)


iant (approved):
:+1:  looks good to me @urbanwide

PR #2205, opened 2022-11-14
Add last term date to calendar overview

ldodds:
Make it easier for admins to see at a glance how far ahead the term dates are for a regional calendar.


urbanwide (approved):
👍

PR #2206, opened 2022-11-14
Remove link to report if we have limited target data

ldodds:
If we're showing the `_limited_data.html.erb` partial then we can't currently provide a progress report for a school, so removing this link as it shouldn't be shown. Avoids users seeing an error.

Am leaving in the helper for now as I continue to rework these pages.


urbanwide:
Looks good 👍

urbanwide (approved):

PR #2211, opened 2022-11-16
Rename note model to issue

urbanwide:
Rename "note" model to "issue". Not quite as simple as it sounds as controllers, spec, views, routes, database tables etc need renaming so this touches lots of files...


iant (approved):
looks good to me @urbanwide

PR #2213, opened 2022-11-17
Updated Energy Assets data config format

iant:


lib/tasks/deployment/20221117103329_create_energy_assets2_data_feed_config.rake
```diff
@@ -0,0 +1,24 @@
+namespace :after_party do
+  desc 'Deployment task: create_energy_assets2_data_feed_config'
+  task create_energy_assets2_data_feed_config: :environment do
+    puts "Running deploy task 'create_energy_assets2_data_feed_config'"
+
+    new_config = {}
+    new_config['description'] = 'Energy Assets'
+    new_config['identifier'] = 'energy-assets2'
+    new_config['number_of_header_rows'] = 1
+    new_config['date_format'] = '%Y-%m-%d '
```
ldodds, line 10: A quick test with strptime shows that the trailing space doesn't make a difference. So suggest you remove that.

ldodds (changes requested):
See comment, suggest getting on to test so Claudia can confirm its OK.

ldodds (approved):

PR #2214, opened 2022-11-17
Allow notes to be pinned

urbanwide:


iant (approved):
looks good @urbanwide :+1:

PR #2216, opened 2022-11-17
Fix homepage case studies download button translations

iant:


urbanwide:
As discussed with Ian:

For case studies translations implementation, it was decided that two download links would be required for each case study, one to the English download file and one to the Welsh download file, if the Welsh version was available (both to be shown in both languages). The Welsh download link is in Welsh and the English download link in English as below:

Screenshot 2022-11-17 at 17 41 12

So on investigation, there is no Welsh version of the link shown because there is no Welsh case study files available to download on the live site. See: https://energysparks.uk/admin/case_studies

So the main issue is that the Welsh versions of the download files are missing.
Also, we might want to change the way the download links work at some point as it looks a bit clunky with two links. Maybe just have one link in the current language and either not displaying the download link if the file is not available in Welsh or as Ian suggested, disabling the button with "download not available" alt text.

ldodds:
Decision was to:

- only show a single button (in correct language)
- offer download of Welsh case study from Welsh interface if available, otherwise fall back to English
- on English site, offer download of English

iant:
hi @ldodds @urbanwide discussed points (above) are now addressed so I think this pr's ready for another review now.

urbanwide (approved):
Looks good @iant 👍

PR #2219, opened 2022-11-18
2579 unvalidated upload with subset

jhigman:
The cronjobs which upload aggregated schools and unvalidated data to S3 are causing server issues (probably because of the amount of yaml serialisation that they do).

The aggregated schools are used less, so have been disabled.

The unvalidated data is more useful for testing, so this PR limits the number of schools that we upload data for. 

The intention is to use the list of specific schools in the config/test_schools.yml, plus the first couple of schools from each school group, so that we have a range of data available.


iant (approved):

PR #2220, opened 2022-11-18
Fix edf historic config

iant:


app/models/amr_reading_data.rb
```diff
@@ -80,7 +80,7 @@ def invalid_row_check
 
       if reading_date.present? && valid_reading_date?(reading_date)
         warnings << :future_reading_date if future_reading_date?(reading_date)
-        warnings << :inconsistent_reading_date_format if inconsistent_reading_date_format?(reading_date)
+        # warnings << :inconsistent_reading_date_format if inconsistent_reading_date_format?(reading_date)
```
iant, line 83: disabling this for the time being as it's giving a false positive on escaped date strings

ldodds (approved):

PR #2222, opened 2022-11-18
Add inconsistent reading date format feature flag

iant:


ldodds:
Fine with adding this, but if we're having issues with specific feed configs we'll need to track them down.

ldodds (approved):

PR #2223, opened 2022-11-18
Updated solar loader period to 2 years and added spec

jhigman:


iant (approved):
:+1: All looks good to me @jhigman

PR #2224, opened 2022-11-18
2175 fix template used when previewing sent alert messages

jhigman:


iant (approved):
all looks good to me @jhigman

PR #2227, opened 2022-11-21
Make issues polymorphic

urbanwide:
This PR makes issues polymorphic & also does a bit of tidying as a result.
There will be further use of the polymorphic relationship in the next PR.
I also ran annotate --models


app/controllers/admin/issues_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      @issue.destroy
+      redirect_back_or_index notice: 'was successfully deleted'
+    end
+
+    def resolve
+      @issue = Issue.find(params[:issue_id]) # Shouldn't have to do this
+      notice = "was successfully resolved"
+      unless @issue.resolve!(updated_by: current_user)
+        notice = "Can only resolve issues (and not notes)."
+      end
```
iant, line 46: perhaps you could do a single assignment to notice here?
```
notice = if @issue.resolve!(updated_by: current_user)
           'was successfully resolved'
         else
           'Can only resolve issues (and not notes).'
         end
```

app/controllers/admin/issues_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+
+    def issueable_index_url
+      @issue.issueable ? polymorphic_url([:admin, @issue.issueable, Issue]) : polymorphic_url([:admin, Issue])
+    end
+
+    def issueable_notice(notice)
+      issueable_notice = "#{@issue.issue_type.capitalize} #{notice}"
+      issueable_notice = "#{@issue.issueable.model_name.human} #{issueable_notice}" if @issue.issueable
+      return issueable_notice
```
iant, line 67: perhaps similar here too?
```
def issueable_notice(notice)
  if @issue.issueable
    "#{@issue.issueable.model_name.human} #{issueable_notice}"
  else
    "#{@issue.issue_type.capitalize} #{notice}"
  end
end
```

app/models/issue.rb
```diff
@@ -31,7 +32,7 @@
 class Issue < ApplicationRecord
   include CsvExportable
 
-  belongs_to :school
+  belongs_to :issueable, polymorphic: true #, optional: true
```
iant, line 35: does `#, optional: true` need to be removed or uncommented here?

app/views/admin/school_groups/_issues.html.erb
```diff
@@ -11,16 +11,16 @@
...
+    <% school_group.school_issues.issue.status_open.by_pinned.by_updated_at.each do |issue| %>
       <tr>
-        <td><%= fa_icon('exclamation-circle text-secondary') %></td>
-        <td><%= link_to issue.title, admin_school_issue_path(issue.school, issue) %> <%= fa_icon('thumbtack text-secondary') if issue.pinned? %></td>
-        <td><%= link_to issue.school.name, admin_school_issues_path(issue.school) %></td>
-        <td><%= render 'admin/schools/issues/fuel_type', issue: issue %></td>
-        <td><%= render 'admin/schools/issues/owned_by', issue: issue %></td>
+        <td><%= issue_type_icon(issue.issue_type) %></td>
+        <td><%= link_to issue.title, polymorphic_path([:admin, issue.issueable, issue]) %> <%= fa_icon('thumbtack text-secondary') if issue.pinned? %></td>
+        <td><%= link_to issue.issueable.name, polymorphic_path([:admin, issue.issueable, Issue]) %></td>
```
iant, line 18: just to double check, should this be `issue` and not `Issue` on this line?  (lines 17 above and 23 below have a lower case I)

config/routes.rb
```diff
@@ -314,6 +321,7 @@
             post :make_visible
           end
         end
+        # concerns :issueable
```
iant, line 324: can this commented line be removed now?

iant (approved):
:+1: Looks great @urbanwide couple of minor suggestions (nitpicks) for possible possible improvements but looks good to me!

app/controllers/admin/issues_controller.rb
urbanwide: OK can do that, thanks :-)

app/models/issue.rb
urbanwide: I left it like this as it might be optional in the future... but I'll just make a note of the syntax elsewhere instead :-)

app/views/admin/school_groups/_issues.html.erb
urbanwide: Yes it should be Issue - polymorphic_path uses it indicate to point to the index

config/routes.rb
urbanwide: Just about to put it back in in the next PR so will leave it there for now.

PR #2230, opened 2022-11-22
Reworked school targets page

ldodds:
When schools have set a target we provide an overview page which summarises their progress, then links to a monthly report that gives more detail.

This page doesn't currently handle circumstances where the target is expired. In this case the user should be prompted to add a new target and the current page becomes a reference to how they did.

This PR does several things:

* removes the school target feature flags as they are no longer required, this simplifies some of the code and tests
* updates admin report to indicate which targets are expired
* reworks the school target page to use a table, similar to the dashboard. This provides a better summary of current and previous progress
* improves layout of prompts to add, revise targets
* adds a prompt to set a new target when required
* add a template to show summary of an expired target, with a timeline of school events
* reworked tests to provide better coverage across different users


app/views/schools/school_targets/expired.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    <%= render "prompt_to_set_new_target", school_target: @school_target %>
+
+    <p>
+      Your school set a target to reduce its energy usage between between <strong><%= nice_dates(@school_target.start_date) %></strong> and <strong><%= nice_dates(@school_target.target_date) %></strong>. This page summarises your results.
+    </p>
+
+    <% if @progress_summary.present? %>
+      <% if @progress_summary.any_passing_targets?(check_recent: false) %>
+      Well done! You achieved your goal to reduce your <%= t_fuels_as_sentence(@progress_summary.passing_fuel_targets(check_recent: false)) %> usage.
+      <% end %>
```
iant, line 17: would a safe nav operator `&.` work here (and below)? <% if @progress_summary&.any_passing_targets?(check_recent: false) %> if yes you could remove the ` <% if @progress_summary.present? %>` line above?

iant (approved):
all looks good to me @ldodds   a few comments on possible improvements for readability (nitpicks, mainly).

PR #2231, opened 2022-11-22
Update scoreboards to use mobility for names

iant:


app/controllers/admin/scoreboards_controller.rb
```diff
@@ -42,7 +42,7 @@ def destroy
   private
 
     def scoreboard_params
-      params.require(:scoreboard).permit(:name, :description, :academic_year_calendar_id, :public)
+      params.require(:scoreboard).permit(:name, :name_en, :name_cy, :description, :academic_year_calendar_id, :public)
```
urbanwide, line 45: Small thing - could t_params be used here instead?
```
translated_params = t_params(Scoreboard.mobility_attributes)
params.require(:scoreboard).permit(:name, translated_params, :description, :academic_year_calendar_id, :public)

```
 Also, is :name required too as well as :name_en and :name_cy?

app/models/scoreboard.rb
```diff
@@ -73,6 +77,10 @@ def scored_schools(recent_boundary: 1.month.ago, academic_year: this_academic_ye
 
   private
 
+  def update_name
+    self[:name] = self.name_en
```
urbanwide, line 81: Just wondering about this and why it's needed. I may have missed the point of something..

iant: I think the db name field can probably be removed at some point but currently has a non null constraint so needs to be set on save.  (I figure safer to leave in for now so easier to reverse the changes in this pr if needed)

app/controllers/admin/scoreboards_controller.rb
iant: :+1: will look at using t_params :)

app/views/admin/scoreboards/_form.html.erb
```diff
@@ -1,7 +1,8 @@
 <%= simple_form_for([:admin, scoreboard]) do |f| %>
 
   <%= f.input :academic_year_calendar_id, as: :select, collection: Calendar.national, include_blank: false, hint: 'Used for scoreboard academic years'%>
-  <%= f.input :name %>
+  <%= f.input :name_en %>
```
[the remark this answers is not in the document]
iant: :+1: yep definitely better to use locale_tabs here.  will update the pr.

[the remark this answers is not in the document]
urbanwide: 👍

app/controllers/admin/scoreboards_controller.rb
urbanwide: 👍

app/models/scoreboard.rb
urbanwide: Ahh I see :-)

iant:
hi @urbanwide updates pushed to use t params and locale tabs.  I think is ready for another review now :)

urbanwide (approved):
Looks good @iant 👍

PR #2233, opened 2022-11-23
Support alternative images by locale for programme types

iant:


app/models/programme_type.rb
```diff
@@ -15,11 +15,13 @@
 class ProgrammeType < ApplicationRecord
   extend Mobility
   include TransifexSerialisable
+  include TranslatableAttachment
```
iant, line 18: thank you for putting together this concern @urbanwide 🙏  (nice implementation too!)  made implementing the multi file uploads super simple :)

urbanwide (approved):
Looks good @iant 👍

app/models/programme_type.rb
urbanwide: Yay, glad it has been useful :-)

iant: it really was!  now I just need to do the same for activity_categories, activity types, intervention type groups, and intervention types :D

urbanwide: Hehe good luck :-)

PR #2234, opened 2022-11-23
Revise progress report pages to use stored progress reports for expired targets

ldodds:
When a target has expired we need to switch to using the stored version of the progress report, so that we're always showing consistent results.

This PR:

* Changes the url structure for the progress reports, so they're scoped under the school targets
* Retains the old controller with redirects
* Adds new template for showing expired vs current target, which doesn't contain charts
* Improves feedback reporting for both current and expired targets
* Improve test coverage


app/controllers/schools/school_targets/progress_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+            @debug_error = e.message
+            begin
+              @debug_problem = TargetsService.new(aggregate_school, @fuel_type).target_meter_calculation_problem
+              @bad_estimate = bad_estimate?(@debug_problem[:type])
+            rescue => ex
+              Rails.logger.error ex
+              Rollbar.error(ex, scope: :target_meter_calculation_problem, school_id: @school.id, school: @school.name, fuel_type: @fuel_type)
+            end
+          end
+          render :current
```
iant, line 62: perhaps lines 39-62 could be extracted into another private method (`render_school_target_current`?)?   (and maybe lines 64-69 to a private method called `render_school_target_expired`?)

app/views/schools/school_targets/progress/_footnotes.html.erb
```diff
@@ -0,0 +1,47 @@
...
+
+<div class="ml-2">
+  <% if show_storage_heater_notes %>
+    <p>
+      This report only shows progress on reducing your electricity usage. Your <%= link_to 'storage heater progress report', storage_heater_school_school_target_progress_index_path(school_target.school, school_target) %> can be viewed separately.
+    </p>
+  <% end %>
+
+  <p>
+  We use colour coding to indicate when you have <span class="bg-positive-dark">met your target</span> or <span class="bg-negative-dark">failed to meet your target</span> in a given month
```
iant, line 11: should there be a full stop at the end of this line?

iant (approved):
:+1: looks good to me @ldodds  (comment on extracting some lines to shorten a method and a possible missing fullstop but nothing major :)

PR #2235, opened 2022-11-23
Support alternative images by locale for Activity Categories

iant:


urbanwide (approved):
Looks good 👍

app/views/activity_categories/_image.html.erb
```diff
@@ -1,5 +1,7 @@
-<% if activity_category.image.present? %>
-  <%= image_tag activity_category.image, class: "card-img-top img-fluid " + css_class %>
+<% if I18n.locale.to_s == 'cy' && activity_category.t_attached(:image, :cy).present? %>
```
urbanwide, line 1: If you end up with this pattern quite a bit for all these types of changes, maybe there could be a helper for translated images?

iant: :+1: @urbanwide I think that would be good idea.  may even be a simple as a single shared _image partial.  I'll make a note on the card to do that at the end once all pr's are merged down.

urbanwide: Sounds like a good plan @iant 👍

PR #2236, opened 2022-11-23
Add school targets to main school timeline

ldodds:
Adds support for including school targets into the schools timeline of activities

* adds a new `:school_target` `Observartion` type and links to `SchoolTarget` model
* automatically add an Observation when creating a target
* add template for displaying school target observations in timeline
* exclude these observations from the custom timeline we should on the expired target page
* adds an after party task to create observations for all existing targets, to populate the database

### DIFF
--- a/app/models/school_target.rb
+++ b/app/models/school_target.rb
@@ -5,8 +5,10 @@
 #  created_at               :datetime         not null
 #  electricity              :float
 #  electricity_progress     :json
+#  electricity_report       :jsonb
 #  gas                      :float
 #  gas_progress             :json
+#  gas_report               :jsonb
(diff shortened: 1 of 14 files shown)


app/models/school_target.rb
```diff
@@ -119,4 +127,20 @@ def must_have_one_target
...
+      school: school,
+      observation_type: :school_target,
+      school_target: self,
+      at: start_date,
+      points: 0
+    )
+  end
+
+  def ensure_observation_date_is_correct
+    observations.each do |observation|
```
iant, line 142: are you able to use `observations.update_all!(at: start_date)` here?

iant (approved):
:+1: looks good to me @ldodds

PR #2239, opened 2022-11-24
Remove images from Intervention Type Groups

iant:


iant:
@ldodds @urbanwide I'm not sure on this one - it doesn't look like Intervention Type Group images are used anywhere except on the form? (I may be wrong) Should we remove instead?

app/views/admin/intervention_type_groups/show.html.erb
```diff
@@ -1,5 +1,5 @@
 <h1>Intervention Category</h1>
-<p><strong>Title:</strong> <%= @intervention_type_group.title %></p>
+<p><strong>Title:</strong> <%= @intervention_type_group.name %></p>
```
urbanwide, line 2: Wondering why this changed?

iant: sorry @urbanwide  I should have left a comment explaining that.  it's a bug fix - no `title` on InterventionTypeGroups 
https://rollbar.com/energysparks/EnergySparksProduction/items/1642/
(prob should have gone in a different pr but i only noticed the issue when testing this one)

urbanwide: Ahh cool, makes sense :-) Thanks for explaining

PR #2257, opened 2022-11-29
Translate onboarding completion page

iant:


config/locales/common.yml
```diff
@@ -19,8 +19,11 @@ en:
       edit: Edit
       manage: Manage
       next: Next
+      no_label: 'No'
```
iant, line 22: `i18n-task normalize` changes `no:` to `false` and `yes` to `true` for some reason 🤷‍♂️  hence the `_label` append here

config/locales/views/onboarding/onboarding.yml
```diff
@@ -12,6 +12,82 @@ en:
...
+        mpan_mprn: MPAN/MPRN
+        name: Name
+        opening_times: Opening times
+        optional_steps: Optional steps
+        optional_steps_message: It's not necessary for you to fill in the following sections to complete the sign-up process, but the more information you provide the quicker your school can start using Energy Sparks and our energy calculations will become more accurate.
+        password: Password
+        review_and_update_your_answers: Review and update your answers before completing setup
+        role: Role
+        school_meters_empty_message: If you have the details of your energy meters to hand then you can input them here. If not, we can get your meter details from your Local Authority, Multi-academy Trust or school group.
+        section_1:
```
ldodds, line 37: I wonder if the sections could have names that match what they're showing, e.g. `.admin_account`, `.school_details`, etc. If we reorganise these as partials, or reorder the page the number might become awkward.

ldodds (changes requested):

PR #2260, opened 2022-11-30
Translate school contact management page

iant:


app/views/schools/contacts/index.html.erb
```diff
@@ -1,32 +1,36 @@
...
-  <%= @school.name %> Alert Contacts
+  <%= t('schools.contacts.index.title', school_name: @school.name) %>
 <% end %>
 
-<h1 property="name"><%= link_to @school.name, @school %> Alert Contacts</h1>
+<h1 property="name">
+  <%= link_to @school.name, @school %> <%= t('schools.contacts.index.alert_contacts') %>
+</h1>
 
 <%= render 'shared/contact_explainer' %>
```
iant, line 9: Text extracted for this in a seperate pr https://github.com/Energy-Sparks/energy-sparks/pull/2261

app/views/schools/contacts/edit.html.erb
```diff
@@ -2,21 +2,21 @@
...
 <% if @contact.user == current_user %>
   <h2>
-    Update contact details for email or text alerts
+    <%= t('schools.contacts.edit.update_contact_details_for_email_or_text_alerts') %>
   </h2>
   <%= render 'shared/contact_explainer' %>
 <% else %>
   <h2>
-    Update contact details <%= "for #{@contact.user.name}" if @contact.user %>
+    <%= t('schools.contacts.edit.update_contact_details_for', user_name: @contact.user.name) if @contact.user %>
```
urbanwide, line 10: I think this used to append "for #{@contact.user.name}" if a @contact.user is present. Not sure the logic you have replaced it with is the same - looks like it won't display the whole line now if there is no @contact.user? Is this right?

app/views/schools/contacts/new.html.erb
```diff
@@ -2,18 +2,18 @@
...
 <% if @contact.user == current_user %>
   <h2>
-    Sign up for email or text alerts
+    <%= t('schools.contacts.new.sign_up_for_email_or_text_alerts') %>
   </h2>
   <%= render 'shared/contact_explainer' %>
 <% else %>
   <h2>
-    Enable alerts <%= "for #{@contact.user.name}" if @contact.user %>
+    <%= t('schools.contacts.new.enable_alerts_for', user_name: @contact.user.name) if @contact.user %>
```
urbanwide, line 10: Same as above, the logic is not the same as the original. "Enable alerts" also won't be displayed if there is no @contact.user. Not sure if this is intentional or not?

app/views/schools/contacts/edit.html.erb
iant: oops yep!  good spot.  i'll fix that.   Thanks @urbanwide

iant: hi @urbanwide I've pushed a fix for this and the same in new.html.erb.  would you mind giving another quick review before i merge.  thanks for spotting this :)

urbanwide (approved):
Looks good @iant 👍

PR #2261, opened 2022-12-01
Translate contact explainer

iant:


config/locales/views/shared/shared.yml
```diff
@@ -35,6 +35,19 @@ en:
   common_nav_bar_menus:
     actions: Actions
     activities: Activities
+  contact_explainer:
```
urbanwide, line 38: I've not done a lot of translation stuff, so I could be wrong but would this be better split into 3? No worries if not :-)

iant: :+1: I've pushed an update extracting to 3 sections.  could you give another quick review when you get a mo please.  Thanks

PR #2269, opened 2022-12-05
Add ActiveRecord::RecordNotFound to filter

ldodds:
Add ActiveRecord::RecordNotFound exceptions to ignore list.

Rollbar currently logs these as warnings. But we're getting quite a few of them and they're either from crawlers accessing scoreboards or schools that no longer exist, or crawlers/bots  requesting random ids. So lets exclude for now.

https://docs.rollbar.com/docs/ruby#exception-level-filters


iant (approved):
:+1:

PR #2270, opened 2022-12-06
Translate managing school and community use times form

iant:


app/views/schools/times/_school_time_fields.html.erb
```diff
@@ -35,7 +35,7 @@
           <%= f.select :calendar_period, options_for_select(SchoolTime.calendar_periods.map {|key, value| [key.titleize, SchoolTime.calendar_periods.key(value)]}, f.object.calendar_period), {required: :true}, { class: 'form-control' } %>
```
ldodds, line 35: There's a couple of select boxes on this form that display names of days and names of calendar periods. Currently these are just displaying the enums. So these values need looking up somewhere?

ldodds (changes requested):

app/views/schools/times/_school_time_fields.html.erb
iant: :+1:  oops missed those - I've just pushed an update to support translation for the school time enums.

jhigman (approved):
Looks good

ldodds (approved):

PR #2271, opened 2022-12-06
Fix flickering tests

urbanwide:
Background: This build failed (which passed locally): https://github.com/Energy-Sparks/energy-sparks/actions/runs/3628959858/jobs/6120599772

For the issues tests, I am wondering if sometimes that let! and before statements, if in the same block, do not always execute in their given order. So I am thinking in the case of the issues tests, that sometimes the test data might not have been setup before the page loaded.

./spec/system/activity_types_spec.rb:49 & ./spec/system/intervention_types_spec.rb:55
For these two, it could be because when doing a search, postgres is calculating a rank for the results, and then ordering by this rank. Wonder if this can be different sometimes.

./spec/system/activity_types_spec.rb:86
For this one, I think it could be a similar issue as for issues tests. i.e. the let! and before execution order not being consistent

### DIFF
--- a/spec/system/activity_types_spec.rb
+++ b/spec/system/activity_types_spec.rb
@@ -5,7 +5,7 @@ describe 'activity types', type: :system do
   context 'activity types search page' do
 
     let!(:activity_type_1) { create(:activity_type, name: 'foo', description: 'activity') }
-    let!(:activity_type_2) { create(:activity_type, name: 'bar', description: 'activity the second') }
+    let!(:activity_type_2) { create(:activity_type, name: 'bar', description: 'second activity') }
 
     it 'links from activity categories page and shows empty page' do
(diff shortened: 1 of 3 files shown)


ldodds:
There's definitely an ordering dependency, see e.g.:
https://www.joshmcarthur.com/til/2018/03/28/til-let-and-before-order-is-important.html

But am not sure if that's what was happening here. 

I think for the search test, its possible that the search ranking for the two results is close enough that sometimes a different one is ranked higher? Having more differences in the text to impact the ranking should help.

ldodds (approved):

urbanwide:
Thanks Leigh. It's so hard to debug intermittent failures. I agree on let!/ before ordering should be as given, but I'm sure I've seen this issue before but could be wrong.

For the search issues - I agree yes. I've changed the second item to *I hope* rank lower. Just hope it's low enough and it's not doing some rounding to rank them the same.

PR #2272, opened 2022-12-06
Log fewer target errors for schools with limited data

ldodds:
Currently we're logging too many exceptions from the `GenerateProgressService`. This is because we are always trying to generate a progress report, even if we don't have enough data for a specific fuel type.

This PR refactors the service to check whether there is enough data before trying to generate a progress report.

However I've added logging in case there are errors checking for whether there is enough data, which has been a problem in the past.

As part of this refactoring and tracking down the errors I've also:

- removed an old Wisper event listener which was updating a timestamp. The `GenerateProgressService` does this directly, so its not needed. This will also help with debugging as currently the timestamp is still being set
- fixed another bug relating to the month roll-over. During the first few days of the month we may not have enough data to generate their estimated progress, so we fall back to the previous months data

There may still be some data-related issues to work through here. But this fixes one issue and should reduce number of unnecessary exceptions.


app/services/targets/generate_progress_service.rb
```diff
@@ -82,6 +78,20 @@ def has_target_for_fuel_type?(fuel_type)
...
+    end
+
+    def fetch_latest_figures(hash_of_months_to_values)
+      val = hash_of_months_to_values[reporting_month]
+      #sometimes the schools meter data may be lagging only a few days or a week
+      #behind. this means that the progress report does not have data for this month,
+      #it only has data for the previous month. So if there's no entry for the reporting
+      #month, look for earlier data. This typically only happens around the beginning of
+      #month when we're running a little behind on data
+      val.present? ? val : hash_of_months_to_values[reporting_month.prev_month]
```
iant, line 92: I might be wrong here but maybe lines 86 and 92 could be replaced with
`hash_of_months_to_values[reporting_month] || hash_of_months_to_values[reporting_month.prev_month]`
?

iant (approved):
Looks good @ldodds 1 comment on a possible change to the `fetch_latest_figures` method

PR #2273, opened 2022-12-06
Fix trix non-scrolling issue for notes/issues

urbanwide:
Also makes sure page flow is correct after validation fail iterations (spotted while doing this fix).


iant (approved):
:+1: Looks good to me @urbanwide

PR #2278, opened 2022-12-07
Translate school user management forms (users, cluster users)

iant:


app/views/schools/users/_form.html.erb
```diff
@@ -1,3 +1,3 @@
-<%= f.input :name, required: :true %>
-<%= f.input :email, required: :true %>
-<%= f.input :staff_role_id, label: 'Role', required: :true, collection: StaffRole.order(:title), label_method: :translated_title, hint: 'What is their primary role in relation to Energy Sparks?'%>
+<%= f.input :name, required: :true, label: t('schools.users.form.name') %>
+<%= f.input :email, required: :true, label: t('schools.users.form.email') %>
+<%= f.input :staff_role_id, label: t('schools.users.form.role'), required: :true, collection: StaffRole.all.map { |staff_role| [staff_role.translated_title, staff_role.id] }.sort, hint: t('schools.users.form.what_is_their_primary_role') %>
```
jhigman, line 3: Maybe could use a factored-out helper..?

app/views/schools/users/index.html.erb
```diff
@@ -21,13 +21,13 @@
             <td><%= user.email %></td>
             <td><%= y_n(user.confirmed?) %></td>
             <td><%= y_n(user.contact_for_school) %></td>
```
jhigman, line 23: Wondering whether the "y_n" function should have a translated version.. what do you think, have we had the same elsewhere?

iant: hi @jhigman translation support to the y_n method was in a pr merged earlier in the week while you were away so should be all ok now :)  it looks like this now 
```
  def y_n(boolean)
    boolean ? I18n.t('common.labels.yes_label') : I18n.t('common.labels.no_label')
  end 
```

jhigman (approved):

PR #2280, opened 2022-12-07
Transifex pull (07/12/2022)

iant:


jhigman:
Is this just to pull down the latest translations? Looks good.

PR #2282, opened 2022-12-08
Allow user to set preferred default language

iant:
This pr provides functionality to set a users preferred language (locale)
- [x] Add preferred language to User
- [x] Allow user to set preferred language when confirming account
- [x] Allow a school admin to update the preferred language for a user
- [x] When a user completes a login, redirect to their preferred language website
- [x] Add a feature flag to disable the login redirect by default

test with
```
FEATURE_FLAG_REDIRECT_TO_PREFERRED_LOCALE=true rails s
```


app/controllers/application_controller.rb
```diff
@@ -11,6 +11,15 @@ class ApplicationController < ActionController::Base
     redirect_to root_url, alert: exception.message
   end
 
+  def after_sign_in_path_for(user)
```
ldodds, line 14: Can you add a spec for this please?

The spec should exercise the basic behaviour and also check the existing redirects we do. See HomeController.redirect_if_logged_in.

E.g. spec should check that a) user gets directs to Welsh site and b) then gets redirected to school home page or whatever destination is correct.

config/locales/views/onboarding/onboarding.yml
```diff
@@ -171,6 +171,8 @@ en:
         title: Edit account
         update_account: Update account
       form:
+        preferred_locale: Preferred language
+        preferred_locale_hint: This will be the language website and email content will be in for this user
```
ldodds, line 175: "Set your preferred language for using the website and receiving emails"

config/locales/views/schools/schools.yml
```diff
@@ -388,6 +388,8 @@ en:
       form:
         email: Email
         name: Name
+        preferred_locale: Preferred language
+        preferred_locale_hint: This will be the language website and email content will be in for this user
```
ldodds, line 392: "Set the user's preferred language for using the website and receiving emails"

ldodds (changes requested):

app/controllers/application_controller.rb
ldodds: Also, thinking we only need to do this redirect if they user isn't already on the right site?

ldodds:
@iant another couple of thoughts on this one. It looks like a school admin can set the preferred language for a user when they add them during onboarding or later in the process.

Perhaps the default option for that field should be whatever locale the user is currently using. So, e.g. a Welsh admin would see that Welsh language is selected by default for any new users they add.

Secondly, just wanted to check that when a user confirms their account that they can choose their preferred language at that point, alongside the other options to sign-up to newsletter, alerts, etc. This will allow the user to change their preference during the confirmation.

config/locales/views/onboarding/onboarding.yml
iant: :+1: done

config/locales/views/schools/schools.yml
iant: 👍 done

app/controllers/application_controller.rb
iant: :+1: some tests added.  i think we're ok here for the redirect will resolve to the same path.

iant:
hi @ldodds - you wrote:
```
@iant another couple of thoughts on this one. It looks like a school admin can set the preferred language for a user when they add them during onboarding or later in the process.

Perhaps the default option for that field should be whatever locale the user is currently using. So, e.g. a Welsh admin would see that Welsh language is selected by default for any new users they add.
```
I've pushed an update to add the selector to `app/views/onboarding/users/_form.html.erb` with the selected value being the `current_user`s (admin) preferred locale if it's a new user record (new) or the user's preferred locale if the user is being edited.

```
Secondly, just wanted to check that when a user confirms their account that they can choose their preferred language at that point, alongside the other options to sign-up to newsletter, alerts, etc. This will allow the user to change their preference during the confirmation.
```
I've also pushed an update to add the selector to `app/views/devise/passwords/edit.html.erb` - afaics this seems to be the only place where the checkbox for alerts and newsletters is?

I'll make a note to add alerts and newsletters confirmation checkboxes on the user profile form ticket.

ldodds (approved):

PR #2283, opened 2022-12-08
Content in consent statements is translated and displayed

jhigman:


iant (approved):
:+1: Looks good to me @jhigman

PR #2291, opened 2022-12-09
Translate onboarding emails

iant:


config/locales/mailers/mailer.yml
```diff
@@ -19,19 +19,75 @@ en:
...
+      join_a_webinar: Join a webinar
+      need_some_help: Need some help?
+      paragraph_1_html: You can now use Energy Sparks to begin recording your energy saving activities. Explore our pupil-lead <strong><a href="%{activity_categories_url}">activities and educational resources</a></strong>which are designed to teach your pupils about energy whilst saving money for their school.
+      paragraph_2_html: We also have a range of <strong><a href="%{intervention_type_groups_url}">energy saving actions</a></strong> that you, other members of staff and volunteers can carry out to start improving you energy usage.
+      paragraph_3_html: View your <a href="%{school_url}">school dashboard</a> for suggestions for how to get started.
+      paragraph_4: We will be in touch shortly to confirm when we have completed getting access to your energy data.
+      paragraph_5_html: For help using Energy Sparks please read our <a href="https://energysparks.uk/resources/12/download">User Guide</a>, watch our <a href="%{user_guide_videos_url}">Training Videos</a> or join one of our <a href="%{training_url}">live webinar training</a> sessions.
```
urbanwide, line 30: Assume the translators be able provide the Welsh url here?

urbanwide: (and the same for the other urls in the yaml?)

spec/mailers/onboarding_mailer_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+        expect(email.body.to_s).to include(email_content.gsub('%{school_name}', 'Test School'))
+      end
+    end
+  end
+
+  describe '#completion_email' do
+    it 'sends the completion email' do
+      OnboardingMailer.with(emails: ['test@blah.com'], school_onboarding: school_onboarding).completion_email.deliver_now
+      email = ActionMailer::Base.deliveries.last
+      expect(email.subject).to eq("Test School has completed the onboarding process")
```
urbanwide, line 23: Just wondered if you could use the subject from the yaml file here instead - means you don't have to change the test later if the yaml changes?

spec/mailers/onboarding_mailer_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+        expect(ActionController::Base.helpers.sanitize(email.body.to_s)).to include(email_content.gsub('%{school_name}', 'Test School'))
+      end
+    end
+  end
+
+  describe '#activation_email' do
+    it 'sends the activation email' do
+      OnboardingMailer.with(to: 'test@blah.com', emails: ['test@blah.com'], school: school).activation_email.deliver_now
+      email = ActionMailer::Base.deliveries.last
+      expect(email.subject).to eq("Test School is live on Energy Sparks")
```
urbanwide, line 45: As above...

urbanwide (changes requested):
Looks good @iant  - Just a couple of minor things ...

config/locales/mailers/mailer.yml
iant: hi @urbanwide the resource user guide file is English only at the moment (I think the plan is to link to the en and cy ones in drive directly) so we can keep this link as is for the time being.  however, we do have a `user_guide_download_path` method for that specific file so will push an update so that's used here instead.

spec/mailers/onboarding_mailer_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    it 'sends the welcome email' do
+      OnboardingMailer.with(emails: ['test@blah.com'], user: user).welcome_email.deliver_now
+      email = ActionMailer::Base.deliveries.last
+      expect(email.subject).to eq("Welcome to Energy Sparks")
+      I18n.t('onboarding_mailer.welcome_email').except(:subject).values.each do |email_content|
+        expect(ActionController::Base.helpers.sanitize(email.body.to_s)).to include(
+          email_content.gsub('%{school_name}', 'Test School').gsub('%{contact_url}', 'http://localhost/contact')
+                       .gsub('%{contact_url}', 'http://localhost/contact')
+                       .gsub('%{activity_categories_url}', 'http://localhost/activity_categories')
+                       # .gsub('%{intervention_type_groups_url}', 'http://localhost/intervention_type_groups')
```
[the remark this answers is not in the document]
iant: nope will remove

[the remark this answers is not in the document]
urbanwide: 👍

config/locales/mailers/mailer.yml
urbanwide: OK thanks for clarifying 👍

PR #2295, opened 2022-12-12
Remove unnecessary SchoolTarget observations

ldodds:
A silly mistake(*) in an `after_save` method means we're creating duplicate `school_target` observations for schools.

This PR adds:

- check in SchoolTarget model to see if an observation already exists
- update to the spec
- after party task to remove all but one of the existing observations for a target

(*) by me!

### DIFF
--- a/lib/tasks/deployment/20221212165015_remove_duplicate_school_target_observations.rake
+++ b/lib/tasks/deployment/20221212165015_remove_duplicate_school_target_observations.rake
@@ -1,0 +1,18 @@
+namespace :after_party do
+  desc 'Deployment task: remove_duplicate_school_target_observations'
+  task remove_duplicate_school_target_observations: :environment do
+    puts "Running deploy task 'remove_duplicate_school_target_observations'"
+
+    SchoolTarget.all.each do |school_target|
+      obs = Observation.where(school_target: school_target).order(created_at: :desc)
+      if obs.length > 1
+        obs[0..-2].each do |obs| obs.destroy end
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


lib/tasks/deployment/20221212165015_remove_duplicate_school_target_observations.rake
```diff
@@ -0,0 +1,18 @@
+namespace :after_party do
+  desc 'Deployment task: remove_duplicate_school_target_observations'
+  task remove_duplicate_school_target_observations: :environment do
+    puts "Running deploy task 'remove_duplicate_school_target_observations'"
+
+    SchoolTarget.all.each do |school_target|
+      obs = Observation.where(school_target: school_target).order(created_at: :desc)
+      if obs.length > 1
+        obs[0..-2].each do |obs| obs.destroy end
```
urbanwide, line 9: Just checking: assume there can only be one observation per school target?

ldodds: Yes, there should be. I'm just keeping the oldest here.

I've not added a validation rule to the model here as it'll invalidate the existing data until we've done this. Could tighten up later?

urbanwide: Ahh that makes sense. Was just checking incase this accidentally deleted loads of data it shouldn't have.

All good to go I think for now and good to tighten up later 👍

urbanwide (approved):
Looks good @ldodds 👍

PR #2305, opened 2022-12-14
Swap date_field for date picker (2)

iant:


app/views/admin/jobs/_form.html.erb
```diff
@@ -8,7 +8,12 @@
 
   <div class="form-group">
     <%= f.label :closing_date, "Closing date", class: "form-label" %>
-    <%= f.date_field :closing_date, class: 'form-control' %>
+    <div class="input-group date" id="datetimepickerform" data-target-input="nearest">
+      <%= f.text_field  :closing_date, class: 'form-control datetimepicker-input', data: { target: "#datetimepickerform" }, value: job.closing_date&.strftime('%d/%m/%Y') %>
+      <div class="input-group-append" data-target="#datetimepickerform" data-toggle="datetimepicker">
+        <div class="input-group-text"><i class="fa fa-calendar"></i></div>
+      </div>
+    </div>
```
iant, line 16: @ldodds this would make a good view component - maybe one for a seperate ticket/pr?

app/views/admin/jobs/_form.html.erb
```diff
@@ -31,3 +36,13 @@
   <%= f.submit job.persisted? ? "Update" : "Create", class: "btn btn-primary" %>
   <%= link_to "View all jobs", admin_jobs_path, class: "btn btn-primary" %>
 <% end %>
+
+<script type="text/javascript">
```
urbanwide, line 40: Wonder if this code ought to go in a js file somewhere as it's used a couple of times?

app/views/admin/jobs/_form.html.erb
urbanwide: I would put this in a partial at least if it's easy enough? (assuming it's going to be re-used?)

PR #2308, opened 2022-12-14
Synchronise email and sms alert template content with Transifex

iant:
We need to synchronise the templates used to build the weekly alert emails with Transifex.

This content is held in the AlertTypeRatingContentVersion model, in the `email_title` and `email_content` attributes.

We already synchronise some fields from this model with Transifex, for the messages that appear on the pupil dashboard, school dashboard and management priorities section of the school dashboard.

This pr covers extending the existing synchronisation and translation support to:

- [x] the `email_content` and `email_title` fields. 
- [x] the `sms_content` field
- [x] ensuring the `TransifexSerialisable` settings for the model cover the extra fields
- [x] revising the `tx_resources` method so that it selects not just ratings with text for the dashboard, but those where there is active email text
- [x] updating the content management editor so that the Welsh language text for the email is displayed to admins.


db/migrate/20221214140623_add_translated_alert_type_rating_content_version_email_title.rb
```diff
@@ -0,0 +1,7 @@
+class AddTranslatedAlertTypeRatingContentVersionEmailTitle < ActiveRecord::Migration[6.0]
+  def change
+    AlertTypeRatingContentVersion.all.each do |alert_type_rating_content_version|
+      alert_type_rating_content_version.update(email_title_en: alert_type_rating_content_version[:email_title])
+    end
+  end
+end
```
jhigman, line 7: I think we've previously done this sort of data migration in after_party tasks, like 20220629153807_copy_programme_type_fields_to_mobility.rake

(I think the general approach was to keep schema changes in the migrations, but data changes - especially with AR - in the after_party tasks).

Does this achieve the same thing?

db/schema.rb
```diff
@@ -1705,6 +1705,7 @@
     t.datetime "confirmation_sent_at"
     t.bigint "school_group_id"
     t.string "unlock_token"
+    t.string "preferred_locale", default: "en", null: false
```
jhigman, line 1708: Wondering if this actually belongs with a different PR..?

iant: oops yep.  will remove

db/migrate/20221214140623_add_translated_alert_type_rating_content_version_email_title.rb
iant: hi @jhigman yep does the same thing but I'll move to after_party tasks (email & sms fields could be updated in a single migration too)

iant: hi @jhigman I've added the after_party task and removed the migrations.  would you be able to give another review when you get a mo please 🙏   thanks!

ldodds (approved):

PR #2316, opened 2022-12-15
Guard against missing staff roles when sorting users for bill requests

jhigman:


iant (approved):
looks good to me @jhigman

PR #2322, opened 2022-12-18
CSV download of list of meters for a datasource

urbanwide:


app/controllers/admin/data_sources_controller.rb
```diff
@@ -1,9 +1,18 @@
 module Admin
   class DataSourcesController < AdminController
-    include Pagy::Backend
```
urbanwide, line 3: Don't need this so removed

iant (approved):
Looks good to me @urbanwide :+1:

PR #2332, opened 2022-12-22
Add date picker view component

iant:
This PR is a small refactor of the changes in https://github.com/Energy-Sparks/energy-sparks/pull/2305 to use a view component instead.

Changes in this pr also include some view component configuration so that any files with the `_component.js` suffix (such as app/components/widget_component.js) will be compiled into the Webpack bundle (as per https://viewcomponent.org/guide/javascript_and_css.html). This should also work for css files but css remains untested.


iant:
hi @urbanwide I've updated the pr to scope the javascript to the associated date picker component (as per your suggestion yesterday).  think this is ready for another review now when you get a mo please 🙏

urbanwide (approved):
Looks good @iant 👍

PR #2343, opened 2023-01-06
Initial controller and views, task to add baseload advice page

jhigman:


yarn.lock
```diff
@@ -1787,9 +1787,9 @@ caniuse-api@^3.0.0:
     lodash.uniq "^4.5.0"
 
 caniuse-lite@^1.0.0, caniuse-lite@^1.0.30000981, caniuse-lite@^1.0.30001109, caniuse-lite@^1.0.30001219:
-  version "1.0.30001228"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001228.tgz#bfdc5942cd3326fa51ee0b42fbef4da9d492a7fa"
-  integrity sha512-QQmLOGJ3DEgokHbMSA8cj2a+geXqmnpyOFT0lhQV6P3/YOJvGDEwoedcwxEQ30gJIwIIunHIicunJ2rzK5gB2A==
+  version "1.0.30001442"
```
urbanwide, line 1790: Just checking this isn't accidental - assume it was needed to get your environment working again?

spec/system/schools/advice_pages_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+        expect(page).to have_link('Learn More')
+      end
+    end
+
+    it 'shows breadcrumb' do
+      click_on key
+      within '.advice-page-breadcrumb' do
+        expect(page).to have_link('Schools')
+        expect(page).to have_link(school.name)
+        expect(page).to have_link('Advice')
```
urbanwide, line 52: Should we check for the selected item here too? i.e.
`expect(page).to have_text(@advice_page.key)
`

urbanwide (commented):
Looks good @jhigman, just a couple of small queries...

urbanwide (approved):

app/views/schools/advice/_breadcrumb.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="advice-page-breadcrumb">
```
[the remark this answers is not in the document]
jhigman: Yes you're right..! I'll remove it.

spec/system/schools/advice_pages_spec.rb
jhigman: Good idea.

yarn.lock
jhigman: Yes this was from the reinstalls - thought I would leave it there as it seemed like and updated version, but can remove if it causes problems..

urbanwide: I'm sure it'll be alright - yes can always revert if need be later

PR #2345, opened 2023-01-10
Add nvmrc to gitignore

iant:
This pr adds `.nvmrc` to the `.gitignore` file 

I've also added a note on https://github.com/Energy-Sparks/energy-sparks/wiki/Setting-up-a-developer-environment on using a node version manager

`Using a node version manager, like [nvm](https://github.com/nvm-sh/nvm) (see also [here](https://github.com/nvm-sh/nvm#bash) for bash and zsh scripts to automatically call nvm use using a .nvmrc file in the project directory)`


iant:
@urbanwide @jhigman @ldodds Alternatively we could include the `.nvrmc` file with the repo so we're aligned on which version of node to use (e.g. my `.nvrmc` is currently set to `14`).  what do you think?

iant:
hi @jhigman I've updated the pr to add the .nvmrc file if you want to push the other dotfile I'll get this merged

urbanwide (approved):
Looks good @iant 👍

PR #2347, opened 2023-01-11
Reconfigure & Upgrade Font Awesome

urbanwide:
* Removes the import lines from stylesheets.scss & adds a single import line to application.js
* Upgrades the fontawesome node package so we can use FA 6 rather than 5. This gives us 400+ more available icons. See: https://fontawesome.com/v6/search?m=free

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -13,6 +13,7 @@ require("trix")
 require("@rails/actiontext")
 require('awesomplete')
 
+import "@fortawesome/fontawesome-free/css/all.css";
 import Trix from 'trix'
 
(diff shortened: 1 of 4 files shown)


ldodds:
Not sure we need this now, if we've got the other workaround?

urbanwide:
Yep fine, I’ll remove it. Can I still upgrade font awesome at some point too though? More icons are always useful!

urbanwide:
No longer required. Fixed by: https://github.com/Energy-Sparks/energy-sparks/pull/2348
However, see the following new PR that upgrades FA version from 5 to 6: https://github.com/Energy-Sparks/energy-sparks/pull/2347

PR #2349, opened 2023-01-11
Downgrade sass-rails and sprockets

ldodds:
Downgrade sass-rails and sprockets gems to avoid segfaults.

See:

https://github.com/sass/sassc-ruby/issues/207#issuecomment-1058148376


urbanwide (approved):
Looks good @ldodds 👍

PR #2350, opened 2023-01-11
Basic page nav component

urbanwide:
This PR adds the page nav component, which is currently mostly based on Leigh's prototype.
Probably would be a good idea to get this merged in as is, if approved. The following still needs doing, which can be done in a separate PR:

- Add the ability to highlight current page
- Styling
- Spec
- Moving css to own component scoped file
- Think about scoping each component to it's own directory (as the components directory will get messy very quickly)
- For the advice pages nav, links are grouped by fuel type. In the nav partial, we need to add the ability to turn sections off if current school doesn't have given fuel type


app/components/page_nav_component.html.erb
```diff
@@ -0,0 +1,40 @@
+
+<style>
```
ldodds, line 2: Can we move this to a separate scss file. I only inlined some of the CSS in the prototype to minimise changes.

Also wondering whether "analysis-page-nav" might be better scoping for the class name?

app/views/schools/advice/_nav.html.erb
```diff
@@ -1,8 +1,16 @@
-<h2>Pages</h2>
-<% if advice_pages.any? %>
-  <% advice_pages.each do |advice_page| %>
-    <li><%= link_to advice_page.key, school_advice_tab_path(school, advice_page.key) %></li>
+<%= render(PageNavComponent.new(name:'Menu', icon: 'bars', bgcolor: '#232b49')) do |c| %>
+  <% c.with_section do |s| %>
+    <% s.with_item(name: 'Total energy use', href: '#') %>
+  <% end %>
+  <% c.with_section(name: "Electicity", icon: 'bolt', bgcolor: '#93e1f6') do |s| %>
```
ldodds, line 5: typo "Electicity"

app/components/page_nav_component.html.erb
urbanwide: Yes - that's on my todo list at the top :-)

I thought the nav could be reusable for other sections of the site - hence the broader scoping?

app/views/schools/advice/_nav.html.erb
urbanwide: Thanks!

app/components/page_nav_component.html.erb
urbanwide: Moved the css out for now but hope to move it later into a component css file.

ldodds (approved):

PR #2351, opened 2023-01-11
Initial outline of analysis tab with content

jhigman:
This PR starts building the analysis tab of the baseload advice page. (The content is not yet complete - other PRs to follow).

The controllers for individual advice pages are subclasses of the AdviceController, which allows them to have specific methods for the insights, analysis and learn more sections (without needing dynamic data from the advice pages themselves).

The learn more tab will usually just expose the content from the model, so it is left as a method on the parent advice controller.

Each advice controller is expected to have a template for the insights and analysis tabs. They will need to share the generic layout partial (which wraps the individual content). Additional controllers for other advice pages will need to be added to the routes.rb.

The analysis page itself for baseload uses the Baseload::BaseloadCalculationService and Baseload::BaseloadBenchmarkingService services to retrieve school-specific data.


config/routes.rb
```diff
@@ -138,11 +138,18 @@
         get "/#{tab}", to: redirect("/schools/%{school_id}/analysis")
       end
 
-      get 'advice', to: 'advice#index'
-      get 'advice/:key(/:tab)', to: 'advice#show', as: :advice_tab
+      resource :advice, controller: 'advice', only: [:show] do
+        resource :baseload, controller: 'advice/baseload', only: [:show] do
```
urbanwide, line 142: Just a note for future implementation - if this is a re-useable pattern, the member section *I think* could be defined as a concern.

app/helpers/advice_page_helper.rb
```diff
@@ -0,0 +1,9 @@
+module AdvicePageHelper
+  def advice_page_path(school, advice_page, tab = :insights)
+    polymorphic_path([tab, school, :advice, advice_page.key.to_sym])
+  end
+
+  def advice_baseload_high?(val)
```
urbanwide, line 6: This feels like it should not be a helper but be part of the benchmark service. I have no concept of how tricky this would be though so please ignore me as required :-)

urbanwide (approved):
Looks good @jhigman - just a couple of comments / queries 👍

app/helpers/advice_page_helper.rb
jhigman: Yes good point - will look into whether this can be part of the service

config/routes.rb
jhigman: Yes - will try a concern or similar when we introduce more controllers (left it pretty basic in this initial version)

PR #2354, opened 2023-01-12
Allow advice pages to have associated activities

ldodds:
Updates the AdvicePage model to have an associated list of activity types.

Provides a simple admin interface for managing the activities, with a priority number. The interface is basic, but similar to what is used to manage similar relationships across the site.

This will be used to recommend activities from the advice page Insights tab.


iant (approved):
looks good to me @ldodds

PR #2357, opened 2023-01-13
2761 further baseload advice page analysis

jhigman:
This PR expands on the analysis page template:

- clickable table of contents at the top
- scroll top adjustment for the section titles in the advice pages
- month/year text for the chart description
- test that the analysis tab renders
- move the page title to sit with the content tabs
- add back-to-top links for each section

es-advice

### DIFF
--- a/app/views/schools/advice/_advice_page.html.erb
+++ b/app/views/schools/advice/_advice_page.html.erb
@@ -1,22 +1,20 @@
-<div class="advice-page">
-  <h1><%= "Advice page: #{advice_page.key.humanize}" %></h1>
+<h1><%= "Advice page: #{advice_page.key.humanize}" %></h1>
 
-  <div class="advice-page-breadcrumb">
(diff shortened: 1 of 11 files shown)


app/views/schools/advice/_advice_page.html.erb
```diff
@@ -14,6 +12,7 @@
     <%= render 'schools/advice/nav', school: school, advice_pages: advice_pages %>
   </div>
   <div class="col-md-10 advice-page-tabs">
+    <h1><%= "#{advice_page.key.humanize} analysis and advice" %></h1>
```
iant, line 15: does this string need extracting out so it can be translated?

iant (approved):
looks good to me @jhigman 1 comment where i think an extracted string may have been missed

app/views/schools/advice/_advice_page.html.erb
jhigman: Good point! - I'll make it translatable

PR #2358, opened 2023-01-13
Chart controls, placeholder table, long term content

jhigman:
This PR adds more of the text content to each of the sections of the analysis tab of the baseload advice page.

The tables are just placeholders (static html) at the moment.

Still to do:

- proper table contents (annual average baseload and meter breakdown,  seasonal variation, intraweek variation)
- show/hide text and meter breakdown section depending on whether multiple meters are present
- chart and select box for meter breakdown of baseload
- check the date ranges for charts ("This chart shows the electricity baseload for your school between Month Year..") as they may not be the full years currently shown (will depend on available data?)
- confirm definition of high/low usage (currently high is just estimated savings> 0.0)
- text at end of recent trend section for increase/decrease may need adding? unclear from doc


ldodds (approved):
Looks OK if you want to merge, then work on removing the placeholders

PR #2359, opened 2023-01-16
Update advice page navbar component

urbanwide:
After lots of fudging round, here is v2 of the page nav component with a caveat that the css will need further tweaking (mainly because I would like the nav bar to appear from the left and not above when the menu button is clicked (it was working this way before but broke when further page content was added)).


app/views/schools/advice/show.html.erb
```diff
@@ -1,18 +1,18 @@
...
+
 <div class="row">
-  <div class="col-md-2 advice-page-nav">
+  <div class="col-md-3 col-lg-3 col-xl-2 advice-page-nav">
     <%= render 'nav', school: @school, advice_pages: @advice_pages %>
   </div>
-  <div class="col-md-10">
-    <h2>All pages</h2>
+  <div class="col-md-9 col-lg-9 col-xl-10 advice-page-tabs">
+    <h2><%= render(PageNavComponent::CollapseButton.new(display: 'inline')) %> All pages</h2>
```
iant, line 15: does "All pages" & "Advice" (above) need extracting to a translation yml?

app/views/schools/advice/_nav.html.erb
```diff
@@ -1,16 +1,16 @@
-<%= render(PageNavComponent.new(name:'Menu', icon: 'bars', bgcolor: '#232b49')) do |c| %>
+<%= render(PageNavComponent.new(name:'Advice', icon: 'home', href: school_advice_path(@school), options: { match_controller: true })) do |c| %>
```
iant, line 1: does 'Advice' (and 'Electricity' etc)  need extracting?

iant (approved):
looks good to me @urbanwide - couple of comments on text that maybe needs extracting?

PR #2361, opened 2023-01-17
Fix synchronisation of sms and email fields

ldodds:
Fixes the transifex synchronisation of content for email and sms alerts:

* add all the variables to `tx_valid_attribute` methods, checking if email or sms is active
* configure `email_title` and `sms_content` as templated values, so we do variable escaping correctly
* add tests to ensure the fields are correctly serialised when the sms and email content is marked as active
* revise existing spec to check that sms and email active content ratings are serialised


iant (approved):
:+1:

PR #2363, opened 2023-01-18
Use baseload data for tables

jhigman:


ldodds (approved):
Happy for this to be merged as work in progress. We can review further when the implementation is largely complete. There are some opportunities to refactor the controller a bit I think.

PR #2366, opened 2023-01-19
2761 baseload advice additional columns

jhigman:
This PR adds more data to the baseload tables.

I’ve put methods from the controller in a concern - not my favourite arrangement, but in line with what we’ve got in other controllers - and allows the methods to be tested in isolation (and reused later)
we can change that if they fit better elsewhere though (e.g. in a service)

couple of things that need addressing:
- I’ve only shown the “Change from previous year” as a kw figure so far (not as a %) in the meter breakdown
- the table titles and labels (like “Totals”, “Limited variation”..) aren’t translated
- probably need to check the numbers being shown against what we expect

The main element not present yet is the per-meter charts with dropdown in the meter breakdown


ldodds:
@jhigman am going to merge this, added a couple of the todo list items to the trello card.

ldodds (approved):

PR #2368, opened 2023-01-23
Select box submits remote form to select meter chart

jhigman:


ldodds:
Where we're doing this currently, e.g. on the pupil charts that allow both meter and date switching, we're using custom Javascript to trigger updates to Highcharts.

We didn't hit any cross-site issues here, so problems with embedding tokens and CORS look like they're specific to this UJS approach.

It looks like it triggers loading a section into the page, which means re-initialising Highcharts. I think you're likely to hit problems with the chart controls being initialised/working correctly. At the moment those are done on page load. For the pupil charts they relevant functions are called in the custom Javascript.

Given we can already add an mpan_mprn to a chart config, and some existing code for that, I'm wondering whether it'd be better to do this using the existing approach? Are there extra benefits to the approach taken here?

The only issue I can see is getting the start/end dates for the meter. But we could add those in as a further iteration. Either by updating the JSON the chart controller returns. Or by simply embedding the relevant dates into the page, so any custom Javascript can switch as required.

PR #2370, opened 2023-01-23
Create routes, default templates and basic specs for all advice pages

ldodds:
* adds routes for all the advice pages
* adds a basic controller for each page
* adds default templates for each insights and analysis tab
* adds keys for page title, plus insights and analysis tab titles
* adds all pages to the navbar
* add default specs for all pages, which just confirms they load OK with basic navigation
* some tidying of the baseload spec, separate out tests for navigation and restricted view into separate spec


app/controllers/schools/advice/boiler_control_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Schools
+  module Advice
+    class BoilerControlController < AdviceController
+      include AdvicePages
+
+      def show
```
urbanwide, line 6: Am wondering if this redirect ought to be done once in the superclass rather than each controller - assuming it's a standard pattern? You have the advice_page_key which could be used in the redirect?

spec/system/schools/advice_pages/boiler_control_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  context 'as school admin' do
+
+    let(:user)  { create(:school_admin, school: school) }
+
+    before do
+      sign_in(user)
+      visit school_advice_boiler_control_path(school)
+    end
+
+    it 'shows the nav bar' do
```
urbanwide, line 22: I wonder if these (repeated) tests could be shared examples?

app/views/schools/advice/boiler_control/analysis.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= render 'schools/advice/advice_page', school: @school, advice_pages: @advice_pages, advice_page: @advice_page, tab: @tab do %>
+  <h2><%= t("advice_pages.#{@advice_page.key}.analysis_title") %></h2>
+<% end %>
```
urbanwide, line 3: I was reading here: https://guides.rubyonrails.org/v6.1/layouts_and_rendering.html#template-inheritance that "Similar to the Layout Inheritance logic, if a template or partial is not found in the conventional path, the controller will look for a template or partial to render in its inheritance chain". So we could potentially have an analysis.html.erb in /advice with the above in, that rendered a partial called _analysis.html.erb for the page content, which would be found in the page (boiler_control in this case) sub-directory. Would need to try it though!

urbanwide (approved):
👍

PR #2372, opened 2023-01-23
Recommendations view component

urbanwide:
This PR is for the recommendations view component, but also does a few other things that were required, so is quite a big PR unfortunately.

* The main change is of the file structure for advice pages to help prevent code repetition that was starting to occur for insights and analysis pages. Each page directory now has  _insights.html.erb and _analysis.html.erb files which are pulled in by the advice_base/insights.html.erb & advice_base/analysis.html.erb (and by use of template 'inheritance').
(mentioned here: https://github.com/Energy-Sparks/energy-sparks/pull/2370#discussion_r1084258941). I did look at using nested layouts here instead, but they would be kept out of the directory structure of the advice pages (in the layouts) directory, which didn't feel right for very specific insights and advice tab pages.
* also introduces a new method to translatable attachments and spec (which encapsulates a code pattern that is used in lots of other places)
* also changes the insights and analysis title key structure of all pages to match that of baseload
* also puts section titles in the baseload insights page to match the analysis page.


app/views/schools/advice/advice_base/insights.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= render 'schools/advice/advice_page', school: @school, advice_pages: @advice_pages, advice_page: @advice_page, tab: @tab do %>
+
+  <h2><%= t("advice_pages.#{@advice_page.key}.insights.title") %></h2>
+
+  <%= render 'insights' %>
+
+  <%= render 'schools/advice/section_title', section_id: 'recommendations', section_title: t('advice_pages.insights.recommendations.title') %>
+  <p><%= t("advice_pages.#{@advice_page.key}.insights.next_steps") %></p>
+
+  <%= render(RecommendationsComponent.new(recommendations: @activity_types, title: 'Work with the pupils', classes: 'pb-4')) %>
```
iant, line 10: do "Work with the pupils" & "Take action.." strings need to be extracted to translation files?

app/models/concerns/translatable_attachment.rb
```diff
@@ -20,4 +20,12 @@ def t_attached(name, locale = I18n.default_locale)
     end
     send("#{name}_#{locale}")
   end
+
+  def t_attached_or_default(name, locale = I18n.locale)
+    if locale != I18n.default_locale && t_attached(name, locale).present?
+      t_attached(name, locale)
+    elsif t_attached(name, I18n.default_locale).present?
+      t_attached(name, I18n.default_locale)
+    end
```
iant, line 29: I'm likely reading what this method is doing totally wrong so please do ignore this if i am but i'm wondering if this could be simplified (I'm guessing `t_attached` returns nil if not found?)
```
  def t_attached_or_default(name, locale = I18n.locale)
    t_attached(name, locale) || t_attached(name, I18n.default_locale) || nil
  end
```
?

iant (approved):
this looks good @urbanwide couple of v-minor comments

PR #2373, opened 2023-01-24
Refactor concern into standalone service

ldodds:
Initial version of the baseload controller used the `AdvicePages` concern. This PR extract the code from that concern into a separate service class. This avoids that concern getting very large, tidies the interface and allows us to share common code across different services.

I'm not sure whether "Service" is the right name here as it doesn't implement a simple service interface. But it's not quite a set of helper methods either.

* add service
* add specs for the service
* revise system specs to stub the service
* remove the concern

Adds some additional helper methods to the service to extract more code from the controller.

Still to do:

* [x] implement remaining tests
* [x] alter code that provides meter breakdown to inject the `Meter`
* [x] change views to use the above Meter rather than dynamically loading


ldodds:
Had to work around a quirk in the analytics here.

I was calling `.aggregate_meter?` which was fine in the tests as it was stubbed out. In the application it failed with an exception:

```
Unexpected inconsistency in aggregate meter logic see aggregate_meter2?
```

The `aggregate_meter?` method is defined twice in the Meter code, the first version returns false, the second does some additional logic. Will investigate.

ldodds:
Note: this looks like a big PR, but I had to update all the stubbed controllers to remove the `AdvicePages` concern.

app/services/schools/advice/base_service.rb
```diff
@@ -0,0 +1,26 @@
+module Schools
+  module Advice
+    class BaseService
+      def initialize(school, meter_collection)
+        @school = school
+        @meter_collection = meter_collection
+      end
+
+      def date_range(meter: aggregate_meter)
+        [meter.amr_data.start_date, meter.amr_data.end_date]
```
iant, line 10: Just a thought but maybe this could return an struct or openstruct here too? (so we can use something like`date_range.end_date` instead `date_range[1]` or assigning to separate `start_date` and `end_date` variables)

app/services/schools/advice/base_service.rb
```diff
@@ -0,0 +1,26 @@
...
+      end
+
+      # from analytics: lib/dashboard/charting_and_reports/content_base.rb
+      def calculate_rating_from_range(good_value, bad_value, actual_value)
+        actual_value = actual_value.abs
+        [10.0 * [(actual_value - bad_value) / (good_value - bad_value), 0.0].max, 10.0].min.round(1)
```
iant, line 16: I realise this is bought over from the analytics but I'm thinking this method would benefit from a comment on what it's returning and why (and the use the abs of `actual_value`).  or maybe it could go in as a `describe` block in the spec.

app/services/schools/advice/baseload_service.rb
```diff
@@ -0,0 +1,191 @@
+# rubocop:disable Naming/AsciiIdentifiers
+module Schools
+  module Advice
+    class BaseloadService < BaseService
+      def has_electricity?
+        @school.has_electricity?
+      end
+
+      def multiple_meters?
```
iant, line 9: `multiple_electricity_meters?`?

app/services/schools/advice/baseload_service.rb
```diff
@@ -0,0 +1,191 @@
...
+            variation_by_meter[meter.mpan_mprn] = calculate_intraweek_variation(meter, meter.amr_data.end_date, true)
+          end
+        end
+        variation_by_meter
+      end
+
+      private
+
+      def asof_date
+        @asof_date ||= date_range[1]
```
iant, line 111: I've left a comment above on a possible use of openstruct for date_range (e.g. `date_range.end_date`)

app/services/schools/advice/baseload_service.rb
```diff
@@ -0,0 +1,191 @@
...
+      def multiple_meters?
+        @school.meters.electricity.count > 1
+      end
+
+      def average_baseload_kw(period: :year)
+        @average_baseload_kw ||= baseload_service.average_baseload_kw(period: period)
+      end
+
+      #This could be pushed down into the underlying service
+      #TODO cache results
```
iant, line 18: Do these comments need removing?

app/services/schools/advice/baseload_service.rb
```diff
@@ -0,0 +1,191 @@
...
+
+      def seasonal_variation_by_meter
+        variation_by_meter = {}
+        if @meter_collection.electricity_meters.count > 1
+          @meter_collection.electricity_meters.each do |meter|
+            variation_by_meter[meter.mpan_mprn] = calculate_seasonal_variation(meter, meter.amr_data.end_date, true)
+          end
+        end
+        variation_by_meter
+      end
```
iant, line 92: maybe have an early return here and in the `intraweek_variation_by_meter` method below? Something like

```
def electricity_meters
  @electricity_meters ||= @meter_collection.electricity_meters
end

def seasonal_variation_by_meter
  return {} unless electricity_meters.count > 1

  electricity_meters.each_with_object({}) do |meter, variation_by_meter|
    variation_by_meter[meter.mpan_mprn] = calculate_seasonal_variation(meter, meter.amr_data.end_date, true)
  end
end
```

iant (approved):
this looks good to me @ldodds couple of comments on possible changes

app/services/schools/advice/base_service.rb
ldodds: I'll add a comment. I'm also thinking that the ratings may need to be pushed down into the analytics services.

app/services/schools/advice/base_service.rb
ldodds: I've just removed the method for now. It wasn't doing much work and its not really adding any clarity to the code.

So I've changed the controller to access the date range it needs directly, and the asof_date method to do the same.

PR #2374, opened 2023-01-25
Added erb file extension for calendar legend so that directives are p…

jhigman:
…rocessed


jhigman:
Before fix:

es-cal


After fix:

es-cal-fixed

iant (approved):

PR #2376, opened 2023-01-25
2761 baseload meter charts

jhigman:


app/assets/javascripts/usage_charts.js
```diff
@@ -13,26 +13,21 @@ $(document).ready(function() {
...
-      }
-    } else {
-      chartConfig.mpan_mprn = null;
-      chartConfig.series_breakdown = 'meter';
+    var series_breakdown = $("#series_breakdown").val();
+    if (series_breakdown) {
+      chartConfig.series_breakdown = series_breakdown;
     }
 
-//    chartConfig.y_axis_units = measurement;
```
ldodds, line 31: Just wanted to check that this isn't needed any more? Looks like there's no axis switching on these charts so possibly not.

ldodds (approved):

app/assets/javascripts/usage_charts.js
jhigman: Yes, as far as I can tell this has been commented out since Feb 2022, and I can't see any field in the usage pages with id=measurement (which is where the value would come from ), so I'm assuming not used any more

PR #2380, opened 2023-01-26
Make baseload advice tolerant of limited data

ldodds:
The main baseload page will run if a school has < 12 months of data, but the seasonal and intraday variation won't.

We also shouldn't show the yearly comparison of baseload, or last weeks baseload if there isn't enough recent data

PR:

* Updates control to check to see if there is one years data, then skip some analysis if less than a year
* Changes chart and page contents to use "X months" rather than "12 months"
* Ensure that seasonal and intraday variation sections are skipped if there isn't one years data
* Don't show % change from last year if there's less than a years data
* Don't show last week baseload if there's no recent data

### DIFF
--- a/app/helpers/advice_page_helper.rb
+++ b/app/helpers/advice_page_helper.rb
@@ -56,4 +56,21 @@ module AdvicePageHelper
     return 0.0 if base == 0.0
     (current - base) / base
   end
+
+  def recent_data?(end_date)
+    end_date > (Time.zone.today - 30)
+  end
+
+  def one_years_data?(start_date, end_date)
+    return (end_date - 364) >= start_date
+  end
+
+  def months_analysed(start_date, end_date)
+    months = months_between(start_date, end_date)
+    months > 12 ? 12 : months
+  end
+
+  def months_between(start_date, end_date)
+    ((end_date - start_date).to_f / 365 * 12).round
+  end
 end
--- a/app/helpers/advice_page_helper.rb
+++ b/app/helpers/advice_page_helper.rb
@@ -22,4 +22,38 @@ module AdvicePageHelper
   def format_rating(rating)
     rating > 4 ? "Limited variation" : "Large variation"
   end
+
+  #link to a specific benchmark for a school group, falls back to the
+  #generic benchmark page if a school doesn't have a group
+  def benchmark_for_school_group_path(benchmark_type, school)
+    if school.school_group.present?
+      benchmark_path({ "benchmark_type" => benchmark_type, "benchmark[school_group_ids][]" => school.school_group.id })
+    else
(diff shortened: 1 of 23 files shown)


app/views/schools/advice/baseload/_current_baseload.html.erb
```diff
@@ -9,21 +9,32 @@
...
-          <%= up_downify( format_unit(percentage_change_week, :relative_percent)) %>
-        </td>
+        <% if analysis_dates.recent_data %>
+          <td class="text-right"><%= format_unit(current_baseload.average_baseload_kw_last_week, :kw) %></td>
+          <td class="text-right">
+            <%= up_downify( format_unit(current_baseload.percentage_change_week, :relative_percent)) %>
+          </td>
+        <% else %>
+          <td class="text-right old-data">
+            no recent data
```
iant, line 19: does this need extracting out to a translation file?

app/helpers/advice_page_helper.rb
```diff
@@ -50,10 +50,31 @@ def categorise_school_vs_benchmark(school, benchmark_school, exemplar_school)
...
     return 0.0 if base == 0.0
     (current - base) / base
   end
+
+  def recent_data?(end_date)
+    end_date > (Time.zone.today - 30)
+  end
+
+  def one_years_data?(start_date, end_date)
+    return (end_date - 364) >= start_date
```
iant, line 69: return not needed here

config/locales/views/advice_pages/baseload.yml
```diff
@@ -13,7 +13,7 @@ en:
       analysis_baseload_low_details_html: Your school's baseload is low, which is good and as a result you are saving %{estimated_savings} per year versus the average of other schools of a similar size.
       analysis_electricity_baseload_chart_explanation: Each point on this chart represents the average over the hours a school is closed on that day. You can click on a point in the chart and it will drill down to show you the usage on that day. Ideally, there should be little variation between different days and seasons.
       analysis_electricity_baseload_chart_subtitle: This chart shows the electricity baseload for your school between %{start_month_year} and %{end_month_year}.
-      analysis_electricity_baseload_chart_title: Electricity baseload for the last 12 months
+      analysis_electricity_baseload_chart_title: Electricity baseload for the last %{months_analysed} months
```
iant, line 16: does this need a one and other key? (e.g. for 1 "Electricity baseload for the last month") (similarly the 2 translations below)

iant (approved):
looks good @ldodds couple of minor comment/questions

config/locales/views/advice_pages/baseload.yml
ldodds: I'll handle this in the next PR

PR #2382, opened 2023-01-26
Test link from admin pages and sort index by key

jhigman:
There was already a link from the admin pages to the advice page management, so this PR just adds a test (and also ensures that the index page list is sorted).


iant (approved):
:+1: looks good too @jhigman

PR #2384, opened 2023-01-26
2863 error handling for advice page controllers

jhigman:
When an error occurs in the advice pages, we rescue the exception in the advice pages base class, log it to Rollbar, and show a styled error page (which includes the usual advice page navigation).

### DIFF
--- a/app/controllers/schools/advice/advice_base_controller.rb
+++ b/app/controllers/schools/advice/advice_base_controller.rb
@@ -12,6 +12,11 @@ module Schools
 
       before_action :check_aggregated_school_in_cache, only: [:insights, :analysis]
 
+      rescue_from StandardError do |exception|
+        Rollbar.error(exception)
+        render 'error'
+      end
+
       def show
         redirect_to url_for([:insights, @school, :advice, advice_page_key])
       end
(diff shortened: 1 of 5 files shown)


config/locales/views/advice_pages/advice_pages.yml
```diff
@@ -7,6 +7,9 @@ en:
       other: Improving
     breadcrumbs:
       root: Advice
+    error:
+      message: We encountered an error attempting to generate your analysis. The Energy Sparks team have been notified of the problem.
+      title: Error
```
ldodds, line 12: Maybe "Sorry, something has gone wrong"

app/controllers/schools/advice/advice_base_controller.rb
```diff
@@ -12,6 +12,11 @@ class AdviceBaseController < ApplicationController
 
       before_action :check_aggregated_school_in_cache, only: [:insights, :analysis]
 
+      rescue_from StandardError do |exception|
+        Rollbar.error(exception)
```
ldodds, line 16: Are we able to add some context to this, just wondering if we can add in e.g:

Rollbar.error(error, advice_page: @advice_page.key, school: @school.name, school_id: @school.id) assuming that's available on the controller.

Just helps a bit with debugging if we can

ldodds (approved):

PR #2386, opened 2023-01-27
Show warning if data is old

jhigman:
A data warning is shown on the analysis and insights tabs of the advice page if the data is not recent (end date is more than 30 days ago).

The warning message needs the fuel type, so the advice pages now have a fuel type (implemented as an enum, similar to e.g. AlertTypes)

The flag for data warning is set in the base controller, and passed to the analysis and insights templates.


iant (approved):
👍 looks good to me @jhigman

PR #2389, opened 2023-01-27
Created chart component and replaced charts on baseload analysis

jhigman:
A chart component allows charts to be added to a page, specifying a chart type, school, and chart config.

Some other parameters (analysis controls, zoom, html class) can also be provided.

The title, subtitle, header and footer text can be optionally added - implemented as slots.

The title tag has an id to enable linking to that specific chart.

### DIFF
--- a/app/components/chart_component.html.erb
+++ b/app/components/chart_component.html.erb
@@ -3,5 +3,5 @@
 
 <div id="chart_wrapper_<%= chart_type %>" class="chart-wrapper">
   <%= render 'shared/analysis_controls', chart_type: chart_type, axis_controls: true, analysis_controls: analysis_controls %>
-  <%= chart_tag(school, chart_type, no_zoom: no_zoom, wrap: false, chart_config: chart_config) %>
+  <%= chart_tag(school, chart_type, no_zoom: no_zoom, wrap: false, html_class: html_class, chart_config: chart_config) %>
 </div>
--- a/app/components/chart_component.html.erb
+++ b/app/components/chart_component.html.erb
@@ -6,15 +6,11 @@
   <h5 class="chart-subtitle"><%= subtitle %></h5>
 <% end %>
 
-<% if header %>
-  <p><%= header %></p>
-<% end %>
+<%= header %>
 
 <div id="chart_wrapper_<%= chart_type %>" class="chart-wrapper">
   <%= render 'shared/analysis_controls', chart_type: chart_type, axis_controls: true, analysis_controls: analysis_controls %>
   <%= chart_tag(school, chart_type, no_zoom: no_zoom, wrap: false, html_class: html_class, chart_config: chart_config) %>
 </div>
 
-<% if footer %>
-  <p><%= footer %></p>
-<% end %>
+<%= footer %>
--- a/app/components/chart_component.html.erb
+++ b/app/components/chart_component.html.erb
@@ -1,16 +1,28 @@
-<% if title %>
-  <h4 id="chart-section-<%= chart_type %>" class="chart-title"><%= title %></h4>
+<% if valid_config? %>
+
+  <% if title %>
+    <h4 id="chart-section-<%= chart_type %>" class="chart-title"><%= title %></h4>
+  <% end %>
+
+  <% if subtitle %>
+    <h5 class="chart-subtitle"><%= subtitle %></h5>
+  <% end %>
+
+  <%= header %>
+
+  <div id="chart_wrapper_<%= chart_type %>" class="chart-wrapper">
+    <%= render 'shared/analysis_controls', chart_type: chart_type, axis_controls: true, analysis_controls: analysis_controls %>
+    <%= chart_tag(school, chart_type, no_zoom: no_zoom, wrap: false, html_class: html_class, chart_config: chart_config) %>
+  </div>
+
+  <%= footer %>
+
+<% else %>
+
+  <div class="padded-row">
+    <div class="bg-negative-light">
+      <p>The chart can't be displayed - please check the configuration</p>
+    </div>
+  </div>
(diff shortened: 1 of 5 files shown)


app/components/chart_component.rb
```diff
@@ -0,0 +1,21 @@
...
+  renders_one :title
+  renders_one :subtitle
+  renders_one :header
+  renders_one :footer
+
+  attr_reader :school, :chart_type, :chart_config, :analysis_controls, :no_zoom, :html_class
+
+  include ChartHelper
+
+  def initialize(chart_type: '', school: nil, chart_config: {}, analysis_controls: false, no_zoom: true, html_class: 'analysis-chart')
```
ldodds, line 13: It looks like we'll end up with lots of repeated calls to `create_chart_config(@school, :baseload_lastyear)` across charts. And both of the school and the chart type are already passed as parameters. Could we just let the component call `create_chart_config` to avoid repeated code. As I think in most cases that's all we'll want.

We could allow `chart_config` to be provided as an optional parameter, if we needed something particularly custom?

app/components/chart_component.html.erb
```diff
@@ -0,0 +1,20 @@
+<% if title %>
+  <h4 class="chart-title"><%= title %></h4>
+<% end %>
+
+<% if subtitle %>
+  <h5 class="chart-subtitle"><%= subtitle %></h5>
+<% end %>
+
+<% if header %>
+  <p><%= header %></p>
```
ldodds, line 10: Could the header/footer just render whatever partial its been provided rather than assuming the content is wrapped in a paragraph? E.g. we might want a couple of paras, or some bullet points in a footer.

app/components/chart_component.rb
ldodds: Can you also check the defaults here, as I think in most cases we have both analysis controls and allow zoom? Can probably check across what we're doing in a few of the partials across the app currently?

ldodds (changes requested):

app/views/schools/advice/baseload/_analysis.html.erb
```diff
@@ -26,15 +26,13 @@
...
-  <h5 class="chart-subtitle"><%= t('advice_pages.baseload.analysis.charts.baseload_chart_subtitle', start_month_year: month_year(@analysis_dates.one_years_data ? @analysis_dates.end_date - 1.year : @analysis_dates.start_date), end_month_year: month_year(@analysis_dates.end_date)) %></h5>
-
-  <div id="chart_wrapper_baseload_lastyear" class="chart-wrapper">
-    <%= render 'shared/analysis_controls', chart_type: :baseload_lastyear, axis_controls: true, analysis_controls: true %>
-    <%= chart_tag(@school, :baseload_lastyear, no_zoom: true, wrap: false, chart_config: create_chart_config(@school, :baseload_lastyear, nil)) %>
-  </div>
-
-  <p><%= t('advice_pages.baseload.analysis.charts.baseload_chart_explanation') %></p>
+  <%= render(ChartComponent.new(chart_type: :baseload_lastyear, school: @school)) do |c| %>
+    <% c.with_title { t('advice_pages.baseload.analysis.charts.baseload_chart_title', months_analysed: @analysis_dates.months_analysed) } %>
```
urbanwide, line 30: I was going to suggest here to use component slots for title and subtitle (rather than passthrough slots) so that you could pass parameters but have realised that you probably are using blocks in case you want to pass a chunk of html through to the title / subtitle? Otherwise you would have passed them through as parameters to the Chart component itself.

spec/components/chart_component_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        c.with_subtitle { "I'm a subtitle" }
+        c.with_header   { "<strong>I'm a header</strong>".html_safe }
+        c.with_footer   { "<small>I'm a footer</small>".html_safe }
+      end
+    end
+    it { expect(html).to have_selector("h4", text: "I'm a title") }
+    it { expect(html).to have_selector("h4", id: "chart_baseload_title") }
+    it { expect(html).to have_selector("h5", text: "I'm a subtitle") }
+    it { expect(html).to have_selector("strong", text: "I'm a header") }
+    it { expect(html).to have_selector("small", text: "I'm a footer") }
```
urbanwide, line 34: Do you need to test what happens if you don't pass the optional parameters?

urbanwide (commented):
Looks good @jhigman - just a question about the rspec 👍

ldodds (approved):

spec/components/chart_component_spec.rb
jhigman: I've added a test that the h4 and h5 tags aren't rendered if the title and subtitle aren't set. More difficult to test for a missing header or footer, as there's nothing rendered when they're empty.

Do you think that covers it sufficiently?

urbanwide: Hadn't thought of those - was more thinking of what happens if school and chart_config aren't passed through for example? They both have defaults of nil so I assume they are optional?

app/components/chart_component.rb
```diff
@@ -0,0 +1,31 @@
...
+    @html_class = html_class
+  end
+
+  def chart_config
+    @chart_config ||= create_chart_config(@school, @chart_type)
+  rescue StandardError
+    nil
+  end
+
+  def valid_config?
```
urbanwide, line 28: Ok that works nicely :-) Assume if school is nil that everything behaves?

urbanwide: If chart_config and school aren't optional parameters, you should pass them as follows (without the nil): 

`def` initialize(chart_type: '', school:, chart_config:, analysis_controls: true etc

PR #2390, opened 2023-01-30
Allow relative links

ldodds:
Change the validation for links to properly support relative paths.

Required to support links to other analysis pages and page sections from the advice page editor.


urbanwide (approved):
Looks good @ldodds 👍

PR #2393, opened 2023-02-02
Create notice view component

urbanwide:
This PR implements the Notice Component and moves existing notices over to use the component.
As part of this, I've created a new components helper, so you are able to write the following (which I think reads better):

`<%= component 'notice', status: :negative do |c| %>
`
Rather than:
`<%= render(NoticeComponent.new(status: :negative)) do |c| %>`

Happy to move other components over to using it as part of this PR if all ok.

### DIFF
--- a/app/components/notice_component.html.erb
+++ b/app/components/notice_component.html.erb
@@ -1,0 +1,6 @@
+<div class="p-4 advice-page-notice neutral">
+  <%= content %>
+  <div class="text-right">
+    <%= link_to link_name, href %>
+  </div>
+</div>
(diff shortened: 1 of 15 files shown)


app/components/notice_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="p-4 page-notice<%= classes %>">
```
ldodds, line 1: nice addition to be able to pass in other classes

ldodds (approved):

app/components/notice_component.scss
```diff
@@ -0,0 +1,12 @@
+.page-notice {
```
ldodds, line 1: Actually, I'm now wondering whether we should scope the css at all here, e.g. .component-page-notice, or .notice-component?

ldodds (approved):

PR #2395, opened 2023-02-03
Add text and charts to electricity recent changes analysis

jhigman:


db/schema.rb
```diff
@@ -827,21 +827,21 @@
     t.index ["replaced_by_id"], name: "index_global_meter_attributes_on_replaced_by_id"
   end
 
-  create_table "good_job_processes", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+  create_table "good_job_processes", id: :uuid, default: -> { "public.gen_random_uuid()" }, force: :cascade do |t|
```
ldodds, line 830: This doesn't seem to be part of this change? I noticed this had popped up after running latest migrations locally

app/views/schools/advice/electricity_recent_changes/_analysis.html.erb
```diff
@@ -0,0 +1,49 @@
...
+    no_zoom: true,
+    chart_config: @chart_config,
+    html_class: 'usage-chart')) do |c| %>
+    <% c.with_title { t('advice_pages.electricity_recent_changes.analysis.compare_recent_weeks.chart_title') } %>
+    <% c.with_subtitle { t('advice_pages.electricity_recent_changes.analysis.compare_recent_weeks.chart_subtitle') } %>
+    <% c.with_header do %>
+      <p><%= t('advice_pages.electricity_recent_changes.analysis.compare_recent_weeks.chart_description') %></p>
+    <% end %>
+  <% end %>
+  <%= render 'controls', chart_config: @chart_config, period: :weekly, supply: :electricity, split_meters: @split_meters, meters: @meters %>
```
ldodds, line 27: Would this work as the `footer` for the chart component?

ldodds (approved):

app/views/schools/advice/electricity_recent_changes/_analysis.html.erb
jhigman: Aha yes I'd already made that change but hadn't pushed it..

db/schema.rb
jhigman: Not sure why that changed - some version thing perhaps - I've reverted it to be on the safe side..

ldodds (approved):

PR #2396, opened 2023-02-03
Update js to handle fields within chart section

jhigman:
This is a PR to update the javascript that displays usage charts so that multiple charts can be displayed on the same page.

Previously, the usage js functions relied on element ids (#period, #supply etc) to fetch arguments from the page markup and render charts. The PR changes to using named elements within a specified section of the page so that the chart, config and form settings can appear multiple times.

Duplicate ids shouldn't really be rendered in the page (not valid html) but that could be cleaned up in a separate PR if necessary (possibly using prefixes for any form tag input elements).

### DIFF
--- a/app/assets/javascripts/usage_charts.js
+++ b/app/assets/javascripts/usage_charts.js
@@ -15,9 +15,9 @@ $(document).ready(function() {
       chartConfig.mpan_mprn = meter;
     }
 
-    var series_breakdown = $(chartDiv).find("input[name='series_breakdown']").val();
-    if (series_breakdown) {
-      chartConfig.series_breakdown = series_breakdown;
+    var seriesBreakdown = $(chartDiv).find("input[name='series_breakdown']").val();
+    if (seriesBreakdown) {
+      chartConfig.series_breakdown = seriesBreakdown;
     }
 
     chartConfig.date_ranges = getDateRanges(chartDiv);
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/usage_charts.js
```diff
@@ -4,27 +4,23 @@ $(document).ready(function() {
...
     var chartConfig = chartContainer.data('chart-config');
 
-    var meter = $("#meter").val();
+    var meter = $(chartDiv).find("select[name='meter']").val();
     if (meter && meter != 'all') {
       chartConfig.mpan_mprn = meter;
     }
 
-    var series_breakdown = $("#series_breakdown").val();
+    var series_breakdown = $(chartDiv).find("input[name='series_breakdown']").val();
```
iant, line 18: should `series_breakdown` be `seriesBreakdown`?  (not that i have a problem with using snake case in js but i guess we need to stick to conventions)

app/assets/javascripts/usage_charts.js
```diff
@@ -33,26 +29,30 @@ $(document).ready(function() {
...
+    if (period == 'weekly') {
       var rangeExtension = 6;
     } else {
       var rangeExtension = 0;
     }
 
     // maintain this order of range addition to match input order to chart order
-    if($('input#second-date-picker').val()){
-      addRange($('input#second-date-picker').val(), dateRanges, rangeExtension);
+    var second_date = $(chartDiv).find("input[name='second-date-picker']").val();
```
iant, line 49: `secondDate`?

app/assets/javascripts/usage_charts.js
```diff
@@ -33,26 +29,30 @@ $(document).ready(function() {
...
+    if(second_date){
+      addRange(second_date, dateRanges, rangeExtension);
     }
-    addRange($('input#first-date-picker').val(), dateRanges, rangeExtension);
+
+    var first_date = $(chartDiv).find("input[name='first-date-picker']").val();
+    addRange(first_date, dateRanges, rangeExtension);
```
iant, line 55: `firstDate`?

iant (approved):
looks good to me @jhigman couple of comments on possible changes to var names

app/assets/javascripts/usage_charts.js
jhigman: 👍

app/assets/javascripts/usage_charts.js
jhigman: 👍

PR #2398, opened 2023-02-03
View component updates

urbanwide:
This PR updates the rest of the view components to use the 'component' helper.
Also have scoped the css to use the name "component-*component name*" for existing components as suggested by Leigh.

Next on my list is to organise the components directory a bit better. There are two options as I see it.
1) the standard way - keep the component rb file in the components directory and have a sidecar directory for the other assets: https://viewcomponent.org/guide/generators.html#place-the-view-in-a-sidecar-directory
2) Or the "extended sidecar" method which puts everything for a single component in it's own directory including the component rb file (this looks like it would require more effort to sort out - and possibly require the view_component-contrib gem) https://github.com/palkan/view_component-contrib#organizing-components-or-sidecar-pattern-extended

Any thoughts?

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -107,6 +107,9 @@ Rails.application.configure do
   config.hosts << "energysparks.development"
   config.hosts << "cy.energysparks.development"
   config.hosts << "mirror.energysparks.development"
+
+  # View components - Always place view in a sidecar directory when using the generator
+  config.view_component.generate.sidecar = true
 end
 
 class MyAppFormatter < Logger::Formatter


ldodds (approved):

ldodds:
Looks like the naming convention is `component_name-component`, e.g. `notice-component` rather than `component-notice`. I prefer that to what I suggested as it reads more like the component name.

re: organisation. While, I do quite like the extended pattern as a way to keep things together, I'm inclined to say let's go with the sidecar pattern as its part of the core gem. So we don't have to deal with an extra dependency or potential issues from non-standard configuration. We might still end up with a lot of `*.rb` files in one directory, but its the erb, scss, js which really clutter things up.

urbanwide:
Yes apologies @ldodds, I did end up with the convention compoment_name-component, which I agree reads better.

ldodds (approved):

PR #2399, opened 2023-02-06
Implement dynamic chart date ranges for subtitles

ldodds:
Our existing chart templates and javascript support dynamically changing the chart titles and subtitles. This is currently used:

- on the pupil charts. Here the title changes as you drill down, and some of the subtitles change to reflect the dates being viewed
- on the existing analysis pages, which have subtitles that change to reflect the date range being viewed
- some admin only pages which have both dynamic titles and subtitles

The text injected into those charts is generated by the analytics.

On the new advice pages we don't (currently) want to dynamically change the whole title and subtitle, but we do want to ensure that dates in the subtitles update as a user moves forward and backward on a chart.

This PR does several things:

- refactors the existing templates and JS code so that dynamically replaced titles (h3) and subtitles (h5) are scoped within a div (`.dynamic-titles`). This preserves the existing behaviour, while allowing new behaviour to be added to other elements
- changes the ChartComponent template so that the chart title, subtitle and header are within the `chart_wrapper`. This makes it easier to find and manipulate those elements
- extends the `ChartDataValues` class to dynamically add start and end dates to the JSON returned to the browser. These should reflect the "top-level" range that the user is viewing. This code uses the existing date ranges supplied by the analytics. The dates are only added/changed when the user moves forward and backwards at the top-level of the chart, and not if they zoom in and then move forward/backward.
- adds support in the Javascript to find and dynamically update `.start-date` and `.end-date` spans within a `.chart-subtitle` element, if that information is in the JSON
- changes the baseload template and YAML files to use this new feature. This required changing date formats and adding HTML to the translation key.

The intention is that we'll use this on all chart subtitles to allow them to be dynamically changed.

We may need to do some additional work to provide default dates when pages first load, but this can be done in a separate PR, or as part of building the specific pages.

### DIFF
--- a/app/models/chart_data_values.rb
+++ b/app/models/chart_data_values.rb
@@ -45,7 +45,6 @@ class ChartDataValues
       @y2_chart_type      = chart[:y2_chart_type]
       @annotations        = []
       @y2_axis_label = '' # Set later
-      puts transformations.inspect
       @transformations = transformations
       @allowed_operations = allowed_operations
       @drilldown_available = drilldown_available
--- a/app/models/chart_data_values.rb
+++ b/app/models/chart_data_values.rb
@@ -30,7 +30,7 @@ class ChartDataValues
       @title              = chart[:title]
       @subtitle           = chart[:subtitle]
       @x_axis_categories  = translate_categories_for(chart[:x_axis])
-      @x_axis_ranges      = chart[:x_axis_ranges] # Not actually used but range of actual dates
+      @x_axis_ranges      = chart[:x_axis_ranges]
       @x_max_value        = chart[:x_max_value]
       @x_min_value        = chart[:x_min_value]
       @chart1_type        = chart[:chart1_type]
@@ -45,6 +45,7 @@ class ChartDataValues
       @y2_chart_type      = chart[:y2_chart_type]
       @annotations        = []
       @y2_axis_label = '' # Set later
+      puts transformations.inspect
       @transformations = transformations
       @allowed_operations = allowed_operations
       @drilldown_available = drilldown_available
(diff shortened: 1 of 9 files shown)


app/models/chart_data_values.rb
```diff
@@ -183,13 +183,29 @@ def self.as_chart_json(output)
...
     ].inject({}) do |json, field|
       json[field] = output.public_send(field)
       json
     end
   end
 
+  def subtitle_start_date
+    return nil unless x_axis_ranges_present?
+    transformations_empty_or_only_move? ? format_subtitle_date(@x_axis_ranges.first.first) : nil
+  end
```
iant, line 198: perhaps this could be reworked so both `nil` returns are included in the early return?  (similar for `subtitle_end_date` below?)
```
  def subtitle_start_date
    return nil unless x_axis_ranges_present? && transformations_empty_or_only_move?
    
    format_subtitle_date(@x_axis_ranges.first.first)
  end
```

app/models/chart_data_values.rb
```diff
@@ -493,4 +509,17 @@ def y2_is_solar?(y2_data_title)
...
   end
+
+  def x_axis_ranges_present?
+    @x_axis_ranges.present? && !@x_axis_ranges.empty?
+  end
+
+  def transformations_empty_or_only_move?
+    return true if @transformations.nil? || @transformations.empty?
+    return true if @transformations.length == 1 && transformation_type(@transformations[0]) == :move
+  end
```
iant, line 520: likely not necessary but might be good to return a false at the end of this method
```
  def transformations_empty_or_only_move?
    return true if @transformations.nil? || @transformations.empty?
    return true if @transformations.length == 1 && transformation_type(@transformations[0]) == :move

    false
  end
```

app/views/admin/analysis/generic_chart_template.html.erb
```diff
@@ -18,7 +18,9 @@ with <%= @school.number_of_pupils %> pupils and a floor area of <%= @school.floo
     <hr class="analysis"/>
   <% end %>
   <div id="chart_wrapper_<%= chart %>" class="chart-wrapper">
-    <h3 class="analysis">Loading: <%= chart.to_s.humanize %></h3>
+    <div class="dynamic-titles">
+      <h3 class="analysis">Loading: <%= chart.to_s.humanize %></h3>
```
iant, line 22: does "Loading" need extracting to a translation yml file?

iant (approved):
Looks good to me @ldodds a few small comments on possible code reworks and a possible string that needs extracting.

app/views/admin/analysis/generic_chart_template.html.erb
ldodds: This is an admin only page, that will get retired soon.

PR #2401, opened 2023-02-07
Add checks for fuel type and enough data to advice base controller

ldodds:
Enforces a consistent approach for checking whether we have the right data available to run some analysis across all the advice page controllers.

This PR changes the base class so that:

- if a school doesn't have the right fuel type, we render a custom error page. This is only likely to happen if all the meters for a specific fuel type has been disabled, but a user has previously bookmarked a url. Ensures we provide a more helpful message
- if we can't run the analysis because we don't have enough data, then we render a custom error page which will ideally indicate when the analysis will be available
- improves the existing custom error handling to preserve the page tabs as these should have already been set in the request cycle. The error page also now returns a 500 status code

Individual page controllers can implement `create_analysable` and return an object that conforms to the `Analysable` mixin from the analytics package. The mixin will return a boolean to indicate if there is `enough_data?` and optionally a date when the analysis should be available from its `available_from` method.

From its `create_analysable` method, a specific page controller might:

- return a custom object or an OpenStruct that implement the right methods (see the advice page spec)
- create a service from the analytics which already implements Analysable and return that as a means of doing the check
- if could even check the `@tab` variable and return different objects for the different tabs if there are different data requirements

This should hopefully simplify this type of error checking. 

We might want to later allow for some more customised messages for individual "not enough data" pages, but this should be easier to add later. E.g. by adding additional content managed field to the AdvicePage model.

### DIFF
--- a/app/controllers/schools/advice/advice_base_controller.rb
+++ b/app/controllers/schools/advice/advice_base_controller.rb
@@ -11,6 +11,7 @@ module Schools
       before_action :load_advice_page, only: [:insights, :analysis, :learn_more]
       before_action :check_authorisation, only: [:insights, :analysis, :learn_more]
       before_action :load_recommendations, only: [:insights]
+      before_action :check_has_fuel_type, only: [:insights, :analysis]
       before_action :check_can_run_analysis, only: [:insights, :analysis]
       before_action :set_data_warning, only: [:insights, :analysis]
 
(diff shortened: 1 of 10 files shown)


iant (approved):
looks good to me @ldodds one comment/suggestion on possibly splitting `check_can_run_analysis`

PR #2411, opened 2023-02-09
Initial analysis page layout and first chart

jhigman:
This PR implements the analysis tab of the electricity intraday advice page, with charts and text.

It includes support for the Analysable checks for date ranges.

It also factors out some support methods into the AdviceBaseController for setting up analysis dates.


iant (approved):

PR #2413, opened 2023-02-09
Allow meters to be associated with school issues and notes

urbanwide:


iant (approved):
looks good to me @urbanwide

PR #2416, opened 2023-02-09
Implement gas out of hours analysis page

iant:


spec/system/schools/advice_pages/gas_out_of_hours_spec.rb
```diff
@@ -3,14 +3,16 @@
...
-  let(:expected_page_title) { "Gas out of hours usage analysis" }
+  let(:expected_page_title) { "Out of school hours gas use" }
   include_context "gas advice page"
 
   context 'as school admin' do
 
     let(:user)  { create(:school_admin, school: school) }
 
     before do
+      allow_any_instance_of(Usage::AnnualUsageBreakdownService).to receive(:usage_breakdown) { nil }
```
ldodds, line 14: It would be better if you could include some basic checks for each page (applies to the electricity one too), e.g:

- are the right sections showing?
- has the right chart(s) been added
- are the expected numbers on the page

The latter means stubbing the data, but should all be fairly simple.

ldodds (changes requested):

iant:
@ldodds as discussed will merge this down and pick up improved system tests for this and electricity out of hours in a separate pr

PR #2420, opened 2023-02-10
Stub out n3rgy api

ldodds:
Looks like some of the tests were calling the sandbox api rather than being properly stubbed out.


iant (approved):
Looks good to me

PR #2423, opened 2023-02-10
Intraday electricity use insights page

iant:


app/controllers/schools/advice/electricity_intraday_controller.rb
```diff
@@ -2,6 +2,10 @@ module Schools
   module Advice
     class ElectricityIntradayController < AdviceBaseController
       def insights
+        @peak_usage_benchmarking = build_peak_usage_benchmarking
+        @peak_usage_calculation = build_peak_usage_calculation(asof_date)
+        @peak_usage_calculation_1_year_ago = build_peak_usage_calculation(previous_years_asof_date)
```
ldodds, line 7: this looks like it could fail if the school doesn't have >1 years of data. So I'd add a check for this in the code and return nil from the calculation. The template can then conditionally include the percentage.

app/views/schools/advice/electricity_intraday/_insights.html.erb
```diff
@@ -0,0 +1,50 @@
...
+    <th class="text-right">
+      <%= t('advice_pages.electricity_intraday.insights.your_current_peak_electricity_use.table.percent_change') %>
+    </th>
+  </thead>
+  <tr>
+    <td class="text-left">
+      <%= format_unit(@peak_usage_calculation.average_peak_kw, :kw) %>
+    </td>
+    <td class="text-right">
+      <%= format_unit(@peak_kw_usage_percentage_change, :percent) %>
```
ldodds, line 22: Wrap this in the `up_downify` helper (see baseload and long term usage insights for examples). This will give the coloured up and down arrows. And check whether we've been able to calculate

app/controllers/schools/advice/electricity_intraday_controller.rb
```diff
@@ -10,6 +14,36 @@ def analysis
 
       private
 
+      def calculate_percentage_change_in_peak_kw
+        # Calculates the relative change between last years average peak kw value
+        # and this years average peak kw value
+        old_peak_kw = @peak_usage_calculation_1_year_ago.average_peak_kw
+        new_peak_kw = @peak_usage_calculation.average_peak_kw
+        ((old_peak_kw - new_peak_kw) / old_peak_kw) # * 100
```
ldodds, line 22: check calc?

ldodds (changes requested):

ldodds (approved):

PR #2425, opened 2023-02-10
Gas long term usage

ldodds:
Implements the insights and analysis pages for the gas long term use analysis.

There's quite a lot of common code and templating between this and the electricity controller.


app/controllers/schools/advice/base_long_term_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+          @annual_usage_meter_breakdown = usage_service.annual_usage_meter_breakdown
+          @meters_for_breakdown = sorted_meters_for_breakdown(@annual_usage_meter_breakdown)
+        end
+      end
+
+      private
+
+      def multiple_meters?
+        @school.meters.active.where(meter_type: fuel_type).count > 1
+      end
```
iant, line 34: not something to change in this pr but maybe this could be a method on the meters or school model?

iant (approved):
@ldodds looks good to me

PR #2426, opened 2023-02-10
Assign Becca as the default issues admin for all groups without one set already

urbanwide:


iant (approved):
looks good to me @urbanwide

PR #2427, opened 2023-02-10
Remove gas intraday page

ldodds:
We're removing this page for the moment, so removing the database entry, nav bar entry, routes, controller and partials.


iant (approved):
looks good @ldodds

PR #2431, opened 2023-02-13
Run i18n tasks earlier in build

ldodds:
We currently check status of i18n files whilst running the specs.

Proposing that we change the workflow to explicitly run the `i18n-tasks health` task before the specs so that they fail early, rather than at the end of the build.

At the moment I've left the rspec test in place, but this could possibly be removed now. In testing I've noticed that its takes up to 45 seconds to run the i18n-spec file so we probably don't want to run these checks twice.


iant (approved):
sounds like a much better approach.  pr looks good to me @ldodds

PR #2432, opened 2023-02-13
Sorting of meter tables

ldodds:
Change default sort order for tables on baseload analysis:

- meter breakdown sorted by % baseload (desc)
- seasonal and intraday variation sorted by estimated saving (desc)


iant (approved):
looks good to me @ldodds

PR #2437, opened 2023-02-14
Implement electricity recent progress insights page

iant:
Screenshot 2023-02-14 at 14 03 05

### DIFF
--- a/app/controllers/schools/advice/electricity_recent_changes_controller.rb
+++ b/app/controllers/schools/advice/electricity_recent_changes_controller.rb
@@ -2,6 +2,7 @@ module Schools
   module Advice
     class ElectricityRecentChangesController < AdviceBaseController
       def insights
+        @analysis_dates = analysis_dates
         @recent_usage = build_recent_usage
       end
 
@@ -11,6 +12,18 @@ module Schools
       end
 
       private
+
+      def analysis_dates
+        start_date = aggregate_school.aggregated_electricity_meters.amr_data.start_date
+        end_date = aggregate_school.aggregated_electricity_meters.amr_data.end_date
+        OpenStruct.new(
+          start_date: start_date,
+          end_date: end_date,
+          one_years_data: one_years_data?(start_date, end_date),
+          recent_data: recent_data?(end_date),
+          months_analysed: months_analysed(start_date, end_date)
+        )
+      end
 
       def build_recent_usage
         OpenStruct.new(
--- a/app/controllers/schools/advice/electricity_recent_changes_controller.rb
+++ b/app/controllers/schools/advice/electricity_recent_changes_controller.rb
@@ -2,10 +2,7 @@ module Schools
   module Advice
     class ElectricityRecentChangesController < AdviceBaseController
       def insights
-        @recent_usage = ::Usage::RecentUsageComparisonService.new(
-          meter_collection: aggregate_school,
-          fuel_type: :electricity
-          ).recent_usage
+        @recent_usage = build_recent_usage
       end
 
       def analysis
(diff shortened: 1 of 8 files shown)


app/controllers/schools/advice/electricity_recent_changes_controller.rb
```diff
@@ -2,6 +2,10 @@ module Schools
   module Advice
     class ElectricityRecentChangesController < AdviceBaseController
       def insights
+        @recent_usage = ::Usage::RecentUsageComparisonService.new(
```
ldodds, line 5: As noted on Trello, suggest we revise this service to allow it to return comparison for last 2 school weeks or last 4 weeks.

config/locales/views/advice_pages/electricity_recent_changes.yml
```diff
@@ -18,6 +18,16 @@ en:
         summary: This section gives a more detailed analysis of how your school’s electricity consumption has changed recently. It can be particularly useful if trying to investigate a change in electricity consumption or monitoring the impact of action taken in the school.
         title: Analysis
       insights:
+        how_do_you_compare:
+          summary: Compare your recent electricity use against other schools in your group
+          title: How do you compare?
+        link: Learn more
         next_steps: Next steps
```
ldodds, line 25: This should be the sentence under "What should you do next?" in the doc

app/views/schools/advice/electricity_recent_changes/_insights.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  <tr>
+    <td><%= t('advice_pages.electricity_recent_changes.insights.your_recent_electricity_use.table.difference') %></td>
+    <td class="text-right"><%= format_unit(@recent_usage.recent_usage_comparison.kwh, :kwh) %></td>
+    <td class="text-right"><%= format_unit(@recent_usage.recent_usage_comparison.£, :£) %></td>
+    <td class="text-right"><%= format_unit(@recent_usage.recent_usage_comparison.co2, :co2) %></td>
+  </tr>
+</table>
+
+<%= render 'schools/advice/section_title', section_id: 'how_do_you_compare', section_title: t('advice_pages.electricity_recent_changes.insights.how_do_you_compare.title') %>
+<%= t('advice_pages.electricity_recent_changes.insights.how_do_you_compare.summary') %>
```
ldodds, line 36: This shouldn't be static text. It should check if there's a school group for this school and if so include a link to the correct school comparison page. See comment in the doc for the name of the benchmark and example code in other pages, e.g. baseload/long term changes

config/locales/views/advice_pages/electricity_recent_changes.yml
```diff
@@ -18,6 +18,16 @@ en:
...
-        title: Insights
+        summary: This page makes note of whether your electricity use has increased or decreased recently. Monitoring recent changes to your energy use can help you to assess the impacts of changes you have made, or quickly identify problems
+        title: What do we mean by recent changes?
+        your_recent_electricity_use:
+          table:
+            date_range: Date range
```
ldodds, line 30: Call this "period" instead. And I think there's a key for that in `advice_pages.yml` already.

config/locales/views/advice_pages/electricity_recent_changes.yml
```diff
@@ -18,6 +18,16 @@ en:
...
+            difference: Difference
```
ldodds, line 31: Maybe "Change" instead?

app/views/schools/advice/electricity_recent_changes/_insights.html.erb
```diff
@@ -0,0 +1,36 @@
+<%= component 'notice', status: :neutral do |c| %>
+  <% c.with_link { link_to t('advice_pages.electricity_recent_changes.insights.link'), learn_more_school_advice_electricity_recent_changes_path(@school) } %>
+  <%= t('advice_pages.electricity_recent_changes.insights.summary') %>.
+<% end %>
+
+<%= render 'schools/advice/section_title', section_id: 'your_recent_electricity_use', section_title: t('advice_pages.electricity_recent_changes.insights.your_recent_electricity_use.title') %>
+
+<table class="mt-2 table table-sm advice-table">
```
ldodds, line 8: Can you add a spec to test for the expected values on the page?

ldodds (changes requested):

app/views/schools/advice/electricity_recent_changes/_insights.html.erb
ldodds, line 8: Given that the comparison is for last 2 weeks, if we have no recent data then I don't think we can show the detail in the table.

We already show a standard notice, but if there's no recent data, I suggest you exclude the detail for each row. See what I did in the baseload insights page.

ldodds (changes requested):

app/controllers/schools/advice/electricity_recent_changes_controller.rb
```diff
@@ -11,6 +13,48 @@ def analysis
 
       private
 
+      def analysis_dates
```
ldodds, line 16: Not sure if you need this, looks to be the same method as in the base class?

ldodds (changes requested):

ldodds (approved):

PR #2442, opened 2023-02-16
Implement heating control page

ldodds:
Implements the initial version of the heating control page.

Sorry this has turned into a large PR due to hitting some issues along the way. 

* Includes a bump of the analytics gem to pick up updated service code
* Reworks the existing scaffolded "boiler_control" page to become "heating_control". This includes renaming the route and all the scaffolded files. There's also an AfterParty task to rename the AdvicePage model for the page
* Adds the `HeatingControlController,` which delegates to the `HeatingControlService` to coordinate across the various parts of the analytics
* Most the files are new or updated partials, for the insights and analysis sections of the heating control page

The heating control page includes a section which has meter-specific charts. This allows a user to switch between different charts.

I've mostly reused the code written to implement this for the baseload page. But in doing that I noticed some bugs in those charts, so have had to:

* update the `usage_charts.js` file to ensure that chart controls aren't initialised twice and reset chart transformations when the meter is changed
* fix the default "no zoom" option on chart controller. This should be off by default
* reworked the baseload page to generalise the meter-specific chart code. There's now a shared partial across these two pages. This could later be turned into an extended version of the Chart Component
* in extracting the partial I improved the formatting of the controls on the meter-specific charts and fixed up the date formatting in the subtitle.

Other improvements that can follow:

- some further tidy up of the helpers used for the meter-specific charts
- extracting `analysis_dates` into a more generalised helper for reuse across controllers


app/assets/javascripts/analysis.js
```diff
@@ -252,6 +252,7 @@ function processAnnotations(loaded_annotations, chart){
 function setupAnalysisControls(chartContainer, chartConfig){
   var controls = $(chartContainer).parent().find('.analysis-controls');
   if(controls.length){
+    //console.log("Setup analysis controls: " + chartConfig.type + "meter: " + chartConfig.mpan_mprn);
```
iant, line 255: can this be removed?

iant (approved):
:+1: this looks good to me @ldodds

PR #2443, opened 2023-02-17
Remove chart subtitle for intraday chart plus cosmetic fixes

ldodds:
* Removes the dynamic dates in the subtitle for an intraday chart.
* Revises title of first chart
* Parameterise description of benchmark schools using school type.
* Improved the rspec to check the right charts are on the page.

The chart framework doesn't return the correct date range for intraday charts. Am unclear how to fix that currently without potentially impacting other code. Because in this case the chart date ranges are clearly labelled as series, I've removed the dates on the chart for now.


iant (approved):
looks good to me @ldodds

PR #2446, opened 2023-02-17
Implement thermostatic insights & analysis pages

iant:


ldodds (changes requested):
Check whether there's enough data, by returning the thermostatic service from create_analysable.

ldodds (approved):

PR #2447, opened 2023-02-17
Implement total energy use page

ldodds:
Adds the Total Energy Use page

This page has a combination of information from other sources, so there's some reuse of partials, etc.

* Updates advice base controller to allow subtitles and "next steps" to be optional
* Removes some unneeded variables in the long term controller (spotted while testing)
* Incorporates the table from the school dashboard, so reuses partial and existing code. Partial has been updated to allow cost saving to be hidden
* Extracts partials for the tables from electricity/gas long term analysis so they can be reused

Page uses the new advice page translation helper (`advice_t`) to reduce length of keys. However I found that this was tripping up `i18n-tasks`, so I have also:

* revised the helper to require passing in the page key, e.g. `total_energy_use` rather than relying on instance variable
* added an i18n pattern matcher for the `advice_t` function so that it can figure out the full key.

Felt it was better to compromise a bit on the helper, to have i18n as a safety net.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,7 @@ gem 'pg_search'
 gem 'calculate_in_group'
 
 # Dashboard analytics
-gem 'energy-sparks_analytics', git: 'https://github.com/Energy-Sparks/energy-sparks_analytics.git', tag: '2.4.9'
+gem 'energy-sparks_analytics', git: 'https://github.com/Energy-Sparks/energy-sparks_analytics.git', tag: '2.4.10'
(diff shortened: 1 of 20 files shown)


ldodds:
Note to myself for Monday: build is failing as there's some shared tests which don't set up necessary context to visit the page. Some of these can probably be removed as they overlap with tests in other pages which are being run as shared examples.

urbanwide (approved):
Looks good @ldodds 👍

PR #2448, opened 2023-02-17
Colour benchmark and exemplar schools on charts

ldodds:
Use different colours for the "Benchmark" and "Exemplar" bars in the school comparison charts. 

Detects the series name (using whichever translated text is being used) and then replaces the relevant series data with an object that specifies a colour. This will trigger highcharts to colour the bar differently to the rest of the y-axis.

Also changes the default colours for electricity and gas so we go from:

- "dark" (school consumption)
- "middle" (benchmark)
- "light" (exemplar)


urbanwide (approved):
Looks good @ldodds 👍

PR #2460, opened 2023-02-22
Add new visual comparison to long term and total use pages

ldodds:
Adds the SchoolComparisonComponent to the gas and electricity long term use pages, plus the total energy use page.

* Removes the partials used on the three pages
* Add the school comparison component, updating the controller to return a School::Comparison not an OpenStruct
* Removes the old advice page helpers which are no longer needed
* Removes unneeded translation keys
* Changes specs to just test that the comparison sections are present, no need to test the contents as this is done separately in the component specs.


iant (approved):
looks good to me @ldodds

PR #2464, opened 2023-02-23
Add footnote modal component

iant:
- [ ] extract css
- [ ] optional modal-dialog classes


app/components/footnote_modal_component.rb
```diff
@@ -0,0 +1,17 @@
...
+  def initialize(title:)
+    @title = title
+  end
+
+  def icon
+    helpers.fa_icon('question-circle')
+  end
+
+  def modal_id
+    'footnote-modal-' + object_id.to_s
```
ldodds, line 15: When the model is added to a page, other elements will need to reference its id -- the bootstrap support requires having another element that refers to it. So I think it would be better to have the id specified in the constructor as a required parameter.

app/components/footnote_modal_component/footnote_modal_component.html.erb
```diff
@@ -0,0 +1,21 @@
+<a href="#" data-toggle="modal" data-target="#<%= modal_id %>"><%= @title %>&nbsp;<span style="color: #007bff;"><%= icon %></span></a>
```
ldodds, line 1: This can't be part of the modal template, as the toggle could appear anywhere on the page, sometimes multiple times).

app/components/footnote_modal_component/footnote_modal_component.html.erb
```diff
@@ -0,0 +1,21 @@
...
+<div>
+  <div class="modal fade" id="<%= modal_id %>" tabindex="-1" role="dialog" aria-labelledby="#<%= modal_id %>-title" aria-hidden="true">
+    <div class="modal-dialog modal-lg modal-dialog-centered" role="document">
+      <div class="modal-content">
+        <div class="modal-header" style="background: #232b49 !important; color: white;">
```
ldodds, line 6: All of the styling should be extracted out into an scss file that sits alongside the component.

app/components/footnote_modal_component.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class FootnoteModalComponent < ViewComponent::Base
+  renders_one :body_content
+
+  def initialize(title:)
```
ldodds, line 6: Nice to have: allow for some optional HTML classes that will be added to the div with the `modal-dialog` class. The defaults would be `modal-lg modal-dialog-centered` which controls the size and centering of the model. Would allow the size to be tweaked for different parts of the site.

ldodds (changes requested):

ldodds (approved):

ldodds:
Have approved, content review has finished, so ready for development. Suggest doing on one page first, e.g. baseload, so we can ensure its working, then add to the others pages in a final PR.

PR #2467, opened 2023-02-23
Fix bug in multiple fuel benchmark series colours

ldodds:
Fixes a bug introduced in my earlier PR to update the colours on benchmark charts.

When a chart has multiple fuel types, the new code wasn't assigning the right colours for each fuel type. So this PR fixes that and introduces a spec.

### DIFF
--- a/app/models/chart_data_values.rb
+++ b/app/models/chart_data_values.rb
@@ -585,7 +585,7 @@ private
         LIGHT_ELECTRICITY
       end
     else
-      return DARK_STORAGE
+      DARK_STORAGE
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/models/chart_data_values.rb
```diff
@@ -536,31 +536,60 @@ def is_benchmark_chart?
...
+        LIGHT_GAS
+      end
+    when 'Electricity'
+      if benchmark_school_category?(category)
+        MIDDLE_ELECTRICITY
+      else
+        LIGHT_ELECTRICITY
+      end
+    else
+      return DARK_STORAGE
```
iant, line 588: return not needed?

iant (approved):
Looks good to me @ldodds

PR #2469, opened 2023-02-23
Factor out default url options and test from bill request email

jhigman:
When emails are sent using a specific locale, the links in the email should use the same locale subdomain.

The functionality to format links is in the url helpers, but call a default_url_options helper method which can override to supply a specific host, dependent on locale.

If no host names are available in the APPLICATION_HOST or WELSH_APPLICATION_HOST env vars, we just use the default url options.


iant (approved):
looks good to me @jhigman :+1:

PR #2471, opened 2023-02-24
Rebuild tx config with finished pages

ldodds:
Just updates the transifex config with the finished advice pages.

Has been used to push text to transifex.


iant (approved):
looks good to me @ldodds

PR #2479, opened 2023-02-24
Add modal link to other advice pages

iant:


app/views/schools/advice/electricity_long_term/_current_usage.html.erb
```diff
@@ -24,6 +24,7 @@
     </tr>
   </tbody>
 </table>
+<%= render 'schools/advice/how_have_we_analysed_your_data_table_caption', data_target: "how-have-we-analysed-your-data-footnotes" %>
```
ldodds, line 27: This won't look right with 2 separate lines. We need to combine into a single table footnote. Suggest just replacing the existing footnote with the new standard on, and moving the "Review our analysis" to a separate line without the right alignment

app/views/schools/advice/electricity_recent_changes/_your_recent_electricity_use_table.html.erb
```diff
@@ -18,6 +18,7 @@
       period: t('advice_pages.electricity_recent_changes.insights.your_recent_electricity_use.table.change')
   %>
 </table>
+<%= render 'schools/advice/how_have_we_analysed_your_data_table_caption', data_target: "how-have-we-analysed-your-data-footnotes" %>
```
ldodds, line 21: Same as above, need to have a single table footnote with both bits of content.

ldodds (changes requested):
Tables should have a single footnote, so we either combine (where they have different content) or separate them out. Have left comments on how to handle the situations I've flagged up.

iant:
hi @ldodds just pushed some updates addressing your review comments.  I think this is ok for a final review.

ldodds (approved):

PR #2481, opened 2023-02-27
Fix up failing spec

ldodds:
Fixes up a system spec: the expected content is no longer included on the page, its been replaced by the modal.


urbanwide (approved):
Looks good 👍

PR #2484, opened 2023-02-28
Fix intermittent failures in next activity suggester test

iant:


iant:
closing this for now.  will re-op/re-visit if it becomes an issue again

PR #2488, opened 2023-03-02
Implement new advice page index

ldodds:
Implements the initial version of the new advice page index.

* Replaces existing page with a new tabbed layout
* Adds a default tab that lists all the pages relevant to a school, with a brief description
* Adds a "priorities" tab that includes the current "management priorities" that is currently on a separate page, but with a new layout
* Adds a "recent alerts" tab with a list of the schools recent alerts, using the new notice component

The links to the advice pages on the priorities and recent alerts tab are either missing or just placeholders for now. Some other changes are required to relate AlertTypes to AdvicePages. This will be done in a follow-up PR

To group together the partials used for these new pages, I've added an `index` directory. Not sure if that's the best option or not, but helps keep related files together.


urbanwide (approved):
Looks good @ldodds 👍

PR #2490, opened 2023-03-02
Add extra fields to alert type

ldodds:
Adds some extra fields to `AlertType`:

* a reference to `AdvicePages` which will allow us to link alerts to their pages, as well as redirect users from the old analysis pages to the new versions
* a `group` attribute that will be used to categorise the alerts based on whether they are carrying out benchmarking of a school, reporting on short term changes, indicate some priority areas for attention, or are just general advice. This will allow us to improve the display and prioritisation of alert content on the advice page index
* a `link_to` property to indicate which tab of the advice page to link to, defaulting to insights
* a `link_to_section` property, to indicate a section of the insights or analysis page to link to

The last two are to allow admins to have some control over where to send users clicking on a dashboard alert.

Also includes: 

* an after party task to populate the group attribute
* an after party task to populate the advice page reference

The other changes will rely on admins editing the content.

### DIFF
--- a/app/models/advice_page.rb
+++ b/app/models/advice_page.rb
@@ -19,6 +19,7 @@ class AdvicePage < ApplicationRecord
 
   translates :learn_more, backend: :action_text
 
+  has_many :alert_types
   has_many :advice_page_activity_types
   has_many :activity_types, through: :advice_page_activity_types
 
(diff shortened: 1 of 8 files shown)


urbanwide (approved):
Looks good @ldodds 👍

PR #2492, opened 2023-03-02
Fixes for language selector on password reset

jhigman:


iant (approved):
looks good to me @jhigman

PR #2493, opened 2023-03-02
Update alert type management forms

ldodds:
Updates the admin pages for managing Alert Types

* Tidy up the index page to make tables sortable, add in the extra fields
* Add extra fields for the overview tab of individual alert types
* Update the form to allow admin to set group, advice page, link to, etc.


iant (approved):
looks good to me @ldodds

PR #2495, opened 2023-03-02
Change find out more linking

ldodds:
Changes whether the "find out more" pages are accessible to users. This is part of retiring that part of the application, replacing them with direct links to the new advice pages.

Uses a Feature Flag to control the behaviour change (`FEATURE_FLAG_REPLACE_FIND_OUT_MORES`)

The majority of the work is done in some new helper methods in the `SchoolsHelper`. This checks the status of the feature flag and then returns either the path to the find out more page (if its off) or the advice page (if there is one). I've added an rspec test which exercises the majority of the paths through this code.

The rest of the changes are then largely in the templates, which now call these methods.

Summary of changes:

* [x] Updates the find out more controller to redirect users to the corresponding advice page if there is one,  or the advice index as a fallback, rather than letting them view the find out more page. Admins can still view the content
* [x] Updates the display of alerts on the pupil and school dashboard to link to the corresponding advice page, rather than find out more
* [x] Changes links in the mgt priorities list on the dashboard, and the dedicated page (which will also later be removed)
* [x] Changes links in alert emails.
* [x] Some minor tidy up of i18n keys for "Find out more" link

### DIFF
--- a/app/controllers/schools/find_out_more_controller.rb
+++ b/app/controllers/schools/find_out_more_controller.rb
@@ -1,9 +1,12 @@
 module Schools
   class FindOutMoreController < ApplicationController
+    include AdvicePageHelper
+
     load_and_authorize_resource :school
     load_and_authorize_resource
 
     skip_before_action :authenticate_user!
+    before_action :redirect_if_disabled_and_not_admin
 
     def show
       activity_type_filter = ActivityTypeFilter.new(school: @school, scope: @find_out_more.activity_types, query: { exclude_if_done_this_year: true })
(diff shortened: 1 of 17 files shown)


app/controllers/schools/find_out_more_controller.rb
```diff
@@ -21,5 +24,15 @@ def show
     def content_managed?
       @find_out_more.alert.alert_type.class_name == "Alerts::System::ContentManaged"
     end
+
+    def redirect_if_disabled_and_not_admin
+      return unless EnergySparks::FeatureFlags.active?(:replace_find_out_mores)
```
iant, line 29: maybe this could go with the `before_action`? e.g.
```
    before_action :redirect_if_disabled_and_not_admin, if: -> { EnergySparks::FeatureFlags.active?(:replace_find_out_mores) }
```

iant (approved):
looks good to me @ldodds :+1:

app/controllers/schools/find_out_more_controller.rb
ldodds: Interesting, hadn't thought of that. I might leave is is for now as the idea is we'll remove all this code once we've launched the new pages.

PR #2496, opened 2023-03-03
Implement the choose a benchmark list page

urbanwide:
This is getting quite big so am drawing a line under it here (It still needs to use the correct links to the benchmarks results pages and maintain filters between pages)

This also changes the tabs to be of the local variety (with js) as it's much faster for the end user and less overhead at our end too.

It also adds the benchmarks & schools counts on the index page. Hopefully I am using analytics correctly there.


ldodds (approved):

urbanwide:
Merging with one failed test - it's one of the known flickering ones though so assume this is ok!

PR #2498, opened 2023-03-03
Add real links to advice index page tabs

ldodds:
Updates the list of priorities and recent alerts to link to the actual advice pages (and sections where configured).


iant (approved):
:+1:

PR #2502, opened 2023-03-06
Factor user table into partial

jhigman:
The users for a school are managed in two tables, one for school admins and one for staff. The tables should use a consistent set of columns, to include preferred language.


urbanwide (approved):
Looks good @jhigman 👍

PR #2505, opened 2023-03-06
Helper method to translate user roles

jhigman:


urbanwide (approved):
Looks good @jhigman 👍

PR #2506, opened 2023-03-06
Added preferred locale field to contact form

jhigman:
When a contact has an associated user, the contact form should include a preferred language field, which reflects the preferred locale set on the user.

The language should be changeable from the contact form, and applied to the user.

If the contact doesn't have a user, then we don't show the preferred locale field.


urbanwide (approved):
Looks good @jhigman 👍

ldodds (approved):

PR #2515, opened 2023-03-09
Revise recent onboarding report

ldodds:
Revises the admin recent onboarding report:

* Order so most recent first
* Change "Onboarding event" column to be date-time when onboarding completed, not last event date
* Add column for whether school is data enabled
* Add column for date-time when school was data enabled
* Calculate number of days between onboarding completed and data being enabled

Pushed most of the logic into the SchoolOnboarding model.

### DIFF
--- a/app/views/admin/school_onboardings/completed.html.erb
+++ b/app/views/admin/school_onboardings/completed.html.erb
@@ -1,4 +1,4 @@
-.<h1>Schools recently onboarded</h1>
+<h1>Schools recently onboarded</h1>
 
 <table class="table table-sorted table-sm">
   <thead>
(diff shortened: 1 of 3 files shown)


app/views/admin/school_onboardings/completed.html.erb
```diff
@@ -1,4 +1,4 @@
-<h1>Schools recently onboarded</h1>
+.<h1>Schools recently onboarded</h1>
```
jhigman, line 1: Spurious character..?

app/views/admin/school_onboardings/completed.html.erb
```diff
@@ -7,7 +7,10 @@
       <th>School group</th>
       <th>School name</th>
       <th>Contact email</th>
-      <th>Onboarded at event</th>
+      <th>Onboarded Completed</th>
```
jhigman, line 10: "Onboarding" Completed..?

jhigman (approved):

PR #2518, opened 2023-03-10
3034 make data download visible to group admins

jhigman:


app/controllers/schools/downloads_controller.rb
```diff
@@ -1,8 +1,18 @@
...
 
     def index
     end
+
+    private
+
+    def authorized?
+      unless can?(:download_school_data, @school)
+        flash[:error] = "You are not authorized to view that page."
+        redirect_to root_path
```
ldodds, line 14: Maybe redirect to school dashboard instead of home page?

ldodds (approved):

app/controllers/schools/downloads_controller.rb
jhigman: 👍

PR #2521, opened 2023-03-10
Implement gas cost advice page

ldodds:
Implements the gas costs advice page.

* Extracts code from electricity costs controller to create a common base class to be used for gas and electricity controllers
* Copies over the templates, specs and translation text for the gas page, adjusting values/keys as required


app/views/schools/advice/gas_costs/_meter_breakdown.html.erb
```diff
@@ -0,0 +1,22 @@
+<%= render 'schools/advice/section_title', section_id: 'meter-breakdown', section_title: advice_t('gas_costs.analysis.meter_breakdown.title') %>
+
+<%= simple_form_for :school, url: meter_costs_school_advice_electricity_costs_path(school, format: :js), method: :get, remote: true do |f| %>
+    <div class="form-group">
```
iant, line 4: spacing looks like it might be a bit off here?

iant (approved):
looks good to me @ldodds

PR #2525, opened 2023-03-14
Implement advice index page benchmarking framework

ldodds:
An initial PR a small framework for generating the benchmark school comparisons to be shown on the advice index page.

For each of the advice pages, we will aim to generate a comparison of that schools data against a "benchmark" or "exemplar" school. E.g. a comparison of their current baseload against what we consider to be good values.

We will generate each of these benchmarks, for each school, every day.

This initial PR covers:

* Added a new model `AdvicePageSchoolBenchmark` which holds the relationship between a page and a school, along with the result of the benchmark
* A base class `SchoolBenchmarkGenerator` which can be extended to implement the benchmarks for each page. The base class just handles creating/updating/removing the above model, but delegates to sub-classes to perform the actual benchmarking
* An initial benchmark implementation for the baseload page
* The above class reuses existing code. This has been moved from the baseload controller to the `BaseloadService` so that it can be called by the generator. This ensures that the advice page and the generator are using the same logic
* I've written tests for the new classes, and updated the existing specs for the baseload controller, service and page to reflect the refactoring

Next step is to add this benchmarking framework to the nightly job.

### DIFF
--- a/app/services/schools/advice/baseload_service.rb
+++ b/app/services/schools/advice/baseload_service.rb
@@ -143,7 +143,6 @@ module Schools
         )
       end
 
-
       private
 
       def asof_date
--- a/app/services/schools/advice/baseload_service.rb
+++ b/app/services/schools/advice/baseload_service.rb
@@ -130,6 +130,20 @@ module Schools
         end
       end
 
+      def benchmark_baseload
+        average_baseload_kw_last_year = average_baseload_kw(period: :year)
+        average_baseload_kw_benchmark = average_baseload_kw_benchmark(compare: :benchmark_school)
+        average_baseload_kw_exemplar = average_baseload_kw_benchmark(compare: :exemplar_school)
+
+        Schools::Comparison.new(
+          school_value: average_baseload_kw_last_year,
+          benchmark_value: average_baseload_kw_benchmark,
+          exemplar_value: average_baseload_kw_exemplar,
+          unit: :kw
+        )
+      end
+
+
       private
 
       def asof_date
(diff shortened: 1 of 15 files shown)


app/services/schools/advice/baseload_service.rb
```diff
@@ -130,6 +130,20 @@ def date_ranges_by_meter
...
+
+        Schools::Comparison.new(
+          school_value: average_baseload_kw_last_year,
+          benchmark_value: average_baseload_kw_benchmark,
+          exemplar_value: average_baseload_kw_exemplar,
+          unit: :kw
+        )
+      end
+
+
```
iant, line 146: looks like there might be an extra line here?

app/services/schools/advice_page_benchmarks/baseload_benchmark_generator.rb
```diff
@@ -0,0 +1,17 @@
+module Schools
+  module AdvicePageBenchmarks
+    class BaseloadBenchmarkGenerator < SchoolBenchmarkGenerator
+      def benchmark_school
+        return nil unless baseload_service.enough_data?
+        comparison = baseload_service.benchmark_baseload
+        return comparison.category
```
iant, line 7: Perhaps this can be simplified?
```
      def benchmark_school
        return unless baseload_service.enough_data?

        baseload_service.benchmark_baseload.category
      end
```

iant (approved):
Looks good to me @ldodds :+1: couple of comments

PR #2526, opened 2023-03-14
Integrate advice page benchmarks into content batch

ldodds:
Integrates advice page benchmark generation into the content batch.

* Adds new service to generate the benchmarks for all pages
* Adds the service to the content batch
* Specs for the above


iant (approved):
looks good to me @ldodds

PR #2527, opened 2023-03-14
Integrate benchmarks into advice page index

ldodds:
Adds the school advice page benchmarks into the index page

* Loads the benchmarks
* Where a benchmark is available, add a column with the appropriate label and background colour
* Further tweaks to page styling.


iant (approved):
:+1: looks good

PR #2528, opened 2023-03-15
3074 implement long term usage benchmark

ldodds:
Adds an advice page benchmark for the long term usage pages

* Move method from `AdviceBaseController` to `AdvicePage` so there's a consistent approach to checking if school has right fuel type for page for benchmarks and display of pages
* Move method from `BaseLongTermController` to `LongTermUsageService` so that benchmark comparison is reusable
* Changes long term usage advice controllers to request benchmark comparison from service
* Ensure school has right fuel type before generating an advice page benchmark
* Adds new `LongTermUsageBenchmarkGenerator` to produce benchmark for index page, and adds it to the factory method in base class
* Ensure `SchoolComparisonComponent` and advice page index are using same translation keys
* Tests for the above


iant (approved):
looks good to me @ldodds

PR #2529, opened 2023-03-15
Implement out of ours school comparison and advice page benchmark

ldodds:
This PR updates the out of hours gas and electricity insights pages to use the new school comparison component, rather than a table. It also adds an advice page benchmark generator for the advice page index.

Currently the out of hours benchmarking only produces values for exemplar and not benchmark schools. Improving that requires further analytics work.

So, for now this PR:

* Revises the `SchoolComparisonComponent` so that it tolerates a missing benchmark school comparison, and correctly renders the example and school values
* Extracts a common `BaseOutOfHoursController` from the existing electricity and gas controller. The controllers overlap for majority of their functionality
* Adds a `benchmark_usage` method to the new controller, to generate the `Schools::Comparison` required by the comparison component
* Removes the table from the insights pages for both fuel types, replacing it with the school comparison component
* Adds `OutOfHoursUsageBenchmarkGenerator` to generate the benchmark for the advice index page
* Tidies up translations, adds specs


urbanwide (approved):
Looks good @ldodds 👍

PR #2531, opened 2023-03-15
Implement storage heaters page

iant:


app/controllers/schools/advice/storage_heaters_controller.rb
```diff
@@ -1,17 +1,78 @@
...
+      end
+
+      def build_annual_usage_breakdown
+        ::Usage::AnnualUsageBreakdownService.new(
+          meter_collection: aggregate_school,
+          fuel_type: :storage_heater
+        ).usage_breakdown
+      end
+
+      def analysis_dates
```
ldodds, line 40: Do you need to override this method? The base class version looks like it creates the same ranges of dates. It relies on `analysis_start_date` and `analysis_end_date` methods returning the initial dates. Those fetch the right aggregate meter based on value of `advice_page_fuel_type`. So assuming the fuel type is same (there may be `storage_heater`/`storage_heaters` issue) you might not need this. Or may only need to override the `analysis_*_date` methods?

ldodds (approved):

ldodds (approved):

PR #2536, opened 2023-03-16
Implement thermostatic benchmark

ldodds:
Implements the school comparison component and advice page benchmark for the thermostatic control page

* Extract some common code into a new `ThermostaticAnalysisService` to be shared by controller and benchmark
* Add the benchmark comparison, which is a simple comparison of r2 values
* Update `Schools::Comparison` to accept a parameter to indicate whether high or low in ranges is good. (So far in all other cases we have been aiming for schools to, e.g. reduce consumption, but here they need to increase r2)
* Implement the benchmark for the advice page index

I've also slightly simplified the insights page as there was duplication between the "What is your baseload" and comparison section. So these have been combined. I've also aligned the page title better with the navbar.

### DIFF
--- a/app/classes/schools/comparison.rb
+++ b/app/classes/schools/comparison.rb
@@ -2,11 +2,12 @@ module Schools
   class Comparison
     attr_reader :school_value, :benchmark_value, :exemplar_value, :unit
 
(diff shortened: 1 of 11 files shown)


urbanwide (approved):
Looks good @ldodds - one very minor comment

PR #2538, opened 2023-03-16
Implement heating control benchmark

ldodds:
Implements the school comparison component and advice page benchmark for the heating control page.

We'll be benchmarking schools based on how often they have heating on during warm weather.

This PR:

* Extracts some code currently in a helper, moving it to the `HeatingControlService` and then adding methods to access the rating
* Revises the controller and templates to use the above when displaying notices about warm weather days
* Fixes a potential issue where we still try and run the seasonal analysis, even though there is limited data
* Updates the `HeatingControlService` to benchmark the school usage and adds the school comparison component to the insight page
* Adds a benchmark generator for the advice index page.

### DIFF
--- a/app/services/schools/advice/heating_control_service.rb
+++ b/app/services/schools/advice/heating_control_service.rb
@@ -3,6 +3,15 @@ module Schools
     class HeatingControlService
       include AnalysableMixin
 
+      #number of days heating on in warm weather => rating
+      WARM_WEATHER_DAYS_RATING = {
+        0..6     => :excellent,
+        6..11    => :good,
+        12..16   => :above_average,
+        17..24   => :poor,
+        25..365  => :very_poor
+      }.freeze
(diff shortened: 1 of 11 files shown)


app/services/schools/advice/heating_control_service.rb
```diff
@@ -3,6 +3,18 @@ module Advice
     class HeatingControlService
       include AnalysableMixin
 
+      #number of days heating on in warm weather => rating
+      WARM_WEATHER_DAYS_RATING = {
+        0..6     => :excellent,
+        6..11    => :good,
```
urbanwide, line 9: There looks like there is an overlapping range here. Should it be 0..5 => :excellent and 6..11 => :good (or similar?)

urbanwide (approved):
Looks good 👍

PR #2543, opened 2023-03-16
Improve benchmark labelling on advice index page

ldodds:
* Add a method and a helper to indicate whether we have an benchmark for a benchmark page
* Add a "not applicable" or "not available" message to advice index page to better indicate whether we do have a benchmark, or whether its not yet available
* Link the colour coded blocks to the relevant insights page
* Styling improvements.


iant (approved):
looks good to me @ldodds .  there's one bit of hard coded text that i think might need extracting out?

PR #2544, opened 2023-03-16
Store welsh version of benchmark results

ldodds:
This PR updates the code that stores and queries the benchmark variables so that it also stores a Welsh version. This is because a small number of the variables include text.

* Migration to remove the old, unused `:data` column from `BenchmarkResult` and add a new `result_cy` JSON column
* Update the services and code that popultes `BenchmarkResult` to remove use of the old column and generate the Welsh version of the results
* Change the `CollateBenchmarkResults` service to choose the right column based on the current locale
* Update the admin report so we can view the Welsh variables

I've revised most of the benchmarks and as far as I can see all of the text is already translated into Welsh.

**Update** - I've also removed an older admin report that listed all benchmark results, for all schools. It was very, very slow and likely would have broken the application if used in production now. There are already per-school reports so this overall report isn't required.


iant (approved):
:+1: Looks very good to me @ldodds

PR #2546, opened 2023-03-17
Log advice page and tab to GA

ldodds:
Updates out Google Analytics custom dimensions to include the advice page key and tab where available.

This should allow us to better report based on which pages users are visiting, rather than relying on page titles.


iant (approved):
:+1:

PR #2550, opened 2023-03-17
Redirect from benchmarks to advice pages

ldodds:
The benchmark pages use a custom route to link to the old analysis pages. That route uses the school url and class name. This is then used to find the latest version of that analysis page and the user is redirected to it.

We need to ensure these urls continue to work for the moment, but as the new advice pages have a predictable URL it shouldn't be needed longer term. 

For now this PR changes the controller to find the right AdvicePage instead and redirects to it. It uses the same feature flag we're using to hide the old analysis pages from the rest of the site.

### DIFF
--- a/app/controllers/concerns/analysis_pages.rb
+++ b/app/controllers/concerns/analysis_pages.rb
@@ -8,6 +8,11 @@ module AnalysisPages
     end
   end
 
+  def find_advice_page_of_class(analysis_class)
+    alert_type = AlertType.where("lower(class_name) = ?", analysis_class.downcase).first
+    return alert_type
+  end
+
   def setup_analysis_pages(analysis_pages)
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/analysis_pages.rb
```diff
@@ -8,6 +8,11 @@ def find_analysis_page_of_class(school, analysis_class)
     end
   end
 
+  def find_advice_page_of_class(analysis_class)
+    alert_type = AlertType.where("lower(class_name) = ?", analysis_class.downcase).first
+    return alert_type
```
iant, line 13: i think this can be simplified?
```
  def find_advice_page_of_class(analysis_class)
    AlertType.where("lower(class_name) = ?", analysis_class.downcase).first
  end
```

iant (approved):
looks good @ldodds   one comment

PR #2551, opened 2023-03-17
Improve layout of mail templates

ldodds:
This PR:

* fixes some layout issues in some of the application mailers. Buttons weren't floating right as templates were using wrong class, and not enough space between paragraphs
* adds some buttons to the activation and data enabled emails to better call out next action to users


iant (approved):
:+1:

PR #2556, opened 2023-03-17
Add partial months to meter costs table

ldodds:
Updates the meter costs component to indicate if there is only partial data for a month.


urbanwide (approved):
Looks good 👍

PR #2561, opened 2023-03-20
Enhance style of new pages

urbanwide:
See what you think @ldodds - I have made the benchmark list page groups into cards - not 100% convinced on the way it looks but easy enough to change. Also see what you think about the search filters. Also I'm not such a fan of the massive header on the front page but it does match the other pages...

Index / search filter page:
- [x] Make sure the index page heading matches that of activities
- [x] Boostrap-ify search filters
- [x] Remove extra spacing above title

Benchmark list page:
- [x] Each benchmark group to have a title, followed by the description, then two columns of benchmarks
- [x] Add button to change benchmark

Benchmark & Results pages
- [x] Style the summary filter so selected items are more prominent - consider using bootstrap badges for selected elements

Any other feedback…

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -52,5 +52,6 @@
 @import "activations";
 @import "toggler";
 @import "advice_page";
+@import "compare";
 
 @import "../../components/**/*"; // loaded here so bootstrap SASS variables are available
--- a/app/assets/stylesheets/compare.scss
+++ b/app/assets/stylesheets/compare.scss
@@ -1,0 +1,18 @@
+.compare .card-columns {
+  @include media-breakpoint-only(xl) {
+    column-count: 2;
+  }
+  @include media-breakpoint-only(lg) {
+    column-count: 2;
+  }
+  @include media-breakpoint-only(md) {
+    column-count: 2;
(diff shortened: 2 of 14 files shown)


ldodds:
I think this looks good. Some minor styling suggestions:

* remove the `button-sm` from the initial choose options page, I think the button can stand to be bigger there. (but keep small elsewhere)
* suggest adding colour to the badges for the school and school type. I tried with `badge-primary` and thought it looked ok
* the badge the school type and country are slightly different. Think it would look better if they were the same

I think the benchmark page looks better.

One thing you could consider trying is adding colours to each section. E.g. `$electric-dark` for the electricity group, etc. That would visually separate them a bit. The "Total Energy Use"  and the "Date limited comparisons" could remain grey.

I've just updated the navbar to use these colours to try and use consistent colours across the site.

Otherwise looks good.

urbanwide:
Thank you for the feedback Leigh, most useful!

Agree colours might be nice for the groups - I was thinking it would be nice to add icons too based on fuel type - but some sections are a mix of fuel types... and I can't find a way of extracting the fuel type, unless I've missed something?

Thanks! Really am going now. Chat Monday.

PR #2569, opened 2023-03-21
Translate month names, buttons, action counts

jhigman:
The month name, buttons and actions counts need to be translated on the interventions index page.

### DIFF
--- a/app/helpers/locale_helper.rb
+++ b/app/helpers/locale_helper.rb
@@ -40,6 +40,10 @@ module LocaleHelper
     I18n.t("date.other.#{day}", default: '')
   end
 
+  def t_month(month)
+    I18n.t('date.month_names')[month.to_i]
+  end
+
   def t_role(role)
     I18n.t("role.#{role}", default: '')
   end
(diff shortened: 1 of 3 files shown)


app/helpers/locale_helper.rb
```diff
@@ -40,6 +40,10 @@ def t_day(day)
     I18n.t("date.other.#{day}", default: '')
   end
 
+  def t_month(month)
```
iant, line 43: maybe we could raise and capture an error here if month number is out of range e.g. `fail unless month.between?(1,12)`?

iant (approved):
:+1: looks good to me @jhigman - 1 comment/suggestion on raising/capturing an error if the month no. is out of range.

app/helpers/locale_helper.rb
iant: (or not an integer)

jhigman: It should handle an out-of-bounds index by returning a nil, and not sure that we've generally logged that level of data error..?

PR #2572, opened 2023-03-21
Fix baseload page when some meters have limited data

ldodds:
The baseload page tries to display the seasonal and intraweek baseload variation for all meters if the school has >1 years of data.

But some of the individual meters may have <1 year of data.

This PR revises the `BaseloadService` and the templates to:

* Check whether there is enough data for both seasonal and intraweek variation
* If there is not enough data, then return a limited variation object that indicates this and provides a date when the analysis can be run for that meter
* Displays a custom row in the table for these meters to indicate to the user that there is limited data and, hopefully when the analysis can be run

### DIFF
--- a/app/helpers/advice_page_helper.rb
+++ b/app/helpers/advice_page_helper.rb
@@ -102,7 +102,7 @@ module AdvicePageHelper
   end
 
   def meters_by_estimated_saving(meters)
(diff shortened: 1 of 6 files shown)


iant (approved):
looks good to me @ldodds

PR #2573, opened 2023-03-21
Hot water page error when there is limited data

iant:


app/controllers/schools/advice/hot_water_controller.rb
```diff
@@ -22,7 +21,11 @@ def not_relevant?
       end
 
       def minimal_use_of_gas?
-        @gas_hot_water.investment_choices.existing_gas.efficiency > 1.0
```
ldodds, line 25: This still isn't correct I'm afraid.

This will result in a user seeing the "not relevant" page for the hot water analysis if they don't have enough data. That isn't what should happen.

The behaviour should be:

- if there's not enough data to run the analysis, they should get the "not enough data" page. (That's the base class behaviour, so you can delegate to the original method if the analysis can't be run)
- if there is enough data, then we do the additional relevance checks. If a school has a swimming pool, or has minimal use of gas, then they get sent to a specific "not relevant" page which has some additional text to explain why we're not analysing their data.

Can you fix up so it follows this please?

ldodds (changes requested):

ldodds:
@iant can you review this please?

PR #2576, opened 2023-03-22
Display alerts in sorted groups

jhigman:


app/views/schools/advice/index/_alert.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= component 'notice', classes: 'mt-4', status: status_for_alert_colour(alert_content.colour) do |c| %>
```
urbanwide, line 1: Just had a look and status_for_alert_colour(colour) would return :unknown here if the passed "colour" is nil. The notice component doesn't recognise :unknown as a valid status and will raise an error. Wonder if this is likely to happen?

urbanwide (commented):
Looks good - just a query about how the notice component is used...

app/views/schools/advice/index/_alert.html.erb
jhigman: Hm, I haven't changed that bit from what was there before (just refactored) - can I raise it as another ticket, to check that behaviour..?

urbanwide: Yes no probs...

urbanwide (approved):

PR #2580, opened 2023-03-22
Update values in electricity recent changes table so they show 3 significant figures

iant:


app/views/schools/advice/electricity_recent_changes/_your_recent_electricity_use_table_row.html.erb
```diff
@@ -1,9 +1,9 @@
 <% if combined_usage_metric.kwh %>
   <tr>
     <td><%= period %></td>
-    <td class="text-right"><%= format_unit(combined_usage_metric.kwh, :kwh) %></td>
-    <td class="text-right"><%= format_unit(combined_usage_metric.£, :£) %></td>
-    <td class="text-right"><%= format_unit(combined_usage_metric.co2, :co2) %></td>
+    <td class="text-right"><%= format_unit(combined_usage_metric.kwh, :kwh, true, :target) %></td>
```
iant, line 4: this and the values below use the `:target` "user_numeric_comprehension_level" as defined here https://github.com/Energy-Sparks/energy-sparks_analytics/blob/master/lib/dashboard/charting_and_reports/scalar/format_energy_unit.rb#L262

PR #2618, opened 2023-03-28
Make benchmark charts play better with bootstrap tabs

urbanwide:


iant (approved):
looks good to me @urbanwide

PR #2621, opened 2023-03-28
Implement services for range checking recent gas and electricity usage

jhigman:


iant (approved):
:+1: looks good to me @jhigman !

PR #2622, opened 2023-03-28
3066 implement holiday breakdown tables for advice pages

ldodds:
This adds a new "Holiday usage" section to the Electricity Out of Hours Page. Other pages to be updated in subsequent PRs.

* Bumps version of analytics to access new service
* Add new section to the page
* If there is > 14 months of data, then show a chart with breakdown of usage by day of week, holiday, etc
* Add table summarising usage for recent school holidays, with comparison with previous years holiday where available

Screenshot of portion of table:

Screenshot from 2023-03-28 18-29-53

### DIFF
--- a/app/controllers/schools/advice/advice_base_controller.rb
+++ b/app/controllers/schools/advice/advice_base_controller.rb
@@ -26,8 +26,6 @@ module Schools
       include DashboardAlerts
 
       rescue_from StandardError do |exception|
-        puts exception
-        puts exception.backtrace
         Rollbar.error(exception, advice_page: advice_page_key, school: @school.name, school_id: @school.id, tab: @tab)
         raise if Rails.env.development? || @advice_page.nil?
         render 'error', status: :internal_server_error
--- a/app/controllers/schools/advice/advice_base_controller.rb
+++ b/app/controllers/schools/advice/advice_base_controller.rb
@@ -26,6 +26,8 @@ module Schools
       include DashboardAlerts
 
       rescue_from StandardError do |exception|
+        puts exception
+        puts exception.backtrace
         Rollbar.error(exception, advice_page: advice_page_key, school: @school.name, school_id: @school.id, tab: @tab)
         raise if Rails.env.development? || @advice_page.nil?
         render 'error', status: :internal_server_error
--- a/app/controllers/schools/advice/base_out_of_hours_controller.rb
+++ b/app/controllers/schools/advice/base_out_of_hours_controller.rb
@@ -10,6 +10,7 @@ module Schools
 
       def analysis
         @annual_usage_breakdown = annual_usage_breakdown_service.usage_breakdown
+        @holiday_usage = holiday_usage_calculation_service.school_holiday_calendar_comparison
         @analysis_dates = analysis_dates
       end
 
@@ -27,8 +28,20 @@ module Schools
           end_date: end_date,
           one_years_data: one_years_data?(start_date, end_date),
           recent_data: recent_data?(end_date),
-          months_analysed: months_analysed(start_date, end_date)
+          months_of_data: months_between(start_date, end_date),
+          last_full_week_start_date: last_full_week_start_date(end_date),
+          last_full_week_end_date: last_full_week_end_date(end_date),
         )
+      end
+
+      #for charts that use the last full week
+      def last_full_week_start_date(end_date)
+        (end_date - 13.months).beginning_of_week - 1
+      end
+
+      #for charts that use the last full week
+      def last_full_week_end_date(end_date)
+        end_date.prev_week.end_of_week - 1
       end
 
       def benchmark_school(annual_usage_breakdown)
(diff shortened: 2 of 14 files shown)


app/controllers/schools/advice/advice_base_controller.rb
```diff
@@ -26,6 +26,8 @@ class AdviceBaseController < ApplicationController
       include DashboardAlerts
 
       rescue_from StandardError do |exception|
+        puts exception
+        puts exception.backtrace
```
iant, line 30: i think these might need removing?

app/helpers/advice_page_helper.rb
```diff
@@ -182,5 +182,24 @@ def dashboard_alerts_for_group(dashboard_alerts, group)
...
+    return false unless holiday_usage.previous_holiday_usage.present?
+    Time.zone.today > holiday.end_date
+  end
+
+  def within_school_period?(school_period)
+    @analysis_dates.end_date > school_period.start_date && @analysis_dates.end_date < school_period.end_date
+  end
+
+  def average_daily_usage(usage, school_period)
+    return usage.kwh / (school_period.end_date - school_period.start_date).to_i
```
iant, line 202: return not needed here.  Also, just to double check you didn't mean to use `to_f` here?

app/views/schools/advice/electricity_out_of_hours/_current_holiday_usage_row.html.erb
```diff
@@ -0,0 +1,27 @@
...
+      <%= format_unit(average_daily_usage(holiday_usage[holiday].usage, holiday), :kwh, true, :target) %>
+    </td>
+    <td class="text-right">
+      <%= format_unit(holiday_usage[holiday].usage.£, :£, true, :target) %>
+    </td>
+    <td class="text-right">
+      <%= format_unit(holiday_usage[holiday].usage.co2, :co2, true, :target) %>
+    </td>
+  <% else %>
+    <td colspan="3" class="text-center old-data">
```
iant, line 23: just to double check... the bit after `<% if holiday_usage[holiday].usage.present? %>` has 4 table cells so does this need to be `colspan="4"` ?

app/views/schools/advice/electricity_out_of_hours/_holiday_usage_table.html.erb
```diff
@@ -0,0 +1,25 @@
...
+      <%= render 'previous_holiday_usage_row', holiday: holiday, holiday_usage: holiday_usage %>
+      <%= render 'current_holiday_usage_row', holiday: holiday, holiday_usage: holiday_usage %>
+      <%= render 'holiday_comparison_row', holiday: holiday, holiday_usage: holiday_usage %>
+    <% end %>
+  </tbody>
+</table>
+<% if within_school_period?(periods.last) %>
+  <div class="text-right advice-table-caption">
+    <sup>*</sup> <%= t('advice_pages.tables.notice.partial_holiday')%>
+  </div>
```
iant, line 24: not something for this pr but we use this pattern in quite a few places so feels like it could be extracted out to a partial or component.

app/views/schools/advice/electricity_out_of_hours/_previous_holiday_usage_row.html.erb
```diff
@@ -0,0 +1,27 @@
...
+    </td>
+
+    <td class="text-right">
+      <%= format_unit(holiday_usage[holiday].previous_holiday_usage.£, :£, true, :target) %>
+    </td>
+    <td class="text-right">
+      <%= format_unit(holiday_usage[holiday].previous_holiday_usage.co2, :co2, true, :target) %>
+    </td>
+  <% else %>
+    <td colspan="3" class="text-center old-data">
```
iant, line 23: just to double check here too if this should be colspan = 4

app/controllers/schools/advice/base_out_of_hours_controller.rb
```diff
@@ -27,10 +28,22 @@ def analysis_dates
...
           one_years_data: one_years_data?(start_date, end_date),
           recent_data: recent_data?(end_date),
-          months_analysed: months_analysed(start_date, end_date)
+          months_of_data: months_between(start_date, end_date),
+          last_full_week_start_date: last_full_week_start_date(end_date),
+          last_full_week_end_date: last_full_week_end_date(end_date),
         )
       end
 
+      #for charts that use the last full week
```
iant, line 37: maybe an additional comment here & below to say for all(?) charts we treat the beginning of the week as a Sunday and end of the week as a Saturday (hence the -1)

iant (approved):
looks good to me @ldodds a few v minor comments/questions/suggestions

app/helpers/advice_page_helper.rb
ldodds: You're right the to_i is not really needed.

app/views/schools/advice/electricity_out_of_hours/_holiday_usage_table.html.erb
ldodds: Yes, I think you're right

PR #2626, opened 2023-03-29
Implement holiday breakdown tables for storage heater page

ldodds:
Adds holiday use table to storage heater page

* Updates controller to load the necessary analysis
* Adds section to the page with necessary translations
* Updated specs to check for presence of all charts

There's no new chart for the holiday use


iant (approved):
looks good @ldodds

PR #2627, opened 2023-03-29
Add holiday use to storage heater page

ldodds:
Adds holiday use table to storage heater page

- Updates controller to load the necessary analysis
- Adds section to the page with necessary translations
- Updated specs to check for presence of all charts

There's no new chart for the holiday use

### DIFF
--- a/app/views/schools/advice/storage_heaters/_holiday_usage_table.html.erb
+++ b/app/views/schools/advice/storage_heaters/_holiday_usage_table.html.erb
@@ -4,7 +4,7 @@
       <th>Holiday</th>
       <th>Period</th>
       <th class="text-right"><%= t('common.table.columns.use_kwh') %></th>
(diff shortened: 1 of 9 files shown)


app/views/schools/advice/storage_heaters/_holiday_usage_table.html.erb
```diff
@@ -0,0 +1,25 @@
+<table class="table advice-table">
+  <thead>
+    <tr>
+      <th>Holiday</th>
+      <th>Period</th>
```
iant, line 5: I think "Holiday" & "Period" may need extracting to translation yml files?

ldodds: OK, these are in the electricity and gas PRs too. I'll do a separate PR to fix that in all three

PR #2630, opened 2023-03-29
Restore missing unit, ensure error page is translated

ldodds:
Transifex synchronisation removed a unit, which is causing thermostatic insights pages to fail in Welsh.

This also expose an issue with the advice error page which isn't rendering with a locale. So have wrapped that in calls to ensure its translated.


iant (approved):
:+1:

PR #2633, opened 2023-03-29
Add note about temp compensation

ldodds:
Adds an extra note to the gas out of hours holiday usage section.


urbanwide (approved):
Looks good 👍

PR #2638, opened 2023-03-30
Solar analysis page issue when school has limited electricity data

ldodds:
If a school doesn't have solar panels, then we need >1 of electricity to calculate benefits

If a school does have panels, then we can calculate the existing benefits with limited data. At the moment we're excluding schools from accessing the solar page when they should.

This PR fixes up the application enough data checks. A follow-up PR should change the service in the analytics.


iant (approved):
:+1:

PR #2639, opened 2023-03-30
Tweak GA events

ldodds:
Am not sure the GA events are being recorded correctly for the advice pages.

This tweaks the event callbacks to match what we're doing elsewhere. Can review changes next week.


iant (approved):
:+1:

PR #2641, opened 2023-03-30
Removed spurious output of latest progress value

jhigman:
This PR removes a spurious print of the latest progress value in the target progress page..


iant (approved):
:+1:

PR #2642, opened 2023-03-30
Baseload multiple meter check

ldodds:
The baseload service wasn't correctly counting meters. It was including inactive meters when it shouldn't


iant (approved):
:+1:

PR #2645, opened 2023-03-31
Report chart errors to Rollbar

iant:


app/models/chart_data.rb
```diff
@@ -38,6 +41,20 @@ def data
...
+    chart_manager.run_chart(
+      transformed_chart_config,       # chart_config
+      transformed_chart_type,         # chart_param
+      true,                           # resolve_inheritance
+      nil,                            # override_config
+      @reraise_exception,             # reraise_exception
+      provide_advice: @provide_advice # provide_advice
+    )
+  rescue => e
+    Rollbar.error(e) if @reraise_exception
```
ldodds, line 54: Can you add more context to the Rollbar error? See the call in the exception handler below. Would be useful to add all the same attributes to help with debugging. 

Note there's a bug in that call, it should be using @school.name not @aggregated_school.name.

ldodds: Could you also add Rails logger call here, to help with debugging in development?

ldodds (changes requested):

app/models/chart_data.rb
iant: :+1: have fixed the school name bug in the below call too.

ldodds (approved):

PR #2655, opened 2023-03-31
Fix javascript error on baseload chart

ldodds:
For some charts we add annotations that indicate when the school took action. At the moment this is limited to the dashboard charts and one of the baseload charts.

The javascript is expecting that if there are annotations then the chart point labels (a date for these charts) should match the label given in the JSON produced for each annotation.

The date format for the baseload chart seems to have changed. But the `Charts::Annotate` class was producing a different format. The mismatched caused an error.

This PR:

* Changes the date format in the `Annotate` class to match the baseload chart
* Adds a check in the JS for whether we have a point before trying to create an annotation


iant (approved):
:+1:

PR #2660, opened 2023-04-04
Dont translate series labels for scatter plots

ldodds:
Avoids trying to translate the series names on the scatter charts. This currently breaks the charts as the float values end up as strings.

The series names aren't actually displayed to users, so it isn't an issue to not translate the dates.


iant (approved):
:+1:

PR #2664, opened 2023-04-05
Allow admins to assign meter status to individual meters (3)

iant:
- [x] Allow School Group to have separate default meter status for electricity, gas and solar meters
- [x] When defaulting the meter status for a Meter, use the right default from the group
- [x] Disallow deleting a meter status if it is in use
- [x] Add a tooltip if the meter status is a group default

### DIFF
--- a/app/models/meter.rb
+++ b/app/models/meter.rb
@@ -103,7 +103,7 @@ class Meter < ApplicationRecord
   end
 
   def admin_meter_status_label
-    admin_meter_status&.label || school&.school_group&.admin_meter_status&.label || ''
+    admin_meter_status&.label || school&.school_group&.send(:"admin_meter_status_#{fuel_type}")&.label || ''
   end
 
   def fuel_type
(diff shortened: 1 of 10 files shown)


app/models/meter.rb
```diff
@@ -103,7 +103,7 @@ def school_name
   end
 
   def admin_meter_status_label
-    admin_meter_status&.label || school&.school_group&.admin_meter_status&.label || ''
+    admin_meter_status&.label || school&.school_group&.send(:"admin_meter_status_#{fuel_type}")&.label || ''
```
ldodds, line 106: Valid meter types are: 

```
enum meter_type: [:electricity, :gas, :solar_pv, :exported_solar_pv]
```

I think the exported solar pv meters should have same default type as solar pv. Can you revise that?

ldodds (changes requested):

ldodds (approved):

PR #2665, opened 2023-04-05
Translate chart period

ldodds:
Updates the version of the analytics to incorporate translations for chart periods.

Removes previous approach implemented in the application which only supported a couple of period types.

Manually added Welsh translations for day, week, year, up to a year so the current translations are used until the full set is done.

There's a small functional change here which is a side-effect of the code changes: if there's no Welsh version of some timescales it will default to "period". I think that's reasonable for the moment and is what we use elsewhere.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,7 @@ gem 'pg_search'
 gem 'calculate_in_group'
 
 # Dashboard analytics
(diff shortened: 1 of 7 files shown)


iant (approved):
:+1:

PR #2677, opened 2023-04-13
Treating '-' as missing data

ldodds:
The Smartest Energy CSV data includes rows which have:

- all 48 half-hourly readings as "-"
- portions of the day as dashes, e.g. 1-2 hours in the morning then dashes

These are currently being passed through for storing. When processed the array of dashes is processed via (`readings.map(&:to_f)`) which converts the dashes to zeroes. Which means it treats the data as zero readings.

When the real data appears the data is backfilled and replaced, but the behaviour isn't correct.

This PR makes a system wide change to treat "-" characters as missing values, which means rows containing that data will be rejected. This may lead to more rejected data, but is better behaviour overall.

The PR also includes an AfterParty task to clean up some existing missing data

The task:

- removes `AmrDataFeedReadings` where the `readings` are all dashes.
- removes `AmrDataFeedReadings` where there is more than one dash in the array

This will remove about a 1000 rows in the live system, leaving the data to be substituted on revalidation.

### DIFF
--- a/app/models/amr_reading_data.rb
+++ b/app/models/amr_reading_data.rb
@@ -103,7 +103,7 @@ class AmrReadingData
   end
 
   def missing_readings?(readings)
-    readings.compact.count(&:present?) < (48 - @missing_reading_threshold)
+    readings.compact.count {|hh| hh.present? && hh != '-'} < (48 - @missing_reading_threshold)
   end
 
   def valid_reading_date?(reading_date)
--- a/app/models/amr_reading_data.rb
+++ b/app/models/amr_reading_data.rb
@@ -103,7 +103,7 @@ class AmrReadingData
   end
 
   def missing_readings?(readings)
-    readings.compact.count {|hh| hh.present? && hh != '-'} < (48 - @missing_reading_threshold)
+    readings.compact.count {|reading| reading.present? && reading != '-'} < (48 - @missing_reading_threshold)
   end
 
   def valid_reading_date?(reading_date)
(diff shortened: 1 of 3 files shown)


spec/models/amr_reading_data_spec.rb
```diff
@@ -152,6 +152,22 @@
         expect(amr_reading.warnings.first[:warnings]).to include(:missing_readings)
       end
 
+      it "with missing readings (as \"-\")" do
+        readings = amr_reading_data[:reading_data].first[:readings]
+        readings[readings.size - 1] = "-"
+        puts readings.inspect
```
urbanwide, line 158: Did you mean to leave this puts in ?

urbanwide (approved):
Looks good @ldodds 👍

app/models/amr_reading_data.rb
```diff
@@ -103,7 +103,7 @@ def inconsistent_reading_date_format?(reading_date)
   end
 
   def missing_readings?(readings)
-    readings.compact.count(&:present?) < (48 - @missing_reading_threshold)
+    readings.compact.count {|hh| hh.present? && hh != '-'} < (48 - @missing_reading_threshold)
```
iant, line 106: very very minor readability nitpick - maybe replace `hh` here with `reading`?
`readings.compact.count { |reading| reading.present? && reading != '-' } < (48 - @missing_reading_threshold)`

iant (approved):

PR #2686, opened 2023-04-18
Optimising import notifier

ldodds:
The import notifier has a number of slow queries. This PR tries to optimise some of them by reworking the queries and/or pushing more checks into the database.

# `meters_running_behind`

This should look through all the data feed configs, then find any meters that have readings that have been imported using that config.

If we find any, then we check to see if the latest validated reading for the meter is older than the number of days in the feed config.

Originally this was joining to the `amr_data_feed_import_logs` table, then to the `amr_data_feed_readings` table and doing a distinct. 

Instead we query directly on the meters table, for active meters, joining it to the `amr_data_feed_readings` table, which has an `amr_data_feed_config_id` and then to the `amr_validated_readings` to do the check on the date in the database.


app/services/import_notifier.rb
```diff
@@ -6,33 +6,39 @@ def initialize(description: nil)
...
-      all_log_meters = import_logs.map do |log|
-        log.amr_data_feed_readings.select {|reading| reading.readings.blank? || reading.readings.all?(&:blank?)}.map(&:meter).compact
-      end
-      all_log_meters.flatten.uniq
-    end
+    Meter.active
+    .joins(:amr_data_feed_readings)
+    .where("amr_data_feed_readings.readings = ARRAY[?]", Array.new(48, '')) #where readings is empty string
+    .joins("INNER JOIN amr_data_feed_import_logs on amr_data_feed_readings.amr_data_feed_import_log_id = amr_data_feed_import_logs.id") #manually join to import logs
+    .where('import_time BETWEEN ? AND ?', from, to) #limit to period
```
iant, line 27: maybe include some named placeholders here? e.g.
.where('import_time BETWEEN :from AND :to', from: from, to: to)

iant (approved):

PR #2687, opened 2023-04-18
Allow a meter to be deleted if there are issues attached

iant:


app/services/meter_management.rb
```diff
@@ -44,7 +44,7 @@ def process_mpan_mpnr_change!
   def delete_meter!
     @meter.transaction do
       AggregateSchoolService.new(@meter.school).invalidate_cache
-      @meter.delete
+      @meter.destroy
```
urbanwide, line 47: Just pondering the impacts of changing this - I guess this means that previously callbacks weren't being run when a meter was deleted (and now associated amr_validated_readings, tariff_prices, tariff_standing_charges and issue_meters will also be removed when a meter is removed). Assume this is ok and also does this mean there are loads of orphaned records in these tables from deleted meters?

iant: Very good point that previously callbacks weren't being run and there's potentially lots of orphaned records.  I actually didn't think to check but will do that now and do a follow up pr if there's any db clean up required.

also just an fyi for why i'd removed the callback from amr_data_feed_readings - they need to be kept around as often ES will be given data before schools are onboarded, so helpful to have the data there. also, if the meter was added to wrong school then this also allows the setup to be fixed without deleting associated meter data.   I'll add that as a comment in the code (and add a test) to prevent the callback being accidentally added back in.

ldodds (approved):

PR #2690, opened 2023-04-19
Add a date limit to the amr data feed reading lookup

ldodds:
Only check to see what configs we've used to load data for a meter in the last 12 months, not all time.

This significantly improves the "meters running behind" check and should involve less paging through data on the database.


urbanwide (approved):
Looks good @ldodds 👍

PR #2695, opened 2023-04-19
Push entire "meters running behind" query into database

ldodds:
Currently we loop over every data feed config and run the same query so that we can check its warning days, if it has one.

For some configs this is very quick, others very slow.

This changes the code to just join against the `amr_data_feed_configs` table, where the `import_warning_days` is not null, then check whether the latest meter data is older than the minimum value of the warning days.

This means we do one query rather than several. Overall I think this version is faster!

### DIFF
--- a/app/services/import_notifier.rb
+++ b/app/services/import_notifier.rb
@@ -4,55 +4,44 @@ class ImportNotifier
   end
 
   def meters_running_behind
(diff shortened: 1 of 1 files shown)


iant (approved):
👍 all looks good to me @ldodds

PR #2697, opened 2023-04-19
Rework school group meters query

ldodds:
Reworks the school group meter report to query directly for meters associated with a school group, so the list of meters, start and end dates and count of zero readings can be done entirely in the database.

This seems faster than the loop over schools and meters.

We might still need to optimise the checks for gappy readings.


ldodds:
Unfortunately, doesn't seem to have helped on the test server, report still won't run, so more work needed.

urbanwide:
Oh that is a shame :-( Was hopeful then!

PR #2719, opened 2023-04-26
Parallelize weekly alert run (subscriptions)

iant:
Every Weds ES reruns alerts:generate_subscriptions. This loops through all schools and invokes the Alerts::GenerateSubscriptions service for each school.  The process is getting slower (and CPU intensive) as we add more schools. Further, because the process starts via cron, it may begin before we’ve finished the daily regeneration of schools which creates the potential that the email contents don’t have the very latest alert content from schools.

In order to address these issues, this PR:

- [x] Adds a new subscription_frequency method to schools that returns the correct subscription frequency depending on if there is an upcoming holiday, or not.
- [x] Creates a new job that will run the Alerts::GenerateSubscriptions service for a school and a subscription_frequency
- [x] Ensures those jobs are submitted to the regeneration queue like the daily regeneration jobs, so they’ll be processed by the same pool of workers
- [x] Sets the DailyRegenerationJob priority to 10 and the new subscription job a priority of 5, so that the regeneration jobs will always be processed first
- [x] Revises the generate_subscriptions rake task to enqueue a job for each school
- [x] Revises the timing of the generate_subscriptions job so that it starts earlier - at 6.30am instead of 7.45 am.

### DIFF
--- a/.ebextensions/cronjob.config
+++ b/.ebextensions/cronjob.config
@@ -161,7 +161,7 @@ files:
         owner: root
         group: root
         content: |
-            45 7 * * Wed root /bin/bash -l -c 'cd /var/app/current && RAILS_ENV=production bin/run_as_webapp bundle exec rake alerts:generate_subscriptions >> log/jobs.log 2>&1'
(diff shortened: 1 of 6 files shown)


urbanwide (approved):
Looks good @iant 👍

PR #2724, opened 2023-04-27
Replace online school meter report with email

urbanwide:
Left to do:

- [x] Write some spec :-)
- [x] Remove links to web report and affected spec


iant (approved):
Looks good to me @urbanwide :+1:

PR #2726, opened 2023-04-27
Rework how Import Notifier identifies warning days

iant:
This PR:
- [x] Adds a new SiteSetting for default_import_warning_days this should be 10 days by default
- [x] Updates the forms at /admin/settings so that an admin can revise it
- [x] Updates the DataSource model and forms to have an import_warning_days field, which defaults to nil
- [x] Revises the import notifier so that instead of querying against the feed readings, it uses the import warning days for the data source for the meter (if there is one and its not empty) otherwise use the default system wide setting


app/services/import_notifier.rb
```diff
@@ -4,15 +4,41 @@ def initialize(description: nil)
...
-    .joins(:amr_validated_readings)
-    .group('meters.id')
-    .having("MAX(amr_validated_readings.reading_date) < NOW() - MIN(amr_data_feed_configs.import_warning_days) * '1 day'::interval")
-    meters.sort_by {|m| [m.school.area_name, m.meter_type, m.school_name, m.mpan_mprn]}
+    find_meters_running_behind.sort_by do |meter|
+      [
+        meter&.school&.area_name || '',
+        meter&.meter_type || '',
+        meter&.school_name || '',
+        meter&.mpan_mprn || 0
```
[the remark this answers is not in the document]
ldodds: Yes all the columns are required.

Agree it would be better to sort in the database, but we can add that separately. What's the reason for doing the defaults in the (e.g. `|| 0`) as the current version doesn't use those and all those fields are required I think?

app/services/import_notifier.rb
```diff
@@ -4,15 +4,41 @@ def initialize(description: nil)
...
+      ]
+    end
+  end
+
+  def find_meters_running_behind
```
ldodds, line 17: This doesn't look quite right to me.

Firstly, if we're finding the warning period using the data source (or a global default) then we don't need to join to `amr_data_feed_readings`. Avoiding that join is the main motivator for this optimisation. So that should be removed.

Secondly, do you need the `MIN` in the `having` clause? In the current version I use MIN because there might be multiple warning periods, so I use the smallest. But in this version there will either be the warning period for the meters data source OR the global default as COALESCE returns the first non-null value. 

Finally, as we're later sorting the results using a `sort_by`, we probably don't need the `.order`?

ldodds (changes requested):

app/services/import_notifier.rb
iant: :+1: I've pushed some updates to the query

iant: `sort_by` fails on nil and (at least in my tests) the `meter&.school&.area_name` could return nil.  As all other fields are required i've updated the area_name method to return an empty string if no school group is present

ldodds (approved):
Looks good, but I've marked this as blocked so we can give the team notice of the changes

ldodds (changes requested):
Can we add an extra spec for when a meter doesn't not have a Data Source, to ensure the system fallback is used?

ldodds (approved):

iant:
Removing blocked label as all ok to merge now

PR #2727, opened 2023-04-28
Create admin report to report on PROB data

iant:


app/views/admin/index.html.erb
```diff
@@ -48,6 +48,7 @@
       <li><%= link_to "Weather Stations", admin_weather_stations_path %></li>
       <li><%= link_to "Data Sources", admin_data_sources_path %></li>
       <li><%= link_to "Procurement Routes", admin_procurement_routes_path %></li>
+      <li><%= link_to "PROB data report", admin_prob_data_reports_path %></li>
```
ldodds, line 51: Can you add this to `/admin/reports/index.html.erb` instead of here, under the "Data & Tariffs" section?

ldodds (changes requested):

app/views/admin/index.html.erb
iant: :+1: done

ldodds (approved):

PR #2731, opened 2023-05-03
Transifex updates and pull (Wed 3rd May 2023)

iant:
This pr includes an update to the tx config with all new content pushed to transifex (see https://app.transifex.com/energy-sparks/energy-sparks/settings/timeline/) and pulling down any new translated content as of 3rd May 2023.

### DIFF
--- a/config/locales/cy/analytics/charts/chart_configuration.yml
+++ b/config/locales/cy/analytics/charts/chart_configuration.yml
@@ -47,7 +47,7 @@ cy:
         - Cymhariaeth o ddefnydd trydan y 2 wythnos ysgol ddiwethaf
     alert_1_year_baseload:
       title:
-        - Llwyth sylfaen kW - y llynedd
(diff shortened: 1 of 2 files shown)


urbanwide (approved):
Looks good @iant 👍

config/locales/cy/analytics/charts/chart_configuration.yml
```diff
@@ -1006,7 +1000,6 @@ cy:
       title:
         - Thermostatig (Penwythnos y Gaeaf)
     y_axis_label_name:
-      r2: ''
```
ldodds, line 1009: one minor thing here, I think this might cause a problem. Transifex strips empty strings, but sometimes we need them in the YAML.

ldodds (changes requested):

ldodds (approved):

PR #2733, opened 2023-05-03
Update subjects in activities

iant:


iant:
putting back in to draft.  need to remove all Design and Technology references from activity types before deleting

urbanwide (approved):
Nice one @iant 👍

PR #2734, opened 2023-05-04
Recent changes page updates

iant:
This pr adds some updates to the Gas and electricity recent changes advice pages:

the table should show up to 2 weeks of data which should be 2 most recent weeks (Sun-Sat) not most recent school weeks.

- [x] Rationalise the recent changes services in the application, to add a fuel type parameter
- [x] Remove the RecentUsagePeriodCalculationService in favour of calling aggregate_meter.amr_data.kwh_date_range method directly in the application service 
- [x] If a school has <1 full week of data, then page is not available
- [x] If a school has >=1 full week of data, then table will display with single row
- [x] If a school has >=2 full weeks of data, then table will have 3 rows (one per week, plus change row)

We also need to remove the RecentUsageComparisonService and RecentUsagePeriodCalculationService classes as they are no longer used - this is addressed in this PR https://github.com/Energy-Sparks/energy-sparks_analytics/pull/533


ldodds (approved):
Changes look good, lets get this merged and on to test for further review.

PR #2735, opened 2023-05-04
Make it more obvious when events are sold out

urbanwide:
I tried very hard to find a count of tickets remaining to also add to the page but could not find an easy way... so I stuck with the original plan of just showing "sold out" or "tickets available".


iant (approved):
Looks good to me @urbanwide

PR #2739, opened 2023-05-05
Fix baseload chart error when meter has no readings

urbanwide:
https://trello.com/c/VMvOODVX/3277-baseload-chart-error-when-meter-has-no-readings


iant (approved):
looks good to me @urbanwide

PR #2742, opened 2023-05-09
Fix baseload analysis divide by zero error

iant:


iant:
Screenshot 2023-05-10 at 08 26 20

PR #2747, opened 2023-05-10
Fix issues with tariffs and costs advice page tables not adding up.

iant:
Screenshot 2023-05-10 at 15 48 28


urbanwide (approved):
Looks good @iant 👍

PR #2756, opened 2023-05-16
Fix meter report email modal for when there are multiple per page

urbanwide:


iant (approved):

PR #2757, opened 2023-05-16
Update electricity longterm trend description text

iant:
Remove reference to non-existent table.


urbanwide (approved):
Looks good @iant 👍

PR #2760, opened 2023-05-17
Rework n3rgy tariff import

iant:


db/schema.rb
```diff
@@ -852,21 +852,21 @@
     t.index ["replaced_by_id"], name: "index_global_meter_attributes_on_replaced_by_id"
   end
 
-  create_table "good_job_processes", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+  create_table "good_job_processes", id: :uuid, default: -> { "public.gen_random_uuid()" }, force: :cascade do |t|
```
ldodds, line 855: Need to remove these changes

ldodds (changes requested):

db/schema.rb
iant: :+1:

PR #2764, opened 2023-05-18
Add cron entry for sending admin issue reports

urbanwide:


iant (approved):

urbanwide:
@ldodds would you mind also checking the times of this cron entry please? Just want to make sure it doesn't clash with anything else.

PR #2772, opened 2023-05-22
Configure new Npower/EON format

ldodds:
New feed format for a file format received from Npower/EON.


urbanwide (approved):
Looks good @ldodds 👍

PR #2773, opened 2023-05-22
Configure new My SEFE Portal config

ldodds:
* Fixes a minor bug in the data feed translator
* Adds new config for the My SEFE portal


urbanwide (approved):
Looks good @ldodds 👍

PR #2775, opened 2023-05-23
Fix school group issues report where a school was appearing that shouldn't have been

urbanwide:


iant (approved):

PR #2790, opened 2023-05-30
Add school group feature flag to switch between current and new templates

iant:


app/views/school_groups/current_scores.html.erb
```diff
@@ -0,0 +1,3 @@
+<% content_for :page_title, @school_group.name %>
```
ldodds, line 1: The page titles should also reflect the action, e.g. " XYZ group | Current Scores"

PR #2791, opened 2023-05-30
Hide school group dashboard for private school groups

iant:


ldodds (changes requested):
Can you add a system spec to exercise the behaviour for public / non-public schools with a range of different user types?

ldodds (approved):
This looks OK. Ideally we'd also have some tests for the `group_admin` role too, but looks like the basics are in place here.

PR #2796, opened 2023-05-31
Change polling for smart meters

ldodds:
When we setup new meters, we check to see if there are known to n3rgy. This allows us to detect smart meters and then get access to data via their API.

There's a rake task that runs daily to check for meters, but this never really does anything as all meters are now checked on creation.

However in some cases meters are upgraded/replaced. This means we need to regularly recheck meters, otherwise we'll miss that they've been setup. The rake task also only checked active meters, which means we might miss some that are disabled pending further setup.

This PR changes the existing Rake task to:

* check any meter that hasn't yet been checked
* check any gas or electricity meters that haven't been flagged as a smart meter (`dcc_meter == false`) and we last checked > 2 months ago


iant (approved):
:+1: looks good to me @ldodds

PR #2797, opened 2023-05-31
Ignore invalid rows in single read converter

ldodds:
It seems like Excel sometimes adds extra blank rows to the end of exported CSV files. This is causing some manually converted files to fail to load as the system is not expecting an empty row.

Improve tolerance of the SingleReadConverter to skip rows without an mpan or date. Normally these would be caught and rejected in the next step of processing.


iant (approved):
:+1:

PR #2798, opened 2023-05-31
Add Recent usage table to new school dashboard

iant:


app/models/school.rb
```diff
@@ -584,6 +584,10 @@ def subscription_frequency
     end
   end
 
+  def recent_usage
```
ldodds, line 587: Just noting that we have a service that does mostly this already: `ManagementTableService`. Wondering if it'd be best to align, so there's one way to load the data?

app/views/school_groups/recent_usage.html.erb
```diff
@@ -1 +1,49 @@
 <%= render 'enhanced_header' %>
+
+<table width="100%" class="mt-3 table table-borderless table-sorted school-groups-recent-usage-table">
```
ldodds, line 3: Can you add the advice-table class so we end up with a table that matches the styling on the advice pages/school dashboard?

app/views/school_groups/recent_usage.html.erb
```diff
@@ -1 +1,49 @@
...
+  <thead>
+    <tr>
+      <th colspan="2"><%= t('common.electricity') %></th>
```
ldodds, line 6: Can you try adding the fuel type icons, as we have on the table on the school dashboard?

app/views/school_groups/recent_usage.html.erb
```diff
@@ -1 +1,49 @@
...
+      <th colspan="2"><%= t('common.gas') %></th>
+      <th colspan="2"><%= t('common.storage_heaters') %></th>
```
ldodds, line 8: I recognise this wasn't specified anywhere, so happy if you want to do in a follow-up PR, but: very few schools have storage heaters (as far as we know) so the storage heater columns should only appear if at least one school in a group has storage heaters. this avoids having empty columns for (most?) groups.

ldodds (changes requested):

app/models/school.rb
iant: yep no probs.  I've updated to use `ManagementTableService`

ldodds (approved):

PR #2799, opened 2023-05-31
Priority actions for school groups

ldodds:
Initial spike of creating a summary of priority actions for school groups.

Things to ensure we've tested and resolved:

* [x] schools for a group may not yet have any priority actions
* [x] the alert type rating content version may have changed since the actions were calculated for a school, so we need to ensure we're using the current content for all schools, not the version they were generated with. This will likely be rare, but could happen
* [x] the Management Priorities are associated with an AlertTypeRating, so we need to aggregate up to the AlertType level, as otherwise we will have separate values for e.g. "Average baseload" and "High baseload", which will mean the table has duplicate rows and the totals will be off
* [x] ensure totals are sorted by price by default
* [x] ensure table sorts correctly
* [x] add popup with list of schools
* [x] add system spec
* [x] extract text to YML
* [x] sort modal entries by school name by defaults
* [x] add some explanatory text to indicate how we calculated these values, including some references to their tariffs. The savings are large so may need to be justified
* [x] further refactor and optimise service

### DIFF
--- a/app/views/school_groups/priority_actions.html.erb
+++ b/app/views/school_groups/priority_actions.html.erb
@@ -1,5 +1,10 @@
 <%= render 'enhanced_header' %>
 
+<%= component 'notice', status: :neutral, classes: 'mt-2 mb-4' do |c| %>
+  <%= t('school_groups.priority_actions.intro_html') %>
+<% end %>
+
+
 <table class="table table-borderless table-sorted advice-table advice-priority-table">
   <thead>
     <tr>
--- a/app/views/school_groups/priority_actions.html.erb
+++ b/app/views/school_groups/priority_actions.html.erb
@@ -28,12 +28,12 @@
             <%= savings.schools.length %>
           </h4>
         </td>
-        <td>
+        <td data-order="<%= savings.average_one_year_saving_gbp %>">
           <h4>
             <%= format_unit(savings.average_one_year_saving_gbp, :£) %>
           </h4>
         </td>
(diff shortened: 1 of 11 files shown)


app/views/school_groups/priority_actions.html.erb
```diff
@@ -1 +1,59 @@
...
+            <%= fa_icon alert_type_icon(alert_type_rating.alert_type,'fa-2x') %>
+          </span>
+        </td>
+        <td>
+          <h4>
+            <a href="" data-toggle="modal" data-target="#action-<%= alert_type_rating.id %>">
+              <%= alert_type_rating.current_content.management_priorities_title %>
+            </a>
+          </h4>
+          <%= render(FootnoteModalComponent.new(title: alert_type_rating.current_content.management_priorities_title.to_plain_text, modal_id: "action-#{alert_type_rating.id}", modal_dialog_classes: 'modal-xl modal-dialog-centered')) do |component| %>
```
urbanwide, line 32: Could this use the component helper?

urbanwide (approved):
Looks good @ldodds - just a very minor comment about using the component helper 👍 

Also, when I ran the code locally, for some school groups (e.g GSDT) I get an error, but it could be because I have old data: 
image

app/views/school_groups/priority_actions.html.erb
ldodds: Ah yes, good idea

PR #2800, opened 2023-06-01
Add a scoreboard tab to the school group dashboard

iant:


app/models/concerns/scorable.rb
```diff
@@ -0,0 +1,13 @@
+module Scorable
+  extend ActiveSupport::Concern
+
+  def scored(recent_boundary: 1.month.ago)
```
ldodds, line 4: This needs to be the scores for the current academic year, not the scores from the last month.

ldodds (changes requested):

app/views/school_groups/current_scores.html.erb
```diff
@@ -1 +1,37 @@
 <%= render 'enhanced_header' %>
```
ldodds, line 1: Can you add a spec please?

ldodds (changes requested):

app/models/concerns/scorable.rb
iant: :+1:

PR #2803, opened 2023-06-05
Make recent usage tab tolerant of limited data

ldodds:
Makes the recent usage page tolerant of 2 cases of limited data:

* If a school is not yet data enabled, then there isn't a summary of recent usage. But the school should be in the list
* If a school has <1 year of data, then there is no year data in the summary of recent usage

The PR also includes some display tweaks:

* Ensure table values are right aligned
* Switch table ordering, to put school names on left, which feels more natural
* Put "School" title on second header row
* Replace "n/a" with a "-"

The PR also tidies up the system spec for the page:

* remove unneeded declared variables
* move other variables inside the 'describe' block where they are used, to make it clearer how they are used
* DRY up repeated checks by adding some shared examples
* Add placeholders where we don't have tests for some of the tab content yet.

I think later PRs may need to review table layout, its quite busy for large groups. Optimise query so we can avoid N+1 queries for the data and instead just load what's needed. Will require changing how we store the management table data which isn't ideal at the moment.

### DIFF
--- a/app/views/school_groups/recent_usage.html.erb
+++ b/app/views/school_groups/recent_usage.html.erb
@@ -31,32 +31,66 @@
     </tr>
   </thead>
   <tbody>
-    <% @school_group.schools.visible.order(:name).each do |school| %>
-      <% recent_usage = school.recent_usage %>
+    <% @school_group.schools.data_visible.visible.order(:name).each do |school| %>
       <tr>
-        <% if @fuel_types.include?(:electricity) %>
-          <td class="<%= recent_usage&.electricity&.week&.message_class %>" data-order="<%= recent_usage&.electricity&.week&.change&.gsub('%','')&.to_f || '0' %>">
-            <%= up_downify(recent_usage&.electricity&.week&.change || 'n/a') %>
-          </td>
-          <td class="<%= recent_usage&.electricity&.year&.message_class %>" data-order="<%= recent_usage&.electricity&.year&.change&.gsub('%','')&.to_f || '0'  %>">
-            <%= up_downify(recent_usage&.electricity&.year&.change || 'n/a') %>
-          </td>
-        <% end %>
-        <% if @fuel_types.include?(:gas) %>
-          <td class="<%= recent_usage&.gas&.week&.message_class %>" data-order="<%= recent_usage&.gas&.week&.change&.gsub('%','')&.to_f || '0' %>">
-            <%= up_downify(recent_usage&.gas&.week&.change || 'n/a') %>
-          </td>
-          <td class="<%= recent_usage&.gas&.year&.message_class %>" data-order="<%= recent_usage&.gas&.year&.change&.gsub('%','')&.to_f || '0' %>">
-            <%= up_downify(recent_usage&.gas&.year&.change || 'n/a') %>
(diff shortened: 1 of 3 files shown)


app/views/school_groups/recent_usage.html.erb
```diff
@@ -27,40 +28,75 @@
...
-            <%= up_downify(recent_usage&.storage_heaters&.week&.change || 'n/a') %>
-          </td>
-          <td class="<%= recent_usage&.storage_heaters&.year&.message_class %>" data-order="<%= recent_usage&.storage_heaters&.year&.change&.gsub('%','')&.to_f || '0' %>">
-            <%= up_downify(recent_usage&.storage_heaters&.year&.change || 'n/a') %>
-          </td>
-        <% end %>
         <td>
           <%= link_to school.name, school_path(school) %>
         </td>
+        <% if school.data_enabled %>
```
iant, line 39: Perhaps some of these nested ifs could be extracted to a single partial?  e.g. replace lines 34-101 with
```
    <% @school_group.schools.visible.order(:name).each do |school| %>
      <tr>
        <td>
          <%= link_to school.name, school_path(school) %>
        </td>
        <% if school.data_enabled %>
          <% recent_usage = school.recent_usage %>
          <%= render 'recent_usage_table_row', fuel_type: :electricity, recent_usage: recent_usage.electricity if  @fuel_types.include?(:electricity) %>
          <%= render 'recent_usage_table_row', fuel_type: :gas, recent_usage: recent_usage.gas if @fuel_types.include?(:gas) %>
          <%= render 'recent_usage_table_row', fuel_type: :storage_heaters, recent_usage: recent_usage.storage_heaters if @fuel_types.include?(:storage_heaters) %>
        <% else %>
          <% (@fuel_types.reject{|f| f == :solar_pv}.length * 2).times do %>
            <td class="text-right" data-order="'0'">-</td>
          <% end %>
        <% end %>
      </tr>
    <% end %>
```
and then a partial `_recent_usage_table_row.html.erb` that looks like
```
<% if recent_usage&.week&.has_data %>
  <td class="text-right <%= recent_usage&.week&.message_class %>"
      data-order="<%= recent_usage&.week&.change&.gsub('%','')&.to_f || '0' %>">
    <%= up_downify(recent_usage&.week&.change) %>
  </td>
<% else %>
  <td class="text-right" data-order="0">-</td>
<% end %>
<% if recent_usage&.year&.has_data %>
  <td class="text-right <%= recent_usage&.year&.message_class %>"
      data-order="<%= recent_usage&.year&.change&.gsub('%','')&.to_f || '0'  %>">
    <%= up_downify(recent_usage&.year&.change) %>
  </td>
<% else %>
  <td class="text-right" data-order="0">-</td>
<% end %>
```

iant (approved):
:+1: Looks good to me @ldodds - 1 comment on a suggested possible refactor/simplification to extract some of the repeated template bits out to a partial

app/views/school_groups/recent_usage.html.erb
ldodds: I may do that in a later PR, but I want to get some feedback on the current table structure first.

PR #2806, opened 2023-06-06
Reduce CloudWatch logging of GoodJob queues

ldodds:
When we first starting using GoodJob we started logging all available metrics for the queues.

Now that we've had it running for a while, its clear that the main thing we want to monitor is the size of the queue. As this shows how quickly the queues get processed.

This PR updates the rake task that logs the metrics to just record that information.


iant (approved):
:+1: All looks good to me @ldodds 

One comment - maybe we could add a "# possible statuses: [:scheduled, :retried, :queued, :running, :finished, :discarded]" comment above the loop as a reminder in case we want to add one back in at a later date?

ldodds:
Have added a comment @iant

PR #2807, opened 2023-06-06
Skip checking whether meters are SMETS2 meters on test server

ldodds:
Add an environment check to the rake task, so its skipped on the test server.

Avoids doing a lot of API calls and then sending an email when we don't need them.


iant (approved):
Looks good to me @ldodds

PR #2809, opened 2023-06-06
Create school group comparison page & component

iant:


app/helpers/advice_page_helper.rb
```diff
@@ -55,6 +55,19 @@ def format_rating(rating)
     rating > 4 ? "Limited variation" : "Large variation"
   end
 
+  def compare_benchmark_key_for(advice_page_key)
```
ldodds, line 58: Minor: but if this is for the school groups page, then maybe the helper should be in the SchoolGroupsHelper?

spec/services/school_groups/categorise_schools_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+  let(:school_27) { create(:school, id: 27, name: "School 27", school_group: school_group_2, active: true, visible: true) }
+  let(:school_28) { create(:school, id: 28, name: "School 28", school_group: school_group_2, active: true, visible: true) }
+  let(:school_29) { create(:school, id: 29, name: "School 29", school_group: school_group_2, active: true, visible: true) }
+  let(:school_30) { create(:school, id: 30, name: "School 30", school_group: school_group_2, active: true, visible: true) }
+
+  let!(:baseload_advice_pages) { create(:advice_page, key: 'baseload') }
+  let!(:electricity_intraday_advice_pages) { create(:advice_page, key: 'electricity_intraday') }
+  let!(:electricity_long_term_advice_pages) { create(:advice_page, key: 'electricity_long_term') }
+
+  before do
```
ldodds, line 43: Is there scope to DRY up the spec by making more use of FactoryBot and maybe some traits or helpers to avoid long lists of data? e.g, something like. 

```
create_list(:school, 10, :with_advice_page_benchmarks, advice_page_benchmarks: { base_load_advice_page: :exemplar, electricity_intraday_page: :well_managed})
```

It'd make the specs easier to read.

app/components/school_group_comparison_component/school_group_comparison_component.html.erb
```diff
@@ -0,0 +1,85 @@
+<div class="pb-2">
+  <strong><%= t("advice_pages.#{@advice_page_key}.page_title") %></strong>
+  <p><%= t("school_groups.comparisons.how_does_it_compare.#{@advice_page_key}", default: '') %></p>
+  <div
+    id="<%= @id %>"
+    class="school-group-comparison-component-callout-row row mt-4">
+    <% SchoolGroupComparisonComponent::CATEGORIES.each do |category| %>
```
ldodds, line 7: I think you could inline some of this code into the component, by adding extra methods that are called by the template, e.g

```
<% categories.each do |category, count| %>
<% end %>
```

Where `categories` is a method on the component that yields the category and a count.

ldodds (approved):

spec/services/school_groups/categorise_schools_spec.rb
iant: I've moved the upsert to a method so it's a little easier to read now.  I think for now i'll leave as is and make a note/ticket to revisit the spec when the group pages are finalised.

PR #2811, opened 2023-06-07
Tweak rollbar error reporting for equivalence errors

ldodds:
Help to track down an intermittent bug generating equivalances, add the id of the equivalence to the Rollbar error.


iant (approved):
:+1:

PR #2815, opened 2023-06-08
Solar panel report

ldodds:
Adds a new admin report to summarise the solar panel configuration across the site.

* Adds a new query to extract the configuration from the MeterAttributes model
* Adds new report to the admin reports section with a simple tabular view of the attributes, with links to school, meter and the configuration

This should help us take stock of which schools are using synthetic solar data and the size of the installations across the service.


iant (approved):
looks good to me @ldodds

PR #2821, opened 2023-06-13
Reduce number of queries for map on /schools and group pages

iant:


ldodds (approved):
This is a definite improvement. Ideally we'd go a step further and just load the attributes we need to populate the map to avoid loading the full School and Configuration model (see note in the Trello card). Maybe as a follow-on PR?

PR #2825, opened 2023-06-13
Allow user to switch between different views of the recent usage table

iant:


ldodds (changes requested):
Can you add some tests around this please? I'd like to see some new/updated rspec tests for all new behaviour really.  E.g. a test for the helper method, which could ensure we're returning the right values, and maybe a simple system spec than checks that the right values are displayed as a result.

iant:
some (helper & system) tests added to this one @ldodds if you'd like to re-review

ldodds (approved):

PR #2826, opened 2023-06-13
Allow usage to switch between different views of the recent usage table (js)

iant:


iant:
Closed in favour of https://github.com/Energy-Sparks/energy-sparks/pull/2825 (see more info in ticket)

PR #2827, opened 2023-06-13
Add navbar to school group chart updates page

iant:
This pr 
- [ ] adds the management subnav to the group chart settings page. 
- [ ] renames the page title 

For consistency with other group pages it also adds a breadcrumb

### DIFF
--- a/app/views/school_groups/chart_updates/index.html.erb
+++ b/app/views/school_groups/chart_updates/index.html.erb
@@ -1,5 +1,4 @@
 <% content_for :page_title, @school_group.name + ' | ' +  t('school_groups.chart_updates.index.group_chart_settings').capitalize %>
-<%= render 'school_groups/sub_nav' if can?(:update_settings, @school_group) %>
 <div class="page-breadcrumb pt-5">
   <%= component 'breadcrumbs' do |c| %>
     <% c.with_items(@breadcrumbs) %>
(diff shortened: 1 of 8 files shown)


app/views/school_groups/chart_updates/index.html.erb
```diff
@@ -1,3 +1,11 @@
+<% content_for :page_title, @school_group.name + ' | ' +  t('school_groups.chart_updates.index.group_chart_settings').capitalize %>
+<%= render 'school_groups/sub_nav' if can?(:update_settings, @school_group) %>
```
ldodds, line 2: Is there a more consistent way we can ensure this appears on all school group pages? 

Currently, we've got an explicit check for the school groups controller in `application.html.erb` (`controller_name == 'school_groups' && can?(:update_settings, @school_group)`) and then we have this check here in a specific template. 

Am wondering if we can add a more generic test / helper in the main template, e.g. to check if this is a controller in the SchoolGroups module, and they have update settings?

ldodds (changes requested):

app/views/school_groups/chart_updates/index.html.erb
iant: :+1: have moved the check to display the group pages sub nav to application.html.erb

iant:
test and sub nav changes added if you want to re-review @ldodds

ldodds (approved):

PR #2828, opened 2023-06-13
Add extra shortcuts to EnergySparks admin section of "Manage Group" menu

iant:


app/views/school_groups/_sub_nav.html.erb
```diff
@@ -24,6 +24,11 @@
                 <div class="dropdown-divider"></div>
                 <%= link_to t('school_groups.sub_nav.group_admin'), admin_school_group_path(@school_group), class: 'dropdown-item' %>
                 <%= link_to t('school_groups.sub_nav.set_message'), edit_admin_school_group_dashboard_message_path(@school_group), class: 'dropdown-item' %>
+                <%= link_to t('school_groups.sub_nav.group_admin'), admin_school_group_path(@school_group), class: 'dropdown-item' %>
```
ldodds, line 27: Could you add a spec for this please? For the school dashboard we have a "manage_school_spec" which checks the menu contents for different roles. It'd be helpful to add the same for these, so can you either expand any existing tests for the menu or add one?

ldodds (changes requested):

iant:
some more tests added to this one so this is ready for review again @ldodds

PR #2830, opened 2023-06-15
Fix breadcrumb whitespace on school group page

iant:


ldodds (approved):
Looks OK, but I think we need a better solution to these header issues

PR #2840, opened 2023-06-15
Change wording on gas insights "how do you compare" section

urbanwide:
Have changed this for all the gas comparisons on the school group pages.
Have changed the text on the gas_out_of_hours and gas_recent_changes advice pages - however the other gas ones seem to work differently so I have left those as is.


iant (approved):
:+1:

PR #2841, opened 2023-06-15
Add CSV downloads for tables on school group dashboard

iant:


app/views/school_groups/recent_usage.html.erb
```diff
@@ -1,16 +1,21 @@
...
+      <% ['change', 'usage', 'cost', 'co2'].each do |metric| %>
+        <div class="form-check form-check-inline">
+          <%= f.radio_button :metric, metric, onclick: 'this.form.submit();', class: 'form-check-input', checked: radio_button_checked_for(metric) %>
+          <%= f.label :metric, t("school_groups.show.metric.#{metric}"), value: metric, class: 'form-check-label' %>
+        </div>
+      <% end %>
+    </div>
+  <% end %>
+  <div>
+    <%= link_to t('school_groups.download_as_csv'), school_group_path(@school_group, format: :csv, metric: params['metric']), class: 'btn btn-primary' %>
```
ldodds, line 16: Can you make the classes: `btn btn-sm`

app/services/school_groups/recent_usage_csv_generator.rb
```diff
@@ -0,0 +1,76 @@
...
+      CSV.generate(headers: true) do |csv|
+        csv << headers
+        @school_group.schools.visible.order(:name).each do |school|
+          recent_usage = school&.recent_usage
+
+          row = []
+          row << school.name
+          if fuel_types.include?(:electricity)
+            if recent_usage&.electricity&.week&.has_data
+              row << Nokogiri::HTML(recent_usage&.electricity&.week.send(@metric)).text
```
ldodds, line 19: As this is to undo the formatting done by SummaryTableData when presenting the data for display, I suggest we make some changes to that class so we can return the unformatted numbers without %.

E.g. this line in that class formats the raw number from the database:

`change: format_number(fetch(fuel_type, period, :percent_change), :comparison_percent, :text),`

We could add:

`change_numeric: FormatEnergyUnit.format(:comparison_percent, VALUE_HERE, :text, false, true, :target)`

Which formats the value for text, e.g. "100%").

Then in the CSV export we can use `change_numeric` rather than `change`?

ldodds (changes requested):

app/controllers/school_groups_controller.rb
```diff
@@ -90,7 +90,17 @@ def current_dashboard
...
-      render 'recent_usage'
+      respond_to do |format|
+        format.html do
+          render 'recent_usage'
+        end
+        format.csv do
+          metric = params['metric'] || 'change'
+          metric_label = I18n.t("school_groups.show.metric.#{metric}")
+          send_data SchoolGroups::RecentUsageCsvGenerator.new(school_group: @school_group, metric: metric).export,
+          filename: "#{@school_group.name} - #{I18n.t('school_groups.titles.recent_usage')} - #{metric_label}.csv"
```
ldodds, line 101: Can you add the date to the filename? So if someone is regularly requesting this, then the versions are clear

ldodds (changes requested):

ldodds (approved):
:+1:

PR #2842, opened 2023-06-16
Align column headings

urbanwide:


iant (approved):
👍 : looks good to me @urbanwide

PR #2844, opened 2023-06-16
Add CSV downloads for tables on school group dashboard: 2. Priority Actions

iant:


app/services/school_groups/priority_actions_csv_generator.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    def service
+      @service ||= SchoolGroups::PriorityActions.new(@school_group)
+    end
+
+    def headers
+      [
+        I18n.t('advice_pages.index.priorities.table.columns.fuel_type'),
+        '',
```
urbanwide, line 39: Does this need a label?

app/controllers/school_groups_controller.rb
```diff
@@ -28,9 +28,17 @@ def comparisons
...
-    @total_savings = sort_total_savings(service.total_savings)
+    respond_to do |format|
+      format.html do
+        service = SchoolGroups::PriorityActions.new(@school_group)
+        @priority_actions = service.priority_actions
+        @total_savings = sort_total_savings(service.total_savings)
+      end
+      format.csv do
+        send_data SchoolGroups::PriorityActionsCsvGenerator.new(school_group: @school_group).export,
+        filename: "#{@school_group.name} - #{I18n.t('school_groups.titles.priority_actions')} (#{Time.zone.now.strftime('%d/%m/%Y')}).csv"
```
urbanwide, line 39: I'd be tempted to not use slashes in the time within the filename? and maybe not brackets or spaces either? (Or was this a requirement?)
`"#{@school_group.name}-#{I18n.t('school_groups.titles.priority_actions')}-#{Time.zone.now.strftime('%Y-%m-%d` ')}".parameterize + ".csv"

urbanwide (approved):
Looks good @iant - just a couple of minor comments depending on requirements 👍

app/controllers/school_groups_controller.rb
iant: :+1: thanks @urbanwide  will push an update going with your suggested change.

app/services/school_groups/priority_actions_csv_generator.rb
iant: It's missing from the table in the ui but probably does need including here.  I'll put the text 'Description' in for now.

PR #2845, opened 2023-06-16
Add CSV downloads for tables on school group dashboard: 3. Current Scores

iant:


app/controllers/school_groups_controller.rb
```diff
@@ -34,6 +34,13 @@ def priority_actions
   end
 
   def current_scores
+    respond_to do |format|
+      format.html {}
+      format.csv do
+        send_data SchoolGroups::CurrentScoresCsvGenerator.new(school_group: @school_group).export,
+        filename: "#{@school_group.name} - #{I18n.t('school_groups.titles.current_scores')} (#{Time.zone.now.strftime('%d/%m/%Y')}).csv"
```
urbanwide, line 41: Same comment as in previous PR about not having slashes in a filename, and possibly not spaces or brackets, something like this?:
`"#{@school_group.name}-#{I18n.t('school_groups.titles.current_scores')}-#{Time.zone.now.strftime('%Y-%m-%d ')}".parameterize + ".csv"`

urbanwide (approved):
Looks good @iant 👍 Just the same small comment as in last PR about the filename...

app/controllers/school_groups_controller.rb
iant: :+1: thanks @urbanwide will change this one too

PR #2846, opened 2023-06-19
Add a funder allocation report

ldodds:


spec/services/schools/funder_allocation_report_service_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+    let!(:solar_meter) { create(:solar_pv_meter, active: true, data_source: data_source_3, procurement_route: procurement_route_3, school: school_1)}
+
+    #only basic data, helps to catch errors checking for nils
+    let!(:school_2)  { create(:school, visible: true, school_group: create(:school_group)) }
+    #not included in export
+    let!(:not_visible)  { create(:school, visible: false, school_group: school_group) }
+
+    let(:csv)   { service.csv }
+
+    it 'returns rights headers' do
```
urbanwide, line 66: Think you could mean "right headers" rather than "rights headers" here?

spec/services/schools/funder_allocation_report_service_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+          nil,
+          nil,
+          gas_meter.data_source.name,
+          nil,
+          nil,
+          gas_meter.procurement_route.organisation_name,
+          nil,
+          nil,
+          solar_meter.data_source.name,
+          nil,
```
urbanwide, line 104: I think you could add test data in for these nil fields at some point (i.e. where there are 1+ of the fields when there can be 3 e.g. electricity_data_sources). I'm sure it'd be fine to go out now though as is as it's urgent 👍

urbanwide (approved):
Looks good @ldodds 👍 Couple of very minor suggestions but nothing that would stop it being deployed.

PR #2859, opened 2023-06-22
Hide group sub nav for current pages

iant:


ldodds (approved):
Tested locally, looks good :+1:

PR #2860, opened 2023-06-22
N3rgy tooling

ldodds:
This PR has some improvements to our admin pages to help to support debugging and investigation of issues with n3rgy meters.

This includes improvements to the dcc meter report to list:

* ignore sandbox meters
* include number of schools with meters
* add a check to see if n3rgy think the meter has been consented
* include a list of the mpans that n3rgy think are consented

And some changes to the meter page for admins:

* Remove the "elements" data
* Remove the "earliest available date" field as its unused
* Improve some of the labelling to explain fields
* Show the symbol describing the status of the meter in n3rgy, rather than just a true/false flag
* Check to see if the mpan_mprn is in the list of consented meters in n3rgy
* Include the available date range

### DIFF
--- a/app/controllers/admin/reports/dcc_status_controller.rb
+++ b/app/controllers/admin/reports/dcc_status_controller.rb
@@ -1,8 +1,21 @@
 module Admin
   module Reports
     class DccStatusController < AdminController
+      before_action :set_consented_mpxns
+
       def index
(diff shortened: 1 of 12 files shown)


iant (approved):
:+1: looks good to me @ldodds

PR #2863, opened 2023-06-23
Fix bug with n3rgy status check

ldodds:
Changes to admin interface to better display information from the n3rgy API introduced an issue with the Meter Review process. A method that should have been returning a boolean was not returning a symbol. This meant that reviews cannot be processed.

This PR fixes that issue by adding a new method in MeterManagement to check whether a meter is in n3rgy, separate from its "status" which returns one of several different symbols.

As part of implementing the fix I've further tidied up the diagnostic information to:

* ensure its only displayed to admins
* reduce number of API calls if a meter isn't known to n3rgy
* fix up some minor display issues and further improve the diagnostics

### DIFF
--- a/app/controllers/schools/meters_controller.rb
+++ b/app/controllers/schools/meters_controller.rb
@@ -20,12 +20,7 @@ module Schools
     end
 
     def show
-      manager = MeterManagement.new(@meter)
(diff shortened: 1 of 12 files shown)


iant (approved):
:+1: looks good to me @ldodds

PR #2866, opened 2023-06-26
Allow group admins and es admins to create clusters

urbanwide:
Allows admins and group admins to create clusters & adds a link to the school group menu.

This is a fairly basic layout at the moment which we may want to enhance depending on what is required. At the moment, the ability to add schools to the cluster is part of the cluster edit form. Do we want this or does it need it's own page / form @ldodds?


ldodds:
Some general feedback on the UX, haven't reviewed the code:

I think the menu item should be "Manage Clusters" instead of "Clusters". Also think we need some introductory text on the page, e.g:

"A cluster is a set of schools that are group together within your [Multi-Academy Trust/Local Authority]. For example, you may want to group all your southern and northern schools, all your primaries, etc. Grouping schools into clusters will allow you to more easily compare related schools"

I think the UI needs to be broken in two:

- a simple form field for creating or deleting a named cluster
- a UX for adding or removing schools from a cluster

For the latter, I'm wondering whether on the clusters index page you just list all of the visible schools with options to assign schools to a cluster by ticking a check box next to each school, then choosing from a select box at the top of the list and then "Assign"? The option should have the option to remove a school from a cluster. 

That tabular view will make it easier to see which schools are in each cluster, as well as those that aren't assigned to any cluster. The current approach doesn't easily let me see which schools are in which cluster or those that aren't assigned to any.

While we might want to allow schools to be in multiple clusters in the future, which would give them some powerful reporting options, for the moment I think we should limit it so that a school is in a single cluster. For the school group dashboard I was expecting to have one extra column which added the name of the cluster.

ldodds (changes requested):
See comments

urbanwide:
Thank you, much appreciated! I'll make those changes 👍

ldodds (approved):
Looks great!

PR #2874, opened 2023-06-28
Fix bug in Programme recording

iant:
This PR fixes a bug in the tracking of whether a school has recorded an activity.

- [x] Revises the ActivityCreator so it only adds an activity for the programme if its one of those in the programme type
- [x] Tidies up the data in the database through an After Party task that will tidy up the extra records in ProgrammeActivity: 
deleting any entries where the ActivityType is not part of the ProgrammeType and tidies up the Programme data by rechecking whether a programme has been completed. 
- [x] Updates the rspec tests around the activity creator service to ensure we’re testing a wider range of behaviours to ensure the error has been fixed


app/models/programme_activity.rb
```diff
@@ -18,4 +18,14 @@ class ProgrammeActivity < ApplicationRecord
   belongs_to :activity_type
   belongs_to :activity
   belongs_to :programme
+
+  validate :activity_type_is_also_a_programme_programme_type_activity_type
```
ldodds, line 22: Its possible that the list of activity type for a programme type might change. E.g. An admin revises the programme type to add or remove an activity type (perhaps replacing it with a more suitable one).

What will happen in this case as I assume those records will become invalid. Will this cause any problems when saving or updating their programme?

ldodds (approved):

PR #2876, opened 2023-06-28
Keep up to a month of batch data

ldodds:
A while ago we started deleting older data generated by the overnight batch processes, removing anything that was >3.months old.

The database is growing more rapidly as we have more schools and as there's less need to keep these artefacts I've changed the window to be one month.

Actually, as the logic is `1.month.ago.beginning_of_month` we'll be keeping a bit more than a month which is fine.


iant (approved):
:+1:

PR #2877, opened 2023-06-28
Provide tool for admin to check ranges of unvalidated data for a list of meters

ldodds:
This PR adds a new admin report that summarises data the unvalidated data in the `AmrDataFeedReading` table for a list of mpans.

The admins need to be able to confirm the latest reading dates for a given list of mpans. Sometimes those mpans are associated with meters, but sometimes they are not yet linked to meters. E.g. if they are starting to load data before the school has been onboarded.

Checking the dates allows them to re-request data from suppliers on a weekly basis, where those feeds are not fully automated.

Creating this type of report is tricky as the `AmrDataFeedReading` `reading_date` column is an unparsed date: we parse it when validating the data. That will get changed in the next few months but, for now I've written a query that is capable of converting the range of date formats used to load data into a date. 

I've translated the Ruby strftime formats we use in the data feed configs into the formats that postgres understands.

I've written some basic specs and have manually tested the query against the live database for performance (which seems OK) and to flush out edge cases (I've added support for some of these and specs for them).

This is a quick prototype that will get reworked following feedback.

### DIFF
--- a/app/controllers/admin/reports/unvalidated_readings_controller.rb
+++ b/app/controllers/admin/reports/unvalidated_readings_controller.rb
@@ -1,0 +1,27 @@
+module Admin
+  module Reports
+    class UnvalidatedReadingsController < AdminController
+      def show
+        @report = run_report
+      end
+
+      private
+
+      def run_report
+        if params[:mpans].present?
+          param = params[:mpans]
+          if param["list"].present?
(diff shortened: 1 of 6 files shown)


iant (approved):
Looks good to me @ldodds

PR #2879, opened 2023-06-29
Fix N3rgyDownloaderDates to default to start and end of day

iant:


ldodds (changes requested):
You'll need to update this with a new tagged version of the analytics.

Is there a need to reload any data? The examples in the analytics PR look like data might be offset, so do we need to reload data from when this bug was introduced?

iant:
@ldodds analytics tag updated.  Yes, i think we may need to reload the data (via a afterparty task?).  I think this should be as part of a separate ticket/pr.

PR #2880, opened 2023-06-30
Rework data source report to be emailed report with attachment

iant:


config/environments/development.rb
```diff
@@ -55,7 +55,7 @@
   # Debug mode disables concatenation and preprocessing of assets.
   # This option may cause significant delays in view rendering with a large
   # number of complex assets.
-  config.assets.debug = true
+  config.assets.debug = false
```
iant, line 58: These settings are to override an `ActionView::Template::Error (The asset "http:/assets/application_mailer.self-b22e42926eeed27303edf398a1ca57accaf6a551e82984b53573f67b71c73a94.css" is not present in the asset pipeline` error - feel free to set back

app/controllers/admin/data_sources_controller.rb
```diff
@@ -7,12 +7,18 @@ def show
       respond_to do |format|
         format.html
         format.csv do
-          send_data @data_source.meters.to_csv,
+          send_data @data_source.to_csv,
```
ldodds, line 10: If we no longer have a CSV download, then we don't need the controller method to respond to CSV.

app/controllers/admin/data_sources_controller.rb
```diff
@@ -7,12 +7,18 @@ def show
...
           filename: "#{t('common.application')}-#{@data_source.name}-meters-#{Time.zone.now.iso8601}".parameterize + '.csv'
         end
       end
     end
 
+    def deliver
+      @data_source = DataSource.find(params[:data_source_id])
+      SendDataSourceReportJob.perform_now(to: current_user.email, data_source_id: @data_source.id)
```
ldodds, line 18: This should be `perform_later` otherwise is runs the job immediately. We want it to run in the background within GoodJobs

app/models/meter.rb
```diff
@@ -238,20 +236,20 @@ def can_withdraw_consent?
...
 
-  def self.csv_attributes
-    %w{school.school_group.name school.name mpan_mprn meter_type.humanize active first_validated_reading last_validated_reading admin_meter_status_label open_issues_count}
+  def open_issues_count
+    open_issues.count
   end
 
-  def smart_meter_tariff_attributes
-    @smart_meter_tariff_attributes ||= Amr::AnalyticsTariffFactory.new(self).build
+  def open_issues_as_list
```
ldodds, line 247: I think it would best best to impose some default order on these, e.g. show the oldest issues as the first in the list.

ldodds (changes requested):

app/controllers/admin/data_sources_controller.rb
iant: 👍 done

iant:
@ldodds this pr's ready for another review now when you get a mo. thanks.

PR #2885, opened 2023-07-13
Fix delete button for dashboard message

urbanwide:


ldodds (approved):

urbanwide:
Thanks @ldodds. I meant horizontally really as there is probably too much space to the left of the icon (in my head anyway) - feels like the icon should align with the "Recent Usage" text. Vertically it doesn't quite look right either though with less text. One to be done separately?

PR #2887, opened 2023-07-14
Clarify error message for admin when uploading files that find meters using serial numbers

ldodds:
A data feed config can be configured to lookup meters by serial number. The admins need to add these to the database. 

If all the serial numbers in a file are missing, then this leads to an error as we have no valid readings in the file. This PR just tweaks the error reporting on the admin form to prompt them to check for this where necessary.

Also ensured there's different messages for "unable to parse file" and "no valid readings" as those are different situations.


urbanwide (approved):
All good @ldodds 👍

PR #2888, opened 2023-07-14
Remove local files when archived to s3

ldodds:
When looking at disk space issues on the servers I noticed that were weren't removing local copes of CSV files after they had been successfully processed.

This PR just adds an extra File.delete step if a file is successfully archived in S3. 

The rest of the changes are just refactoring of the importer and the spec to reduce the amount of code/configuration.

### DIFF
--- a/lib/amr/importer.rb
+++ b/lib/amr/importer.rb
@@ -24,11 +24,22 @@ module Amr
 
   private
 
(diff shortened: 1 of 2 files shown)


urbanwide (approved):
Looks good @ldodds 👍

PR #2890, opened 2023-07-18
Score points for completing a programme

iant:
We want to encourage schools to complete a programme (`ProgrammeType`). To do this we want to award bonus points for completing all the activities.  The points will be specific to the programme type as they vary in complexity.  This PR:
- [x] allows `ProgrammeTypes` to have a bonus score for completing the programme
- [x] updates admin forms to allow the bonus score to be changed
- [x] updates `/programme_types` views to include a score for each programme (“Completing this programme this academic year will score you [x] bonus points”)
- [x] updates programme type show page to include score for each programme (“Completing this programme this academic year will score you [x] bonus points”)
- [x] adds functionality to create an `Observation` to mark when a programme has been completed, which also records the related `ProgrammeType` points (even if bonus points are zero). 
- [x] adds functionality to associate the observation with the Programme so we have an audit trail
- [x] adds support for the new programme type observation to the school timelime


urbanwide (approved):
Looks good @iant 👍

ldodds (changes requested):
The behaviour to award bonus points and record the Observation need to be behind a feature flag as I don't want to switch that on until mid-August.

But the ability to add points should be available ASAP so the admins can add the points.

iant:
@ldodds feature flag added to this one if you'd like to re-review

db/schema.rb
```diff
@@ -854,21 +854,21 @@
     t.index ["replaced_by_id"], name: "index_global_meter_attributes_on_replaced_by_id"
   end
 
-  create_table "good_job_processes", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+  create_table "good_job_processes", id: :uuid, default: -> { "public.gen_random_uuid()" }, force: :cascade do |t|
```
ldodds, line 857: Need to remove the `public.` prefixes

app/models/programme.rb
```diff
@@ -32,6 +32,15 @@ class Programme < ApplicationRecord
     event :complete do
       after do
         self.update(ended_on: Time.zone.now)
+        if EnergySparks::FeatureFlags.active?(:activities_2023)
```
ldodds, line 35: This doesn't meet the specification: "When a school completes a programme, within the same academic year as they started it, then they should be awarded the bonus points."

This code will award them points whenever it was started. You'll need to use the schools calender to find the academic year for the programme start date and then check whether its the same for the completion date. See: `school.academic_year_for(...)`

ldodds (changes requested):

app/models/programme.rb
iant: @ldodds pushed a couple of updates - should (I think, anyway) now check and only create an `Observation` when the school completes a programme within the same academic year as the start date (couple of tests for the same added too).

ldodds: @iant sorry, I maybe I should have been clearer:

- its useful to mark the completion of the programme in the school timeline, regardless of the points. So the Observation can be created when the Programme is complete
- they should only get the bonus points, if its completed in the same academic year

So I think you just need to use the check for current academic year to decide whether to award zero points or the bonus points.

ldodds (changes requested):
See comment, one minor tweak required!

app/models/programme.rb
iant: Ah, ok.  sorry for my misunderstanding :)  no probs at all.  have pushed an update so an observation is still created but points only awarded if the programme is completed in same academic year

ldodds (approved):

PR #2894, opened 2023-07-19
Update development config

iant:


urbanwide (approved):
Looks good @iant 👍 
I have had a look and my ASSET_HOST is set in .env 👍

PR #2895, opened 2023-07-19
Add extra table header to priority action popup

ldodds:
Clarify these are savings by adding extra header row as we do in other versions of the table


iant (approved):
:+1:

PR #2898, opened 2023-07-19
Combine csv downloads for recent usage tab on school group dashboard

ldodds:
This PR changes the CSV downloads for the Recent usage tab on the school group dashboard.

Instead of having separate downloads for each metric, we instead have a single download that includes all of the metrics (%, kwh, £, co2) for both periods (last week, last year) for all fuel types. This is intended to be a convenience for anyone intending to do any analysis that would otherwise involve downloading and combining separate files.

The tests have been revised to use separate values for each fuel type to better check that the right values are in the right columns. 

I've also simplified the system spec so its just checking that the download is available and that the generated file name is correct. The service spec already covers the format and content of the CSV.


iant (approved):
Looks good to me @ldodds

PR #2900, opened 2023-07-19
Award bonus points for adding photos to activities and actions

iant:
This PR changes the scoring system for both pupil and adult lead activities so that bonus points are awarded if the school adds one or more photos to the description.  

- [x] Adds a new system wide setting for default bonus points for adding images to activities and actions
- [x] Adds a note to form field to both forms to indicate bonus points
- [x] Revises logic to ensure bonus points are recorded correctly

### DIFF
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -70,6 +70,7 @@ class Observation < ApplicationRecord
   private
 
   def add_points_for_included_images
+    return unless EnergySparks::FeatureFlags.active?(:activities_2023)
     # Do not add points if site wide photo bonus points are set to nil or zero
     return unless SiteSettings.current.photo_bonus_points&.nonzero?
     return unless description_includes_images?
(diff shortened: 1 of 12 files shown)


iant:
Screenshot 2023-07-20 at 15 31 04

iant:
Screenshot 2023-07-20 at 15 34 40

app/models/observation.rb
```diff
@@ -64,16 +64,32 @@ class Observation < ApplicationRecord
...
 
-  before_save :add_points_for_interventions
+  before_save :add_points_for_interventions, if: :intervention?
+  before_save :add_points_for_included_images, if: proc { |observation| observation.activity? || observation.intervention? }
 
-private
+  private
 
-  def add_points_for_interventions
+  def add_points_for_included_images
```
ldodds, line 72: This isn't quite right. If a school records an activity or intervention for a previous school year and adds a photo, they'll still get the bonus points for the image. They should only get the bonus if they're actually scoring points. 

So think it needs a check to confirm that the bonus is only added if the score is non zero?

ldodds (changes requested):

ldodds (approved):

PR #2903, opened 2023-07-19
Issues email report updates

urbanwide:
* Add group and updated_by to issues report
* Order issues by updated_at rather than created_at


iant (approved):

PR #2917, opened 2023-07-21
Award points for completing assigned activities from an audit

iant:
We want to encourage schools to complete the pupil lead activities assigned to them in their audits.  To do this we will award bonus points for completing all of the activities

This PR: 

- [x] adds a new `audit_activities_bonus_points` to the site settings so it that is configured across the system.
- [x] updates the audit page so the school can see how many bonus points they will score to include the text “Completing all of these activities will award you [X] points” (where the [X] is styled as a green pill similar to how we show scores elsewhere).
- [x] Add a new `activities_completed?` method so we can to identify when all activities in an audit have been completed 
- [x] Award the bonus points as an observation when an activity is saved: 
- [x] bonus points should only be awarded if the activities or actions are completed within 12 months of the Audit being created  
- [x] Only one observation should be created per audit


app/models/observation.rb
```diff
@@ -46,7 +46,7 @@ class Observation < ApplicationRecord
   belongs_to :programme, optional: true
   belongs_to :audit, optional: true
   belongs_to :school_target, optional: true
-  enum observation_type: [:temperature, :intervention, :activity, :event, :audit, :school_target, :programme]
+  enum observation_type: [:temperature, :intervention, :activity, :event, :audit, :school_target, :programme, :audit_activities_completed]
```
ldodds, line 49: Adding a new Observation type means you also need to add a timeline ERB file so that the observation displays as part of the school timeline, as you did for `:programme` in a recent PR. Without that when the observation is added the school timeline will break. So you'll need to add a template for that.

ldodds (changes requested):

app/models/observation.rb
iant: :+1: done (and a reminder comment added too)

ldodds (approved):

PR #2920, opened 2023-07-25
New common data model for storing energy tariffs

ldodds:
This PR creates a new set of data models and relationships for managing energy tariffs.

It builds on the existing "user tariff" models used by the tariff editor, but generalises them so they are a bit more flexible and extensible. Specifically it allows tariffs to be owned by any `tariff_holder` model which will include the system, groups and schools.

This PR will just add the models, a later PR will include some data migration code.

* [x] Create migration for new tables
* [x] Copy over validations and meter attribute code from original models (UserTariff, UserTariffCharge, UserTariffPrice)
* [x] Revise rspecs to allow for model differences
* [x] Refactor and tidy up tests

### DIFF
--- a/app/models/energy_tariff.rb
+++ b/app/models/energy_tariff.rb
@@ -1,0 +1,128 @@
+# == Schema Information
+#
+# Table name: energy_tariffs
+#
+#  ccl                :boolean          default(FALSE)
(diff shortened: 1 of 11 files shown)


iant (approved):
looks good to me @ldodds

PR #2921, opened 2023-07-25
Create energy tariff data migration script

ldodds:
Updates EnergyTariff model associations so prices and charges are destroyed.

Add an after party task to populate the new models from the old user tariff models.

The task is destructive so shouldn't be re-run in this form after release.


iant (approved):
:+1:

PR #2922, opened 2023-07-25
Pin version of chrome driver

ldodds:
Fix for an issue with incorrect Chrome driver version being used.

Issue described here: https://github.com/titusfortner/webdrivers/issues/247

For the moment pinning the version looks like best approach, as switching to use Selenium only has other issues (a number of deprecations and latest version of the gem isn't available for our version of Ruby).


iant (approved):
:+1:

PR #2924, opened 2023-07-25
Remove unused meter consolidation alerts

ldodds:
Removes the meter consolidation alerts:

* [x] Update to new version of analytics with alert removed (https://github.com/Energy-Sparks/energy-sparks_analytics/pull/576)
* [x] Remove the unused meter attributes
* [x] Remove the alerts
* [x] Remove the alert type

### DIFF
--- a/lib/tasks/deployment/20230725161734_remove_meter_consolidation_alert.rake
+++ b/lib/tasks/deployment/20230725161734_remove_meter_consolidation_alert.rake
@@ -1,0 +1,23 @@
+namespace :after_party do
+  desc 'Deployment task: remove_meter_consolidation_alert'
+  task remove_meter_consolidation_alert: :environment do
+    puts "Running deploy task 'remove_meter_consolidation_alert'"
+
+    #Delete unneeded attributes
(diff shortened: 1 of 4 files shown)


lib/tasks/deployment/20230725161734_remove_meter_consolidation_alert.rake
```diff
@@ -0,0 +1,23 @@
+namespace :after_party do
+  desc 'Deployment task: remove_meter_consolidation_alert'
+  task remove_meter_consolidation_alert: :environment do
+    puts "Running deploy task 'remove_meter_consolidation_alert'"
+
+    #Delete unneeded attributes
+    GlobalMeterAttribute.where(attribute_type: "indicative_standing_charge").destroy_all
+
```
iant, line 8: is it worth wrapping this in a transaction?

ldodds: We don't often do that for these changes. I don't think there will be any issues (I've run locally to test).

PR #2928, opened 2023-07-27
Add migration for global meter attributes

ldodds:
This PR adds another step to the data migration for tariffs:

* Creates a separate service to do the migration of user tariffs and global meter attributes. This makes it easier to test and possibly re-run the migration
* Update the after party task to call this step, but remove the "delete all", so we're a little more cautious when doing the migration
* Allows EnergyTariff to be owned by SiteSettings
* Adds some basic specs for the initial migration steps

I'll do the migration of SchoolGroup and School attributes in a later PR. This is a little more involved as we need to selectively migrate some attributes and finesse some of the time ranges.

### DIFF
--- a/app/models/energy_tariff.rb
+++ b/app/models/energy_tariff.rb
@@ -34,7 +34,7 @@
 class EnergyTariff < ApplicationRecord
   belongs_to :tariff_holder, polymorphic: true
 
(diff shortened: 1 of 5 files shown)


iant (approved):
👍

PR #2929, opened 2023-07-27
Analytics 2.9.0

ldodds:
Update analytics to version that has unneeded tariff code removed https://github.com/Energy-Sparks/energy-sparks_analytics/pull/577

Want to release this to test to rule out any issues when run against test database.


iant (approved):
:+1:

PR #2930, opened 2023-07-27
Add a new funder field to schools

iant:


app/models/funder.rb
```diff
@@ -0,0 +1,4 @@
+class Funder < ApplicationRecord
+  has_many :schools
+  validates :name, presence: true
```
ldodds, line 3: Maybe add a validation so that the name is unique too?

app/views/schools/_details_form.html.erb
```diff
@@ -17,6 +17,8 @@
     <%= f.input :activation_date, as: :tempus_dominus_date, default_date: nil, hint: t('schools.school_details.activation_date_hint') %>
     <%= f.input :enable_targets_feature %>
     <%= f.input :public %>
+    <%= f.label :funder %>
+    <%= f.select :funder_id, options_for_select(Funder.all.pluck(:name, :id), school.funder_id), { include_blank: true }, { class: 'form-control' } %>
```
ldodds, line 21: I did suggest adding it to the configuration form rather than the detailed form, but I'd forgotten we had some admin specific things here. At some point I'd like to have all the admin options on the configuration form `/schools/:id/configuration/edit` rather than on this form as it ensure that the admin options are all in one place.

ldodds, line 21: Please sort them by name

ldodds (changes requested):

app/views/schools/_details_form.html.erb
iant: :+1: I've added a funder field to the configuration form too

iant: :+1:

iant: @ldodds those changes pushed (uniqueness validation, sort ordering & added field to configuration form too) so should be ok for another review now

ldodds (approved):

PR #2935, opened 2023-07-28
Add code for migrating school group tariffs

ldodds:
This PR adds extra code to migrate SchoolGroup meter attributes to the new tariff data model.

The main entry point is the `migrate_school_group_meter_attributes` method. This decides whether to migrate the group's economic tariffs or accounting tariffs.

As the team are currently prioritising populating the economic tariffs to follow latest prices, we use those over the accounting tariffs which may be incomplete or out of date. 

When migrating the economic tariffs if the tariff specifies both flat and differential rates, then we just migrate the differential rates. This is because the MeterAttribute editor has the flat rate marked as a required field. But we won't need both tariff types once migrated.

As economic tariffs don't have charges, we just migrate the prices.

If a school group has accounting tariffs then we'll migrate those as a fallback. The code handles both flat rate and differential accounting tariffs with all known charge types.

One note about time ranges for differential tariffs:

The time ranges on the meter attributes are *exclusive* time ranges, so a range of 00:00 to 07:00 means any consumption up until 07:00. However the time ranges in the new energy tariff model are *inclusive* so when migrating we need to turn this period into 00:00 to 06:30 to be consistent. The differences are due to the analytics handling these time ranges differently across different attribute types.

### DIFF
--- a/app/services/database/energy_tariff_migration_service.rb
+++ b/app/services/database/energy_tariff_migration_service.rb
@@ -88,32 +88,140 @@ module Database
       end
     end
 
+    def self.migrate_school_group_meter_attributes
+      ActiveRecord::Base.transaction do
+        SchoolGroup.all.order(:id).each do |school_group|
+          #If a school group has any time-varying economic tariffs (there are 11 of these) then
+          #we will use these to create new EnergyTariff records and their accounting tariffs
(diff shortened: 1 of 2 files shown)


iant (approved):
:+1:

PR #2937, opened 2023-07-31
Migrate school and meter tariffs

ldodds:
Updates the Energy Tariff migration code to finalise the basic migration code for the following:

- [x] School time varying economic tariffs
- [x] Meter specific accounting tariffs (flat & differential)
- [x] TariffPrices and TariffStandingCharges populated from n3rgy
- [x] Global solar default economic tariffs
- [x] DRY up the migration service spec to use shared examples
- [x] add some extra specs for utility methods

In a follow-up PR I'll add:

- some code for running the set of migrations
- some code for doing some precondition checks to confirm that basic assumptions are correct prior to doing a migration

### DIFF
--- a/app/services/database/energy_tariff_migration_service.rb
+++ b/app/services/database/energy_tariff_migration_service.rb
@@ -192,6 +192,29 @@ module Database
       end
     end
 
+    def self.migrate_meter_accounting_tariffs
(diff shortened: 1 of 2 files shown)


iant (approved):
:+1:

PR #2941, opened 2023-08-01
Do not store zero tariffs and prices from n3rgy

iant:


spec/services/amr/n3rgy_tariffs_download_and_upsert_spec.rb
```diff
@@ -45,6 +50,17 @@ module Amr
          .and change { TariffStandingCharge.count }.by(0)
         expect(TariffImportLog.last.error_messages).to be_blank
       end
+
+      it "should request 24 hours of data and upsert a new tariff price and tariff standing charge if there are no existing records of the same date" do
```
ldodds, line 54: think this needs renaming as its testing handling of zero tariffs?

ldodds: Although, as the changes are part of the n3rgy tariff upserter, maybe you should be adding specs to n3rgy_tariffs_upserter_spec.rb instead of, or in additon to the download and upserter?

spec/services/amr/n3rgy_tariffs_download_and_upsert_spec.rb
```diff
@@ -9,12 +9,17 @@ module Amr
     let(:end_date)            { DateTime.now.beginning_of_day }
     let(:start_date)          { DateTime.now.yesterday.beginning_of_day }
     let(:expected_tiered_tariff)      { {:tariffs=>{1=>0.48527000000000003, 2=>0.16774}, :thresholds=>{1=>1000}, :type=>:tiered} }
-    let(:expected_prices)             { [expected_tiered_tariff, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992, 0.15992] }
```
ldodds, line 12: Removing this is unnecessary I think?

app/services/amr/n3rgy_tariffs_upserter.rb
```diff
@@ -11,7 +11,11 @@ def initialize(meter:, tariffs:, import_log:)
     def perform
       return if @tariffs.empty?
       Rails.logger.info "Upserting #{kwh_tariffs.count} tariff_prices and #{standing_charges.count} tariff_standing_charges for #{@meter.mpan_mprn} at #{@meter.school.name}"
-      TariffUpserter.new(prices_array(kwh_tariffs), standing_charges_array(standing_charges), @import_log).perform
+
+      prices_array = prices_array(kwh_tariffs)
+      standing_charges_array = standing_charges_array(standing_charges)
+
+      TariffUpserter.new(prices_array, standing_charges_array, @import_log).perform
```
ldodds, line 18: The upserter will still end up recording that its inserted 0 prices and 0 standing charges, but we'll have no record that we've basically had no tariffs from the meter. I think we should record an error to the import log if all the prices were zero, or the standing charge is zero or negative. Probably needs to be done in this class.

ldodds (changes requested):

ldodds (approved):

PR #2943, opened 2023-08-01
Remove meter id check in unvalidated meter report spec

ldodds:
The spec is checking for the meter id to be present, although that's no longer included on the page.

I think this is sometimes passing when the meter id matches another value on the page. But this is seems to be regularly failing in builds:

https://github.com/Energy-Sparks/energy-sparks/actions/runs/5726381033/job/15516710682


iant (approved):
:+1:

PR #2944, opened 2023-08-01
Fix tariff type migration issue

ldodds:
Fixes an issue with migration meter attributes to EnergyTariffs.

The tariff type wasn't being correctly converted, so added a check for the actual `daytime_rate` (to check for differential tariffs) rather than just presence of the key.

The PR also adds a couple of methods to the migration service to run all the migrations, except for the user tariffs which will already have been run.


iant (approved):
:+1:

PR #2945, opened 2023-08-02
Add tariff holder and creation date to tariff meter attribute

ldodds:
Updates the EnergyTariff model so that when it serialises as a meter attribute it identifies the type of tariff holder and the date that the tariff was created.

Requires updates to analytics from this PR: https://github.com/Energy-Sparks/energy-sparks_analytics/pull/580. Until then tests will fail.

### DIFF
--- a/app/models/energy_tariff.rb
+++ b/app/models/energy_tariff.rb
@@ -77,7 +77,9 @@ class EnergyTariff < ApplicationRecord
       rates: rates,
       vat: vat_rate.nil? ? nil : "#{vat_rate}%",
       climate_change_levy: ccl,
-      asc_limit_kw: (value_for_charge(:asc_limit_kw) if rates_has_availability_charge?(rates))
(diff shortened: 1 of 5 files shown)


app/models/energy_tariff.rb
```diff
@@ -89,6 +91,10 @@ def value_for_charge(type)
 
   private
 
+  def tariff_holder_symbol
+    meters.any? ? :meter : tariff_holder_type.underscore.to_sym
```
iant, line 95: something similar to this would be useful in the editor bits i'm working on so i'm wondering if maybe this could be a public method?

e.g. I'm doing things like this in a few places
```
      case @energy_tariff.tariff_holder_type
      when 'School' then redirect_to school_energy_tariff_energy_tariff_flat_prices_path(@school, @energy_tariff)
      when 'SiteSettings' then redirect_to admin_settings_energy_tariff_energy_tariff_flat_prices_path(@energy_tariff)
      end
```

iant (approved):
looks good to me

PR #2947, opened 2023-08-03
Fix precision on tariff rate display

iant:


iant:
hi @ldodds I've added a new helper method (and some additional tests) so this can be reviewed/merged now

ldodds (approved):

PR #2948, opened 2023-08-07
Map all energy tariffs over to correct type of meter attributes

ldodds:
This PR ensures that all tariff data is correctly converted into meter attributes as required by the new `GenericTariffManager` in the analytics. See: https://github.com/Energy-Sparks/energy-sparks_analytics/pull/581

While the EnergyTariff model already has the code in place to create the right Hash structure that describes a tariff, we still needed to ensure that the tariffs at all levels (`Meter`, `School`, `SchoolGroup`, `SiteSettings`) are passed to the analytics. Prior to this only meter specific tariffs were being converted.

We also need to pass some additional configuration at the School level ("pseudo meter attributes") to provide default tariffs for creation of aggregated meters and specific types of solar meters.

So this PR:

* adds a new `EnergyTariffHolder` concern, to hold all the common code across tariff holders
* updates the `SiteSettings`, `SchoolGroup`, `School` models to add the concern and implement/override methods as required
* update the `GlobalMeterAttribute` model and `ParentAttributeHolder` concern to filter out older style tariff attributes when the `:use_new_energy_tariffs` feature is active
* updates the `Meter` model to include all meter specific and inherited tariffs to the analytics
* update the `School` model to provide the necessary pseudo meter attribute configuration. The need for this might later be removed

Specs have been added for all changes.

Some of this code can later be simplified once we switch to the new tariff system. For now have left some slightly verbose code in places to help with any debugging in next few weeks.


Gemfile
```diff
@@ -23,8 +23,8 @@ gem 'pg_search'
 gem 'calculate_in_group'
 
 # Dashboard analytics
-gem 'energy-sparks_analytics', git: 'https://github.com/Energy-Sparks/energy-sparks_analytics.git', tag: '2.9.1'
-#gem 'energy-sparks_analytics', git: 'https://github.com/Energy-Sparks/energy-sparks_analytics.git', branch: 'aws-eb-test'
+#gem 'energy-sparks_analytics', git: 'https://github.com/Energy-Sparks/energy-sparks_analytics.git', tag: '2.9.1'
+gem 'energy-sparks_analytics', git: 'https://github.com/Energy-Sparks/energy-sparks_analytics.git', branch: '3551-create-a-replacement-meter-tariff-manager'
```
iant, line 27: just to double check, does this need returning to the tagged release?

iant (approved):
:+1: looks good to me @ldodds a few comments.

Gemfile
ldodds: It will do, once we've merged in and tagged the analytics

PR #2949, opened 2023-08-07
Remove older version of tariff editor

ldodds:
Removes the code and tests for the older UserTariff based tariff editor. This has now been duplicated and replaced by the EnergyTariff editor.


app/models/meter.rb
```diff
@@ -211,11 +211,7 @@ def global_meter_attributes
 
   def all_meter_attributes
     meter_attributes_collection = global_meter_attributes + school_group_meter_attributes + school_meter_attributes + meter_attributes.active
-    meter_attributes_collection += if EnergySparks::FeatureFlags.active?(:use_new_energy_tariffs)
-                                     energy_tariff_meter_attributes
-                                   else
-                                     user_tariff_meter_attributes
-                                   end
+    meter_attributes_collection += energy_tariff_meter_attributes
     meter_attributes_collection
```
iant, line 215: I think this method is ok to just read like this now:
```
  def all_meter_attributes
    global_meter_attributes +
      school_group_meter_attributes +
        school_meter_attributes +
          meter_attributes.active +
            energy_tariff_meter_attributes
  end
```

iant (approved):
:+1: looks good me @ldodds 1 comment/suggestion on a simplification of a method's code

if there isn't already, I think we need an additional ticket/pr to update all references to `user_tariff` in the translation files e.g.
```
t('schools.user_tariffs.view_and_manage_tariffs_html', user_tariffs_path: school_energy_tariffs_path(school))
```
to
```
t('schools.energy_tariffs.view_and_manage_tariffs_html', energy_tariffs_path: school_energy_tariffs_path(school))
```

PR #2951, opened 2023-08-08
Return error if require param is missing

ldodds:
We seem to have crawlers or bots hitting the school group comparison page to download CSV, but not adding the parameter for the advice page key. This is adding a bunch of errors to Rollbar.

Update the controller plus add spec to return a 400 error if the parameter isn't replied. This should hopefully discourage them!


iant (approved):
:+1: looks good to me @ldodds

PR #2953, opened 2023-08-09
Create version of tariff editor to support school group tariffs

iant:
These tariffs can only be created or updated by ES admins or group admins.

We won’t make this editor available to group admins initially, so links to the new pages should be in the ES admin only section of the Manage Group menu

- [x] The tariffs will be linked to the School Group as the tariff holder
- [x] Only gas and electricity meter types will be supported. So no tariffs tied to specific meters or support for solar/exported solar.
- [x] The tariffs can have an option start or end time, but not both
- [x] The entry page should list the tariffs for the school group, allowing them to viewed, edited, etc.
- [x] For this version of the editor there will be no “Smart Meter” tariffs tab.
- [x] The editor will have a “Default tariffs” tab. It should list the system wide tariffs i.e. owned by the SiteSettings


app/views/school_groups/_sub_nav.html.erb
```diff
@@ -21,6 +21,7 @@
             <div class="dropdown-menu scrollable" aria-labelledby="manage-school-group" id="dropdown-manage-school-group">
               <%= link_to t('school_groups.sub_nav.chart_settings'), school_group_chart_updates_path(@school_group), class: 'dropdown-item' %>
               <%= link_to t('school_groups.sub_nav.clusters'), school_group_clusters_path(@school_group), class: 'dropdown-item' %>
+              <%= link_to t('manage_school_menu.manage_tariffs'), school_group_energy_tariffs_path(@school_group), class: 'dropdown-item' %>
```
ldodds, line 24: "We won’t make this editor available to group admins initially, so links to the new pages should be in the ES admin only section of the Manage Group menu". Can you move this into the admin section?

If this impacts the tests, then you could comment those out for now. I want to allow us to internally test this out before making it visible to users.

app/services/energy_tariff_default_prices_creator.rb
```diff
@@ -10,6 +10,7 @@ def process
 
     night_times = case @energy_tariff.tariff_holder_type
                   when 'School' then Economy7Times.times(@energy_tariff&.meters&.first&.mpan_mprn)
+                  when 'SchoolGroup' then TimeOfDay.new(0, 0)..TimeOfDay.new(7, 0) # Equivalent to Economy7Times::DEFAULT_TIMES
```
ldodds, line 13: Constant should be visible now in latest analytics. Could do this later as part of further improvements.

app/controllers/energy_tariffs/energy_tariffs_controller.rb
```diff
@@ -4,23 +4,30 @@ class EnergyTariffsController < ApplicationController
...
+      authorize! :manage, @school.energy_tariffs.build if @school
+      authorize! :manage, @school_group.energy_tariffs.build if @school_group
+
       if @school
         @electricity_meters = @school.meters.electricity
         @electricity_tariffs = @school.energy_tariffs.electricity.by_start_date.by_name
         @gas_meters = @school.meters.gas
         @gas_tariffs = @school.energy_tariffs.gas.by_start_date.by_name
       end
+      @site_setting = SiteSettings.current if @school_group
```
ldodds, line 23: With this set, then we end up seeing duplicate tariffs on the school group page:

```
    <%= render 'energy_tariffs_tab_school_groups' if @school_group %>
    <%= render 'energy_tariffs_tab_site_settings' if @site_setting %>
```

Means we see the school groups tariffs, then all of the site settings. The latter should only be in the default view.

app/views/energy_tariffs/energy_tariffs/_energy_tariffs_summary_table.html.erb
```diff
@@ -35,7 +35,7 @@
             <%= link_to t('common.labels.edit'), edit_school_energy_tariff_path(school, energy_tariff), class: 'btn' %>
             <%= link_to t('common.labels.delete'), school_energy_tariff_path(school, energy_tariff), method: :delete, data: {confirm: t('common.confirm') }, class: 'btn btn-danger' %>
           <% end %>
-          <% if energy_tariff.tariff_holder_type == 'SiteSettings' %>
+          <% if energy_tariff.tariff_holder_type == 'SiteSettings' && site_settings_page? %>
```
ldodds, line 38: There's no if statement checking for whether the tariff holder is a SchoolGroup which means there's no "Full details", "Edit" or "Delete" button, so I can't do anything with a tariff once created.

ldodds (changes requested):

app/views/school_groups/_sub_nav.html.erb
iant: done :+1: (see skipped test)

app/services/energy_tariff_default_prices_creator.rb
iant: :+1:

app/controllers/energy_tariffs/energy_tariffs_controller.rb
iant: :+1: done

ldodds (approved):

PR #2954, opened 2023-08-10
Tidy up routing on tariff editor

iant:


app/views/energy_tariffs/energy_tariffs/new.html.erb
```diff
@@ -4,30 +4,12 @@
...
-    <% end %>
-  <% end %>
-
-  <% if @energy_tariff.tariff_holder_type == 'SiteSettings' %>
-    <%= simple_form_for @energy_tariff, url: admin_settings_energy_tariffs_path(@energy_tariff) do |f| %>
-      <%= f.hidden_field :meter_type, value: @energy_tariff.meter_type %>
-      <%= f.hidden_field :tariff_type, value: @energy_tariff.tariff_type %>
-      <%= render 'form', school: @school, energy_tariff: @energy_tariff, f: f %>
+  <%= simple_form_for @energy_tariff.tariff_holder_route + [@energy_tariff] do |f| %>
+    <%= f.hidden_field :meter_type, value: @energy_tariff.meter_type if @energy_tariff.tariff_holder_type == 'School' %>
```
ldodds, line 8: meter type and tariff type are required for all tariffs, regardless of holder. Is there  a reason to only include the hidden fields for School tariffs? Just want to make sure we don't end up creating SchoolGroup and SiteSetting tariffs with the default values, rather than those set by the user.

app/controllers/energy_tariffs/energy_tariff_charges_controller.rb
```diff
@@ -22,11 +22,7 @@ def create
           @energy_tariff.energy_tariff_charges.destroy_all
           @energy_tariff_charges.each(&:save!)
         end
-        case @energy_tariff.tariff_holder_type
-        when 'School' then redirect_to school_energy_tariff_path(@school, @energy_tariff)
-        when 'SchoolGroup' then redirect_to school_group_energy_tariff_path(@school_group, @energy_tariff)
-        when 'SiteSettings' then redirect_to admin_settings_energy_tariff_path(@energy_tariff)
-        end
+        redirect_to polymorphic_path(@energy_tariff.tariff_holder_route + [@energy_tariff])
```
ldodds, line 25: This looks much neater! Looking across how you've used it, there seem to be the following varieties of call to polymorphic_path:

- just the tariff `... + [@energy_tariff]`
- with an extra path segment `... + [@energy_tariff, :energy_tariff_differential_prices]`
- with an id `... + [@energy_tariff, :energy_tariff_flat_price], id: @energy_tariff.energy_tariff_prices.first`
- with an action `action: 'edit'`
- with an action and an id

Wondering if you can wrap those all into an `energy_tariffs_path` helper, e.g:

```
def energy_tariffs_path(energy_tariff, path=[], options={})
  polymorphic_path(@energy_tariff.tariff_holder_route + [@energy_tariff] + path, options)
end
```

We do something similar for `advice_page_path`.

Then you could use:

```
energy_tariff_path(@energy_tariff, :energy_tariff_differential_prices)
energy_tariff_path(@energy_tariff, :energy_tariff_differential_price, id: 123)
energy_tariff_path(@energy_tariff, action: :choose_type)
```

app/views/energy_tariffs/energy_tariff_charges/_charge_vat.html.erb
```diff
@@ -10,7 +10,7 @@
       <%= t('schools.user_tariff_charges.charge_vat.rate') %>
     </td>
     <td class="value">
-      <%= f.select :vat_rate, options_for_select(['0%', '5%', '20%'], energy_tariff.vat_rate), {required: :true}, { class: 'form-control' } %>
+      <%= f.select :vat_rate, options_for_select(['0', '5', '20'], energy_tariff.vat_rate), {required: :true}, { class: 'form-control' } %>
```
ldodds, line 13: Is this an intentional change here? There's a ticket to review the VAT select to allow an empty option, but I think the labels should still be %?

ldodds (changes requested):

app/views/energy_tariffs/energy_tariffs/new.html.erb
iant: 👍 have removed the check for school

PR #2955, opened 2023-08-11
Admin report for school group tariffs

ldodds:
Creates a new report for admins that lists:

- which school groups have Energy Tariffs configured
- how many schools in those groups have their own school level tariffs
- how many schools in those groups have DCC tariffs

...along with links to more information where currently available. This is intended to give an overview of current state of tariffs. The individual group and school pages will provide more information.

The PR declares two new associations on EnergyTariffs which makes it possible to do:

```
EnergyTariff.joins(:school)
EnergyTariff.joins(:school_groups)
```

Which allows joining against those models when doing queries.


iant (approved):
:+1: looks good to me @ldodds

PR #2956, opened 2023-08-11
Allow meters to be optional on School tariff editor

ldodds:
Currently we require users to choose one or more meters when creating a School tariff.

But we want to allow the creation and editing of tariffs that apply to all meters of a type at a school, not just specific meters.

This PR updates the "choose meters" stage of tariff creation to allow user to decide whether to add meters.

- add a check box to the form to prompt user whether they want to add meters, default is now that a tariff is for any meter
- update validation check to be aware of users choice
- update specs to check for validation changes and selections
- add missing translation for error message and new translations for extra form text.


iant (approved):
:+1: looks good to me @ldodds

PR #2959, opened 2023-08-15
Improve layout of tariff page

iant:
- [x] Make the page title the name of the tariff
- [x] The text “Any changes to the tariff will be reflected in the electricity cost analysis and gas cost analysis pages which will update overnight.” should be conditional. E.g. only refer to gas analysis if this is a gas tariff
- [x] Add a “Dates” section, similar to the charges section which lists the start and end dates for the tariff as a separate rows. There should be an “Edit” button to allow the dates to be changed
- [x] Add a “Meters” section, similar to the other sections, which has a list of the meters associated with the tariff. This will either be a single row, e.g. “All electricity meters”, or one row per meter if there are meters associated. When we display meters it should show their MPAN and name. There should be an edit button to allow user to edit the meter list
- [x] Right align the “Edit” buttons

For Energy Sparks admins only:

- [x] Add a Notes section to the bottom of the page with the following rows: Date created, Created by (with user name); Date last updated, Last updated by (with user name)


app/views/energy_tariffs/energy_tariffs/show.html.erb
```diff
@@ -24,12 +83,49 @@
...
+      <div>
+        <h4><%= t('schools.user_tariffs.show.notes_html') %></h4>
+      </div>
+    </div>
+    <div>
+      <table class="table table-charges">
+        <tbody>
+          <tr>
+            <td class="description"><%= t('schools.user_tariffs.show.created_at') %></td>
+            <td class="value"><%= @energy_tariff.created_at&.to_s(:es_compact) %></td>
```
ldodds, line 109: This `es_compact` doesn't seem to work? Viewing a tariff I see: "2023-08-01 13:55:41 UTC" not `%d/%m/%Y %H:%M`

app/views/energy_tariffs/energy_tariffs/choose_meters.html.erb
```diff
@@ -21,16 +21,7 @@
...
-          <%= f.collection_check_boxes(:meter_ids, @meters, :id, :mpan_mprn) do |b|  %>
-            <div class="custom-control custom-checkbox">
-              <%= b.check_box(class: "custom-control-input") %>
-              <%= b.label(class: "custom-control-label spaced") do %>
-                <%= b.object.display_summary(display_data_source: false) %>
-              <% end %>
-            </div>
-          <% end %>
-        </div>
+        <%= render 'form_meter_choices', f: f %>
```
ldodds, line 24: My fault for making the card ambiguous, but rather than have a custom we should just send the user back to the standard meter choices form. As it should be possible to switch the tariff from being for a specific meter, to just all electricity/gas meters. There's a card for this so you could address there: https://trello.com/c/AYSxT3XY/3600-allow-meters-to-be-edited-on-school-tariffs

app/controllers/energy_tariffs/energy_tariffs_controller.rb
```diff
@@ -55,6 +55,35 @@ def choose_meters
...
+        redirect_to energy_tariffs_path(@energy_tariff)
+      else
+        render :edit_meters
+      end
+    end
+
+    def edit_dates
+    end
+
+    def update_dates
```
ldodds, line 79: Rather than a separate route, I'd just send the user back to the user title / date editing form. No need to exclude editing of titles, that's a useful feature in itself

ldodds (changes requested):

app/views/energy_tariffs/energy_tariffs/show.html.erb
```diff
@@ -1,20 +1,79 @@
...
     <br/>
   <% end %>
 
-  <%= render 'energy_tariffs/energy_tariffs/header', energy_tariff: @energy_tariff %>
+  <div class="d-flex justify-content-between">
+    <div>
+      <h4><%= t('schools.user_tariffs.show.dates.title') %></h4>
+    </div>
+    <div>
+      <%= link_to t('common.labels.edit'),
```
ldodds, line 17: Where we've allowed editing of dates and meters can you add a system spec. Could write a test that starts from having an existing saved tariff and then confirm you can a) edit meters (for schools only) and b) edit title/dates (should apply to all tariff holders)

ldodds (changes requested):

app/views/energy_tariffs/energy_tariffs/choose_meters.html.erb
iant:
```
  concern :tariff_holder do
    scope module: 'energy_tariffs' do
    resources :energy_tariffs do
      resources :energy_tariff_flat_prices
      resources :energy_tariff_differential_prices
      resources :energy_tariff_charges
      collection do
        get :choose_meters, to: 'energy_tariffs#choose_meters'
      end
      member do
        get :choose_type, to: 'energy_tariffs#choose_type'
        get :edit_meters, to: 'energy_tariffs#edit_meters'
        post :update_meters, to: 'energy_tariffs#update_meters'
      end
    end
  end
```

sending the user back to the standard meter choices form won't work.  `choose_meters` is a collection, not a resource (meters are selected prior to a new energy tariff being created).  so to edit the meters for an energy tariff requires a new member to be defined in the routes.

app/controllers/energy_tariffs/energy_tariffs_controller.rb
iant: :+1:

app/views/energy_tariffs/energy_tariffs/show.html.erb
iant: :+1: editing functionality tests added for a & b to the relevant shared examples.

iant:
@ldodds this one's ready for a re-review too now.

PR #2960, opened 2023-08-15
Fix additional issues with tariff editor

ldodds:
There were several issues with the tariff editor:

- Differential Prices weren't being added for School tariffs if there were no meters. Fixed the logic to use the defaults in that case
- There were a lot of partials that were called with a `school: @school` variable but the variable wasn't being used in the partials. So have removed that to avoid confusion and clarify the interface to the partials
- Updated the other tariff controllers to use `tariff_holder` as the instance name for the loaded model
- Fixed some redirect routes in controllers. One of which was stopping prices being edited for School Group tariffs
- Fix up the tariff show page to ensure that the Edit links for Prices and Charges are displayed for all tariff holder types (they were missing for School Group) and added a spec for this

### DIFF
--- a/app/controllers/energy_tariffs/energy_tariff_charges_controller.rb
+++ b/app/controllers/energy_tariffs/energy_tariff_charges_controller.rb
@@ -4,8 +4,8 @@ module EnergyTariffs
     include EnergyTariffable
     include EnergyTariffsHelper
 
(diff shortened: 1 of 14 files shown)


iant (approved):
:+1: looks good to me @ldodds

PR #2964, opened 2023-08-15
Improve energy tariff basic model validations

iant:


ldodds (changes requested):
Need to reenable all tests on the migrations service and resolve why they are failing

iant:
this one's ready for review when you get a mo @ldodds (updated validation for start date (if present) < end date)

ldodds (approved):

PR #2966, opened 2023-08-16
Revise table layout on energy tariff index

ldodds:
This PR revises how the tables of tariffs are displayed on the tariff index page tabs.

The current partials have been (mostly) replaced with new components which will make the UI easier to test and customise further.

List of key changes:

* Add `EnergyTariffTableComponent` which will format a list of tariffs into a consistent table, optionally with additional actions
* Add `EnergyTariffComponent` that will display a series of tariff tables for a tariff holder. This uses the above component and now handles most of the generation of the tariff, default tariff and smart meter tabs
* Remove the older custom code for displaying the smart meter tariffs, so that we use the above component
* Reorder the table columns
* Make the table columns sortable
* Apply a custom sort when displaying tariffs so that a nil start date sorts first.
* Various labelling and text updates
* Fixes a bug with not displaying the right message when there are no tariffs for a meter type


iant (approved):
:+1: looks good to me @ldodds

PR #2967, opened 2023-08-17
Add breadcrumbs to energy tariff pages

iant:


ldodds (changes requested):
Overall this looks great. There's 2 minor issues though.

Firstly there's some extra whitespace above the breadcrumbs. If you compare group homepage with the tariff page, or school dashboard with their tariffs, then there's extra space. I think he had the same issue when we introducing the breadcrumbs on the school group page. Can you fix that?

And secondly, the "Pupil Dashboard" and/or "Adult Dashboard" links don't appear on the school navbar. The buttons and the manage menu is there, but not those links. Ideally the display should be consistent. Can you look to see what's missing?

iant:
@ldodds just pushed a fix for those 2 issues so pr's ready for another review when you get a mo,.

ldodds (approved):

PR #2969, opened 2023-08-17
Allow only limited editing of smart meter tariffs

ldodds:


iant (approved):
:+1: looks good to me @ldodds

PR #2970, opened 2023-08-17
Allow meters to be edited on School tariffs (2)

iant:
- [x] don't allow/show meter edit functionality if the energy tariff has a source of dcc


iant:
closing in favour of https://github.com/Energy-Sparks/energy-sparks/pull/2969

PR #2971, opened 2023-08-17
Improve entry and validation of time periods for differential tariffs

iant:
- [x] Allow user to delete or add prices for a differential tariff
- [x] When editing an existing price or adding a new one, the form should validate that its doesn’t overlap with an existing one
- [x] The prices should be labelled based on their time ranges, rather than “Daytime”/”Night time”
- [x] If the current set of prices don’t provide a continuous cover across the day, then display a red error message on the editor (as per a validation error) to indicate that the prices are not complete and prompt the user to add another price
- [x] If the current set of prices don’t provide a continuous cover across the day, then the tariff shouldn’t be considered “complete”. Currently we only provide .complete tariffs to the analytics. This will stop validation errors in the analytics
- [x] If there are zero prices, for a differential tariff, then do the same as if there isn’t continuous cover, but display a specific error message, e.g. “A differential tariff must have at least 2 prices, e.g. a day time and a night-time rate. Please add prices, or reset to default”. The “reset to default” should refresh the page, resetting the prices to the default Economy 7 time periods.
- [x] Move analytics economy 7 times class over to the application 

See also: https://github.com/Energy-Sparks/energy-sparks_analytics/pull/591 (remove economy 7 times class from analytics)

- [x] its should be possible to "reset to default" if you have incomplete ranges, not just if there are no ranges.


ldodds (changes requested):
There's a requirement which hasn't been implemented: "If the current set of prices don’t provide a continuous cover across the day, then the tariff shouldn’t be considered “complete”. Currently we only provide .complete tariffs to the analytics. This will stop validation errors in the analytics"

This was checked off on the PR description but there's no code to do that, that I can see? I've confirmed that deleting a range so that the 24 hour coverage is incomplete breaks the regeneration process. 

You had also suggested that the form to add a range could default to the missing times? This would be helpful to add, as discussed.

Also, I think its should be possible to "reset to default" if you have incomplete ranges, not just if there are no ranges.

iant:
@ldodds i think this one is ready for review again.  there's also a query in the ticket that we can potentially swap out for the usable method but for now everything's in ruby.

ldodds (approved):

PR #2973, opened 2023-08-18
Improve error reporting on tariff forms

ldodds:
Improves the display of errors on the flat rate price and charges forms.

Originally the errors were just using a default error block. I've updated the form fields to check for errors and then provide a custom error message via the SimpleForm `error` attribute on the field.

I've also fixed an issue with form values not displaying properly if they don't match the underlying type. See the `value_allowing_for_errors` helper. This was mainly to catch problems with a user typing a string into the field.

However I also noticed that while we were adding `pattern` to the form fields they were being ignored in some cases. So I've tweaked the fields to be numeric which means you can no longer enter anything except a number. 

So the `value_allowing_for_errors` helper is probably not needed, but have left it in place for now.

The form validation on the editing to name/dates and on the differential prices was already fine.

### DIFF
--- a/app/helpers/energy_tariffs_helper.rb
+++ b/app/helpers/energy_tariffs_helper.rb
@@ -100,6 +100,17 @@ module EnergyTariffsHelper
     settings(charge_type).fetch(:units, []).map { |k| [EnergyTariffCharge.charge_type_units[k], k] }
   end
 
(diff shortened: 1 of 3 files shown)


iant (approved):
:+1: looks good to me @ldodds

PR #2974, opened 2023-08-18
Add option to show prices

ldodds:
When people are reviewing and editing tariffs the most common things to do will be to check dates and prices.

This PR updates the tariff table component to:

- include consumption charges column (but these can be switched off as an option)
- remove the "View details" link in favour of making the tariff name a link
- tidy up the table alignment by adding some `col-x` and text alignment classes

### DIFF
--- a/app/components/energy_tariff_table_component.rb
+++ b/app/components/energy_tariff_table_component.rb
@@ -5,39 +5,20 @@ class EnergyTariffTableComponent < ViewComponent::Base
 
   delegate :can?, :cannot?, to: :helpers
 
-  def initialize(id: 'tariff-table', tariff_holder:, tariffs:, show_actions: true)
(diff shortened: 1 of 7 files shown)


app/components/energy_tariff_table_component.rb
```diff
@@ -5,17 +5,44 @@ class EnergyTariffTableComponent < ViewComponent::Base
...
+  def show_actions?
+    @show_actions
+  end
+
+  def start_time(price)
+    price.start_time.to_s(:time)
+  end
+
+  def end_time(price)
+    price.start_time.to_s(:time)
```
iant, line 43: should this be `price.end_time.to_s(:time)` (end time instead of start time)

iant (approved):
:+1: looks good to me @ldodds .  one question about if something should be end time instead of start time?

app/components/energy_tariff_table_component.rb
ldodds: d'oh

PR #2975, opened 2023-08-18
Add a note if there are no standing charges

ldodds:
Adds a note to to say if there's no standing charges configured, rather than leaving an empty space.

Also adds a proper page title and replaces use of `<br/>` with bootstrap spacers.

### DIFF
--- a/app/views/energy_tariffs/energy_tariffs/_charges_table.html.erb
+++ b/app/views/energy_tariffs/energy_tariffs/_charges_table.html.erb
@@ -1,30 +1,37 @@
 <table class="table table-charges">
   <tbody>
-  <% if energy_tariff.energy_tariff_charges.any? %>
-    <% energy_tariff.energy_tariff_charges.each do |energy_tariff_charge| %>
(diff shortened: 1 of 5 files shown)


app/views/energy_tariffs/energy_tariffs/_charges_table.html.erb
```diff
@@ -1,30 +1,37 @@
 <table class="table table-charges">
   <tbody>
-  <% if energy_tariff.energy_tariff_charges.any? %>
-    <% energy_tariff.energy_tariff_charges.each do |energy_tariff_charge| %>
+    <% if energy_tariff.has_any_standing_charges? %>
+      <% if energy_tariff.energy_tariff_charges.any? %>
```
iant, line 4: I'm not sure this is needed here?  `energy_tariff.energy_tariff_charges.each` should just skip if empty?

iant (approved):
looks good to me @ldodds 1 comment on if `energy_tariff.energy_tariff_charges.any?` is necessary

PR #2977, opened 2023-08-18
Allow user to change tariff to all meters

ldodds:
Revises the "edit meters" flow so that it uses the same form as the "choose meters" step. I've retained the routes and form due to the different contexts (with and without a tariff).


iant (approved):
:+1: looks good to me

PR #2978, opened 2023-08-18
Transifex synch 2023-08-16

ldodds:
Did a push and pull of latest translations in Transifex.

This looks like it has the updated text for the school group dashboards.


iant (approved):
:+1:

PR #2979, opened 2023-08-18
Add I18N key constants

ldodds:
We have a number of dependabot PRs which are all failing with the same error:

https://github.com/Energy-Sparks/energy-sparks/actions/runs/5713575489/job/15900364828?pr=2939

```
       +"Translation missing: en.analytics.series_data_manager.series_name.Electricity consumed from mains" => "#007EFF",
       +"Translation missing: en.analytics.series_data_manager.series_name.Exported solar electricity (not consumed onsite)" =>        -"translation missing: en.analytics.series_data_manager.series_name.Electricity consumed from mains" => "#007EFF",
       -"translation missing: en.analytics.series_data_manager.series_name.Exported solar electricity (not consumed onsite)" =>      # ./spec/models/chart_data_values_spec.rb:155:in `block (3 levels) in <top (required)>'
```

Looks like somewhere in the recent dependency updates the formatting of the "Translation missing" text has changed. 

I'm not sure why we are checking for the translation missing text here in the spec, as it's a bug if the translation can't be found. Fix is to use the constants defined in the `SolarPVPanels` class. Maybe these were missed before.

This wasn't causing a visible application error as the missing colour code meant the charts have just been defaulting to a different colour. This fix will now apply the preferred colour to the exported solar series.

If we can merge this then we can get dependabot to rebase the other changes so we can merge.

### DIFF
--- a/app/models/chart_data_values.rb
+++ b/app/models/chart_data_values.rb
@@ -141,8 +141,8 @@ class ChartDataValues
       I18n.t('analytics.series_data_manager.series_name.storage_heaters') => STORAGE_HEATER,
       '£' => MONEY,
       I18n.t("analytics.series_data_manager.series_name.#{SolarPVPanels::SOLAR_PV_ONSITE_ELECTRIC_CONSUMPTION_METER_NAME_I18N_KEY}") => GREEN,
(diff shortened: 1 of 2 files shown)


iant (approved):
:+1:

PR #2985, opened 2023-08-21
Fix error when reporting tariffs have changed

ldodds:
Update to new version of analytics with fix to tariff manager that stops an error on pages where we display the "tariff  caveats".


iant (approved):
:+1:

PR #2986, opened 2023-08-21
Remove duplicate dcc tariffs and charges

ldodds:
Removes duplicate TariffPrice and TariffStandingCharge records if they are the same as the previous record for that meter.

Also fixes a bug in display of migrated smart meter tariffs


iant (approved):
:+1: looks good to me @ldodds

PR #2987, opened 2023-08-21
3595 change smart meter tariff imports to populate new energytariff model

ldodds:
This PR changes the n3rgy tariff loader to populate the new EnergyTariff model instead.

Reviewing the existing code I noticed two issues:

* Standing charges were not being displayed on the charge forms, so there's an update to include these
* The tariff price migration code was setting an `end_date` on tariffs which should be open ended, this has been fixed

The rest of the changes are to:

* Introduce a new `N3rgyEnergyTariffLoader` to coordinate fetching tariffs each day, then inserting new ones as required
* Introduce a new `N3rgyEnergyTariffInserter` class which handles the inserting of new energy tariffs

This new code is switched in via a feature flag in the rake task.

The bulk of the code and tests are around the second class. Specifically:

* Throwing an error if we encounter a tariff which isn't a simple flat rate or differential tariff
* If we receive a tariff which has zero or negative prices or charges, we do not insert them and also update the existing tariff to add an end date. This seems to happen if the smart meter no longer has tariff information, so we can't assume the existing tariff is still correct
* Check whether the tariff information we've just fetched from n3rgy is different to any existing tariff we have, checking for: new standing charges, new prices, new differential periods or a switch between tariff types

The end result should be that we stop inserting duplicate tariffs, invalid tariffs are ignored (but logged) and the `EnergyTariff` model is properly updated.

### DIFF
--- a/app/services/amr/n3rgy_energy_tariff_inserter.rb
+++ b/app/services/amr/n3rgy_energy_tariff_inserter.rb
@@ -20,8 +20,7 @@ module Amr
     def perform
       return if reject_as_zero_standing_charges?
       return if reject_as_zero_tariffs?
-      return if unexpected_tariff_format?
-      return if summary_tariff.nil?
+      check_unexpected_tariff_format?
 
       energy_tariff = latest_energy_tariff
       if energy_tariff.nil? || tariff_changed?(energy_tariff)
@@ -96,13 +95,13 @@ module Amr
     end
 
     #We only support flat rate and differential tariffs in the EnergyTariff
-    #model currently.
-    def unexpected_tariff_format?
+    #model currently. Raise exception to catch problems early
+    def check_unexpected_tariff_format?
       unless rates.values.all? {|price| price.is_a?(Numeric)}
-        log_error("Unexpected tariff format #{rates.inspect}")
-        return true
+        raise "Unexpected tariff format for #{@meter.mpan_mprn} on #{@start_date}: #{rates.inspect}"
       end
-      false
+      #Trigger parsing of tariff data, which may throw errors
+      summary_tariff
     end
 
     #Returns structures like:
@@ -126,9 +125,6 @@ module Amr
 
     def summary_tariff
       @summary_tariff ||= N3rgyTariffs.new(@tariff).parameterise
-    rescue => e
-      log_error("Error parsing tariff from n3rgy #{e.message}")
-      nil
     end
 
     def end_date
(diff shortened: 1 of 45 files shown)


app/services/amr/n3rgy_energy_tariff_inserter.rb
```diff
@@ -0,0 +1,194 @@
...
+      if energy_tariff.flat_rate?
+        return energy_tariff.energy_tariff_prices.first.value == rates.values.first
+      else
+        return false unless energy_tariff.energy_tariff_prices.count == rates.keys.count
+        rates.each do |times, price|
+          return false unless energy_tariff.energy_tariff_prices.where(start_time: times.first.to_s, end_time: times.last.to_s, value: price).any?
+        end
+      end
+      true
+    end
```
iant, line 73: perhaps use some early returns here?
```
    def same_prices?(energy_tariff)
      return (energy_tariff.energy_tariff_prices.first.value == rates.values.first) if energy_tariff.flat_rate?
      return false unless energy_tariff.energy_tariff_prices.count == rates.keys.count

      rates.each do |times, price|
        return false unless energy_tariff.energy_tariff_prices.where(start_time: times.first.to_s, end_time: times.last.to_s, value: price).any?
      end

      true
    end
```

app/services/amr/n3rgy_energy_tariff_inserter.rb
```diff
@@ -0,0 +1,194 @@
...
+      summary_tariff[:standing_charges].values.first
+    end
+
+    def reject_as_zero_standing_charges?
+      if standing_charge <= 0.0
+        log_error("Standing charge returned from n3rgy for #{@start_date} are zero #{standing_charge}")
+        return true
+      end
+      false
+    end
```
iant, line 89: maybe this could be rewritten to avoid the return? (and similar for `reject_as_zero_tariffs?` below)
```
    def reject_as_zero_standing_charges?
      if standing_charge <= 0.0
        log_error("Standing charge returned from n3rgy for #{@start_date} are zero #{standing_charge}")
        true
      else
        false
      end
    end
```

iant: or even simpler?
```
    def reject_as_zero_standing_charges?
      return false unless standing_charge <= 0.0

      log_error("Standing charge returned from n3rgy for #{@start_date} are zero #{standing_charge}")
      true
    end
```

PR #2988, opened 2023-08-22
Allow nil values when a tariff price is created

iant:


app/views/energy_tariffs/energy_tariff_differential_prices/index.html.erb
```diff
@@ -32,6 +32,11 @@
           <%= t('schools.user_tariff_differential_prices.index.incomplete_message') %>.
         <% end %>
       <% end %>
+      <% if @energy_tariff.energy_tariff_prices.invalid_prices? %>
```
ldodds, line 35: Can I suggest a display tweak here, currently if there's missing prices you'll get a positive notice about full 24 hour coverage, but an error about missing prices.

I think it might be better to only show the positive notice (that prices and coverage are correct) if all the negatives have been addressed.

config/locales/views/schools/user_tariffs.yml
```diff
@@ -49,6 +49,7 @@ en:
         introduction_school: Using your bill, please provide as much detail as you can about the consumption charges associated with this tariff
         introduction_school_group: Using your bill, please provide as much detail as you can about the consumption charges associated with this tariff
         introduction_site_settings: Please provide as much detail as you can about the consumption charges associated with this tariff
+        invalid_values_message: You have one or more invalid prices. Please add valid prices for all rates marked
```
ldodds, line 52: Maybe shorten this to "Please add valid prices for all marked rates", as this works slightly better when visiting the page initially as the user hasn't entered any prices yet.

ldodds (changes requested):

iant:
@ldodds updates pushed for requested changes.  i think ok to re-review 
Screenshot 2023-08-23 at 11 41 22

ldodds (approved):

PR #2990, opened 2023-08-23
Add visual cues to tariffs table

ldodds:
If a tariff is disabled then

* Flag the table row in a different colour
* Do not include it in the list of default tariffs

If a tariff is not usable

* Flag the table row in a different colour
* Add a notice to the tariff page with a link to edit the prices


iant (approved):
:+1: looks good to me

PR #2994, opened 2023-08-23
Allow tariffs dates to be empty

ldodds:
The DateTimePicker component currently defaults to today's date if a value isn't provided.

This isn't the intended behaviour when editing tariffs, where an empty date should remain unpopulated. Rather than set a default in the form I've made the default value configurable on the component.

This preserves behaviour elsewhere in the application but allows an empty value to be provided on the tariff form.


iant (approved):
:+1:

PR #2996, opened 2023-08-23
Sort out display of units on flat rate prices

ldodds:
Display the form field as: "Rate in £ [...] per kWh" to match the standing charges, etc.


iant (approved):
:+1: looks good to me.  one comment

PR #2997, opened 2023-08-23
Disable standing charge form fields for smart meter tariffs

ldodds:
Smart meter tariffs, when imported have prices and a single type of standing charge.

We want to allow users to edit the other charges associated with the tariff but not the prices or that standing charge.

This PR fixes a bug with editing charges by disabling the form field for the `:standing_charge` `charge_type` but allows others types of charges to be updated.


iant (approved):
:+1:

PR #3000, opened 2023-08-24
Remove use of previous next buttons

ldodds:
This is a large refactor of the editor flow and the associated specs.

The key changes are to:

* Remove the "wizard" style interface with previous and next buttons in favour of providing an initial form, then allowing the user to navigate from the energy tariff page to add specific elements
* Use this to refactor the specs to allow more reuse and make it easier to test specific parts of the editor interface
* Allow the type of tariff to be changed, e.g. switch between flat rate and differential, although this requires removing prices

Along the way I also fixed an issue with the EnergyTariff.last_updated_by not updating if we just edit tariffs and charges.

Following this, there's a need to do UX tidy, but this is the basic changes.

### DIFF
--- a/app/views/energy_tariffs/energy_tariffs/show.html.erb
+++ b/app/views/energy_tariffs/energy_tariffs/show.html.erb
@@ -102,7 +102,7 @@
     <tbody>
       <tr>
         <td class="description"><%= t('schools.user_tariffs.tariff_partial.type') %></td>
-        <td class="value"><%= @energy_tariff.flat_rate? ? t('schools.user_tariffs.tariff_partial.simple_tariff') : t('schools.user_tariffs.tariff_partial.differential_tariff') %></td>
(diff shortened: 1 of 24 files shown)


app/views/energy_tariffs/energy_tariffs/show.html.erb
```diff
@@ -87,14 +87,35 @@
...
+  <table class="table table-charges">
+    <tbody>
+      <tr>
+        <td class="description"><%= t('schools.user_tariffs.tariff_partial.type') %></td>
+        <td class="value"><%= @energy_tariff.flat_rate? ? t('schools.user_tariffs.tariff_partial.flat_rate_tariff') : t('schools.user_tariffs.tariff_partial.differential_tariff') %></td>
+      </tr>
+    </tbody>
+  </table>
+
+
```
iant, line 110: remove a space?

iant (approved):
:+1: looks good to me

PR #3002, opened 2023-08-24
Tariff page ux improvements

ldodds:
Some smaller UX improvements to the tariff page

- introduced a component to handle all the page sections. I found these to be difficult to manage
- remove edit button from top of page, in favour of section specific buttons and feedback
- tweak labels to say a tariff is "not finished" rather than "not yet usable" and change edit button to "Finish" if the tariff isn't usable
- added colours to rows on tariff page to indicate if data is OK, needs attention or optional (blue)

There's some more improvements we could make, but this might be a helpful start?

### DIFF
--- a/app/assets/javascripts/form_wizard.js
+++ b/app/assets/javascripts/form_wizard.js
@@ -22,8 +22,30 @@ $(document).ready(function() {
     }
   });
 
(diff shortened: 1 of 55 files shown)


iant (approved):
👍 looks good to me @ldodds will merge

PR #3016, opened 2023-08-29
Allow admins to access tariff editor for their group

iant:


iant:
@ldodds there were but afaics they've been refactored away.  the tariff editing abilities as a group admin user is covered here https://github.com/Energy-Sparks/energy-sparks/blob/master/spec/system/school_groups/energy_tariffs_spec.rb#L22

and sub-nav display for non group admins here
https://github.com/Energy-Sparks/energy-sparks/blob/master/spec/support/school_groups_shared_examples.rb#L162

so i think the changes have fairly good test coverage.

ldodds (approved):

PR #3023, opened 2023-08-31
Show system default tariffs if no school group tariffs

iant:


ldodds (changes requested):
Change looks OK, but can you add a spec to the component?

iant:
@ldodds test added so i think this one's ready for re-review

ldodds (approved):

PR #3031, opened 2023-09-04
Update existing SiteSettings record

ldodds:
The admin site settings controller currently creates a new record when settings are updated, rather than updating the existing record. This works because we use `SiteSettings.current` to access the settings, so older records are ignored.

However, now that the energy tariffs are linked to the SiteSettings record creating a new record means we lose the association with tariffs, which causes a major problem with regenerating schools.

I think the intention is that the SiteSettings is meant to be a "singleton" rather than maintaining a history (which we don't need as we don't display or report on the history). So I've updated the controller to do a more conventional update to the existing record rather than creating a new one.

The PR has a small change to the controller and revisions to the system spec to explicitly test for updates when there is an existing record.

### DIFF
--- a/app/controllers/admin/settings_controller.rb
+++ b/app/controllers/admin/settings_controller.rb
@@ -1,12 +1,13 @@
 module Admin
   class SettingsController < AdminController
+    before_action :load_site_settings
+
     def show
-      @settings = SiteSettings.current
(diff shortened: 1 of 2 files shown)


urbanwide (approved):
Looks good @ldodds 👍

PR #3032, opened 2023-09-04
Remove reference to deleted service in after party task

urbanwide:
Database::EnergyTariffMigrationService.migrate_user_tariffs was removed in this PR: https://github.com/Energy-Sparks/energy-sparks/pull/3025, however it is still referenced by 20230725132841_migrate_to_energy_tariffs.rake which I've just been trying to run.

I have removed the reference for now as it was tricky to unpickle otherwise (think I would have needed to rollback irreversible migrations). Is really only is me that it will affect for now, just means I could run all the other after_party tasks.

I might need to grab a database dump at some point though with the migrated user tariffs in, assuming that's what it does.


ldodds (approved):

urbanwide:
Thanks @ldodds !

PR #3036, opened 2023-09-05
Create service for finding tariffs between dates

ldodds:
This is still a work in progress.

At various places across the site we want to display what tariff information has been used between two dates.

Currently this information is provided by the analytics, which requires that we create (or fetch from cache) a meter collection. This isn't necessary or desirable for all pages.

This service is the start of implementing an alternative interface for fetching this data directly from the database.

The analytics currently handles the tariff lookup by finding what tariff applies on every day between two dates, then combining the ranges. In this implementation I'm trying to avoid this by:

- querying the database to just fetch the tariffs that might apply within the range
- then doing further filtering in the service

What makes this complex is that:

- there is a hierarchy of tariffs to traverse
- tariffs might overlap
- there may be "gap" in the date ranges of tariffs which means we need to use a selection of meter, school, school group and system wide tariffs in order to find the applicable tariffs

What works so far:

- finding tariffs for simple ranges
- finding a mixture of tariffs, some of which are inherited
- finding tariff that are overlapping
- finding tariffs with "gaps"

What doesn't work currently:

- finding tariffs where the inherited ranges have gaps in them, currently its returning some overlapping ranges for some scenarios. These may not arise in practice, but want to make the service handle all cases if possible so we don't have discrepancies between the analytics and the application.

Also needs:

- removal of debug code
- a proper refactor
- some additional test cases being implemented
- 

This could be simplified to iterate through every day as there will be a limited number of tariffs at any time, but if we can do this with less CPU intensive checks, that would be helpful.


ldodds:
Closing, will revert to using existing approach when this is picked up again.

PR #3037, opened 2023-09-05
Add an applies to flag to Energy Tariff

iant:


iant:
hi @urbanwide this one's ready for review but just waiting on confirmation if i need an additional migration/after party task to re-set all meters to false (see details in ticket).  might do that in a follow up pr tho.

PR #3041, opened 2023-09-07
Filter energy tariffs by half-hourly flag not just meter type

iant:


iant:
going to close this as half hourly has now been removed.

PR #3042, opened 2023-09-07
Add an meter system options to Energy Tariff

iant:
- [x] Removes the old `half_hourly` flag from `Meter`
- [x] Adds a new `meter_system` integer/enum with possible values `HH` & `NHH AMR`,`NHH`, `SMETS2/smart` that defaults with `NHH AMR`
- [x] Replace `half_hourly` (Yes/No) with the meter_system value on the Admin Meter show page.
- [x] Replace the `half_hourly` column on the school group meter report with the meter_system value
- [x] Replace the `half_hourly` column on the data source report with the meter_system value
- [x] Replace the `half_hourly` column on the procurement route report with the meter_system value
- [x] Replace the `half_hourly` column on the daily imports email and its attachment with the meter_system value

### DIFF
--- a/app/controllers/schools/meters_controller.rb
+++ b/app/controllers/schools/meters_controller.rb
@@ -113,7 +113,7 @@ module Schools
     end
 
     def meter_params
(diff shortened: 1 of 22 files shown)


urbanwide (approved):
Looks good @iant 👍

PR #3044, opened 2023-09-08
Filter energy tariffs by meter system, not just meter type

iant:


spec/models/concerns/energy_tariff_holder_spec.rb
```diff
@@ -60,6 +62,36 @@
...
+
+        expect(tariff_holder.energy_tariffs.both.count).to eq(2)
+
+        expect(tariff_holder.energy_tariffs.electricity.both.count).to eq(1)
+        expect(tariff_holder.energy_tariffs.electricity.half_hourly.count).to eq(1)
+        expect(tariff_holder.energy_tariffs.electricity.non_half_hourly.count).to eq(1)
+
+        expect(tariff_holder.energy_tariffs.gas.both.count).to eq(1)
+        expect(tariff_holder.energy_tariffs.gas.half_hourly.count).to eq(0)
+        expect(tariff_holder.energy_tariffs.gas.non_half_hourly.count).to eq(0)
```
ldodds, line 76: These aren't actually testing anything, they're just asserting that the dummy data exists? So am not sure they're adding anything.

spec/models/concerns/energy_tariff_holder_spec.rb
```diff
@@ -40,13 +40,15 @@
...
     end
 
     context 'with School' do
-      let!(:site_settings)   { SiteSettings.current }
-      let(:school_group)     { nil }
-      let(:tariff_holder)    { create(:school, school_group: school_group) }
+      let!(:site_settings) { SiteSettings.current }
+      let(:school_group) { nil }
+      let(:tariff_holder) { create(:school, school_group: school_group) }
+      let!(:energy_tariff_electricty_half_hourly) { create(:energy_tariff, :with_flat_price, tariff_holder: tariff_holder, meter_type: "electricity", applies_to: "half_hourly", name: 'Electricity Tariff half_hourly') }
```
ldodds, line 49: Because these three tariffs are only used in one of the specs, its clear if you create a separate context.

E.g.

```
context 'when filtering tariffs' do
 let!(:energy_tariff_electricty_half_hourly) { create(:energy_tariff, :with_flat_price, tariff_holder: tariff_holder, meter_type: "electricity", applies_to: "half_hourly", name: 'Electricity Tariff half_hourly') }
 ...
end
```

spec/models/concerns/energy_tariff_holder_spec.rb
```diff
@@ -60,6 +62,36 @@
...
+        expect(tariff_holder.energy_tariff_meter_attributes(['electricity'], :both).map { |m| m.input_data['name'] }.sort).to eq(["Electricity tariff both"])
+        expect(tariff_holder.energy_tariff_meter_attributes(['electricity'], :half_hourly).map { |m| m.input_data['name'] }.sort).to eq(["Electricity Tariff half_hourly", "Electricity tariff both"])
+        expect(tariff_holder.energy_tariff_meter_attributes(['electricity'], :non_half_hourly).map { |m| m.input_data['name'] }.sort).to eq(["Electricity Tariff non_half_hourly", "Electricity tariff both"])
+
+        expect(tariff_holder.energy_tariff_meter_attributes(['gas']).map { |m| m.input_data['name'] }.sort).to eq(["Gas Tariff both"])
+        expect(tariff_holder.energy_tariff_meter_attributes(['gas'], :both).map { |m| m.input_data['name'] }.sort).to eq(["Gas Tariff both"])
+        expect(tariff_holder.energy_tariff_meter_attributes(['gas'], :half_hourly).map { |m| m.input_data['name'] }.sort).to eq(["Gas Tariff both"])
+        expect(tariff_holder.energy_tariff_meter_attributes(['gas'], :non_half_hourly).map { |m| m.input_data['name'] }.sort).to eq(["Gas Tariff both"])
```
ldodds, line 92: If you add a context as I suggest above, then you can decompose this one large spec into several smaller ones, e.g.:

```
it "defaults to both when no meter system specified"
it "returns both and half_hourly tariffs when filtering for half_hourly"
it "returns both and non_half_hourly tariffs when filtering for non_half_hourly"
it "ignores meter system when requesting gas tariffs"
...etc
```

This makes it much easier to read the overall rspec file and better documents what each set of assertions is doing.

spec/models/concerns/energy_tariff_holder_spec.rb
```diff
@@ -60,6 +62,36 @@
...
+        expect(tariff_holder.energy_tariffs.electricity.non_half_hourly.count).to eq(1)
+
+        expect(tariff_holder.energy_tariffs.gas.both.count).to eq(1)
+        expect(tariff_holder.energy_tariffs.gas.half_hourly.count).to eq(0)
+        expect(tariff_holder.energy_tariffs.gas.non_half_hourly.count).to eq(0)
+
+        expect(tariff_holder.energy_tariff_meter_attributes.map { |m| m.input_data['name'] }.sort).to eq(["Electricity tariff both", "Gas Tariff both"])
+        expect(tariff_holder.energy_tariff_meter_attributes(['electricity', 'gas']).map { |m| m.input_data['name'] }.sort).to eq(["Electricity tariff both", "Gas Tariff both"])
+        expect(tariff_holder.energy_tariff_meter_attributes(['electricity', 'gas'], :both).map { |m| m.input_data['name'] }.sort).to eq(["Electricity tariff both", "Gas Tariff both"])
+        expect(tariff_holder.energy_tariff_meter_attributes(['electricity', 'gas'], :half_hourly).map { |m| m.input_data['name'] }.sort).to eq(["Electricity Tariff half_hourly", "Electricity tariff both", "Gas Tariff both"])
```
ldodds, line 81: Better to use `match_array` unless the ordering is guaranteed/important

spec/models/concerns/energy_tariff_holder_spec.rb
```diff
@@ -40,13 +40,15 @@
...
 
     context 'with School' do
-      let!(:site_settings)   { SiteSettings.current }
-      let(:school_group)     { nil }
-      let(:tariff_holder)    { create(:school, school_group: school_group) }
+      let!(:site_settings) { SiteSettings.current }
+      let(:school_group) { nil }
+      let(:tariff_holder) { create(:school, school_group: school_group) }
+      let!(:energy_tariff_electricty_half_hourly) { create(:energy_tariff, :with_flat_price, tariff_holder: tariff_holder, meter_type: "electricity", applies_to: "half_hourly", name: 'Electricity Tariff half_hourly') }
+      let!(:energy_tariff_electricty_non_half_hourly) { create(:energy_tariff, :with_flat_price, tariff_holder: tariff_holder, meter_type: "electricity", applies_to: "non_half_hourly", name: 'Electricity Tariff non_half_hourly') }
```
ldodds, line 50: If you extract the tariff labels into variables, then you can extract that from the specs below, making them a bit clearer.

app/models/meter.rb
```diff
@@ -239,13 +239,23 @@ def active_meter_attributes
   def energy_tariff_meter_attributes
```
ldodds, line 239: This overrides the default implementation in energy tariff holder to do something slightly different. These are specs for this in meter spec. As you've modified this method, I think you also need to update the specs there to confirm that the `applies_to_meter_system` returns the right values and is used in this method correctly.

iant:
@ldodds updated tests added so I think this might be ready for a re-review now.

iant:
converting back to draft to add a bit more test coverage

ldodds (approved):

PR #3045, opened 2023-09-08
Change bonus points score to be inside a green pill on programme pages

urbanwide:


iant (approved):

PR #3049, opened 2023-09-11
Reduce clutter on adult dashboard

urbanwide:
Logic for training prompt should be: if we are showing the data enabled features (and so its worth the school attending training) then prompt the user to sign up if they confirmed their account in the last 60 days. Currently it always seems to be displaying? Also suggest we reduce the window for the training prompt to 30 days.

Refactored the rspec after removing the standard prompts spec - now also includes spec for school dashboard messages (as well as school group dashboard messages), along with spec for the training prompt.


ldodds:
@urbanwide will add a note to the trello card but it looks like you could also remove a few of those unused partials now too, to tidy up?

urbanwide:
Yes no probs - sorry missed those!

urbanwide:
Should be sorted in here, assuming tests pass: https://github.com/Energy-Sparks/energy-sparks/pull/3063

PR #3056, opened 2023-09-14
Automate reminders to review and update tariffs

iant:


iant:
Waiting on confirmation of email content (and a few other things; see ticket) so adding a blocked label to this for now.

.ebextensions/cronjob.config
```diff
@@ -286,6 +286,27 @@ files:
...
 
+    #
+    # Send Review Group Tariffs Reminders
+    #
+    "/etc/cron.d/send_review_group_tariffs_reminders":
+        mode: "000644"
+        owner: root
+        group: root
+        content: |
+          0 0 1 * * root /bin/bash -l -c 'cd /var/app/current && RAILS_ENV=production bin/run_as_webapp bundle exec rake school_groups:send_review_group_tariffs_reminder >> log/jobs.log 2>&1'
```
ldodds, line 297: I don't think this needs to run every day and better to send the emails later in the day I think rather than at midnight.

You could further simplify the code in the services further by doing all the scheduling in the cronjob, e.g. "0 8 15 3,9 *" will schedule the cronjob to run at 8am on the 15th of March and September.

ldodds (changes requested):

iant:
@ldodds scheduling simplified to cron config only and mailer updates and tests added for cy locale support so I think this is ready for review again when you get a mo.

ldodds (approved):
Happy for this to be merged, but I think the text could be improved. Let's merge this and I'll update Trello with some suggestions which can be done in another PR.

PR #3057, opened 2023-09-14
Make checkbox align left on mobile (not centre as it was doing)

urbanwide:
Small change to checkbox css that was merged earlier...


iant (approved):
:+1:

PR #3058, opened 2023-09-14
Tidy up feature flag activities_2023

urbanwide:
* Removes use of activites_2023 features flag
* Also adds some missing system spec for bonus points in frontend
* Also annotated models


iant (approved):

PR #3066, opened 2023-09-20
Allow display of previous years scores for groups

ldodds:
Updates the current scores view for school groups to allow viewing of the final scores from the previous academic year.

The `Scorable` model concern already supported finding scores by academic year. But, to share more code between the scoreboards and school groups page the following changes were made:

* Added related methods for finding current/this academic year and previous year to `Scorable`
* Added a `Scoring` concern for the controllers to share finding the academic years and scored schools
* Updated the CSV generator to allow passing in of scored schools, so that the previous years scores can be downloaded
* Updated the templates to add links and normalise some variable names
* Added a few extra specs for the current scores tab, as well as a check for the new navigation options

### DIFF
--- a/spec/system/school_groups_spec.rb
+++ b/spec/system/school_groups_spec.rb
@@ -2,7 +2,8 @@ require 'rails_helper'
 
 describe 'school groups', :school_groups, type: :system do
   let(:public)                 { true }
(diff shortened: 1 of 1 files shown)


urbanwide (approved):
Looks good @ldodds 👍

PR #3069, opened 2023-09-21
Disable use of preferences on the pupil analysis charts only

ldodds:
Schools can set their default preferences for how we display charts.

However some of the pupil pages involve exploring the data, allowing users to request specific views. At the moment, for some of these charts (those on the "pupil analysis" pages) the school level preference is overriding what the user has requested to see.

Some of these charts also currently have hard-coded titles and descriptions that refer to costs. 

This PR fixes that bug by displaying use of the preference for just those charts. Other charts will use the school preference.


urbanwide (approved):
All good @ldodds 👍

PR #3075, opened 2023-09-21
Add new show on charts and fuel type fields to activity type and intervention type (2)

iant:


app/views/admin/activity_types/_form.html.erb
```diff
@@ -65,6 +65,10 @@
   <%= f.input :active %>
   <%= f.input :data_driven, label: 'Uses data?' %>
   <%= f.input :custom, label: "Custom ('Other')" %>
+  <%= f.input :show_on_charts, label: 'Show on charts?' %>
+
+  <h2>Fuel types</h2>
+  <%= render 'crud_filter_checkboxes', field: :fuel_type, collection: Meter::CREATABLE_METER_TYPES.map { |m| OpenStruct.new(id: m.to_s, name: m.to_s.capitalize.humanize) }, form: f %>
```
ldodds, line 71: This isn't correct. The values for fuel type should be `:gas, :electricity, :solar, :storage_heater`. See Trello card and the PR you did with the migration and the after party task.

Their ought to be some validation on the `ActivityType `and `InterventionType` models to ensure that only the valid `fuel_type`s can be stored

ldodds (changes requested):

ldodds (approved):

PR #3076, opened 2023-09-21
Hide table if we dont have an average start time

ldodds:
This PR updates the heating control page to hide some sections if we aren't able to determine an average heating start time for the last 7 days:

- remove text on the insights page which shows the time
- hide text and table on the analysis page 

The text for the average annual saving is left in place, providing some useful feedback during the summer/early autumn.


iant (approved):
:+1: looks good to me @ldodds

PR #3079, opened 2023-09-21
Change frequency of checking against DCC

ldodds:
Recheck meters against n3rgy API every month rather than every 2 months.


iant (approved):
:+1:

PR #3080, opened 2023-09-21
3686 fix locale in mailer previews

ldodds:
The rails mailer previews weren't properly displaying translated content for other emails, only the subject and some parts of the text were changed.

Updates all the mailer previews to pass in a locale parameter to the mailer, picking up the value of the url parameter used by the rails preview controller.


iant (approved):
:+1: looks good to me @ldodds

PR #3081, opened 2023-09-22
View component updates to bump to new version

urbanwide:
On hold with this due to this bug: https://github.com/ViewComponent/view_component/issues/1825


iant (approved):

PR #3082, opened 2023-09-22
Show both activities and actions as annotations on charts

iant:


app/services/charts/annotate.rb
```diff
@@ -1,52 +1,101 @@
...
 
-    private
+    def icon_for(observation)
+      case observation.observation_type
+      when 'activity' then observation.activity.activity_category.icon
+      when 'intervention' then observation.intervention_type.intervention_type_group.icon
+      end
+    end
+
+    def icon_color_for(observation)
```
ldodds, line 86: Not sure about the colours. I don't think introducing new colours is best option here and might be distracting.

Can I suggest using $dark and $grey from our standard colours currently, then we can get feedback?

db/migrate/20230922131629_add_icon_to_activity_category.rb
```diff
@@ -0,0 +1,5 @@
+class AddIconToActivityCategory < ActiveRecord::Migration[6.0]
+  def change
+    add_column :activity_categories, :icon, :string, default: 'volleyball-ball'
```
ldodds, line 3: can you use `clipboard-check` as that's what we're using on the school timeline of activities.

Also, if we're adding this as a field, then we'll need to add a control for it on the activity category admin forms, can you include that too?

ldodds (changes requested):

iant:
@ldodds update pushed for the new icon and all white as discussed.  i think this is ok to re-review now.

PR #3084, opened 2023-09-22
Bump puma & capybara

urbanwide:


ldodds:
We need to be careful about bumping Puma as the Elastic Beanstalk environment has a specific version built in.

While we install Puma I've never been 100% certain whether EBS uses it for all functionality.

ldodds:
OK, to help move this forward it'd be useful to have a second opinion on this.

AWS documentation explains that the AWS Linux distros come with Puma pre-installed see:

https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/create_deploy_Ruby.container.html

Version of Puma for current and previous platforms linked from here:

https://docs.aws.amazon.com/elasticbeanstalk/latest/platforms/platforms-supported.html#platforms-supported.ruby

The docs also say you can install your own version of Puma:

> You can configure your application with your own provided Puma server. This provides the option to use a version of Puma other than the one pre-installed with the Ruby platform branch...To do so, you must include and customize a Gemfile in your deployment. You're also required to configure a Procfile to start the application server"

Procfile configuration covered here:

https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/ruby-platform-procfile.html

I think we're doing everything required, so Puma upgrade should be OK.

However, if you ssh into server, then the puma command is available `puma --version`. I think we should confirm that after installing:

* the application launches properly
* the version of puma on the command line is the one we've installed?
* restarting the application via the EBS console works (this will involve a puma / procfile based restart I think)

ldodds (approved):

urbanwide:
Thanks @ldodds - when you say application restart via the console, would this be?:
`sudo bin/run_as_webapp bundle exec rails restart`

NB for later:
On test, `puma --version` yields
`puma version 5.6.4` and we are looking to bump to `6.4.0`
`which puma`
`/opt/elasticbeanstalk/.rbenv/shims/puma`

ldodds:
@urbanwide by console I meant using the ElasticBeanstalk web interface using the "Restart app services" action., as that runs their internal scripts/commands. I thought that might be a good way to confirm that they're not using their own version of the puma command anywhere.

urbanwide:
Ahh I see, thank you!

PR #3086, opened 2023-09-25
Add back in adult dashboard prompts

ldodds:
Until we have implemented the rest of the new engagement options, add back in the prompts to record adult and pupil lead activities on the dashboard.

Also, rather than have a dynamic section with up to 3 alerts on the dashboard, we will now show 2 by default and then a link to all of the "Recent alerts" in the advice pages.


urbanwide (approved):
Looks good to me @ldodds 👍

urbanwide (approved):
👍

PR #3089, opened 2023-09-26
Prevent prices of £1 and above being saved & widen field

urbanwide:
Also widens field, as previously, any longer numbers were being partially covered by the up/down arrows: 

Screenshot 2023-09-26 at 15 08 26

After PR changes:

Screenshot 2023-09-26 at 15 08 02

### DIFF
--- a/app/models/energy_tariff_price.rb
+++ b/app/models/energy_tariff_price.rb
@@ -18,14 +18,16 @@
 #
 class EnergyTariffPrice < ApplicationRecord
   MINIMUM_VALUE = 0.0
+  MAXIMUM_VALUE = 1.0
 
   belongs_to :energy_tariff, inverse_of: :energy_tariff_prices
 
   validates :start_time, :end_time, :units, presence: true
   validates :value, presence: true, on: :update
 
-  validates :value, numericality: { greater_than: MINIMUM_VALUE }, allow_nil: true, on: :create
-  validates :value, numericality: { greater_than: MINIMUM_VALUE }, on: :update
+  validates :value, numericality: { greater_than: MINIMUM_VALUE, less_than: MAXIMUM_VALUE }, allow_nil: true, on: :create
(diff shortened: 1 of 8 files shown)


app/views/energy_tariffs/energy_tariff_flat_prices/_form.html.erb
```diff
@@ -9,7 +9,7 @@
     </td>
     <td class="value align-middle">
       <%= f.input :value, pattern: '[0-9.]+', label: false,
-        input_html: { value: value_allowing_for_errors(energy_tariff_price) },
+        input_html: { value: value_allowing_for_errors(energy_tariff_price), style:"width: 8em" },
```
iant, line 12: perhaps this could be added as a class to `app/assets/stylesheets/energy_tariffs.scss`?

app/models/energy_tariff_price.rb
```diff
@@ -18,14 +18,16 @@
 #
 class EnergyTariffPrice < ApplicationRecord
   MINIMUM_VALUE = 0.0
+  MAXIMUM_VALUE = 1.0
```
iant, line 21: Not something for this PR - but I wonder if this maximum value threshold could be a site setting?
Esp since this now stops the tariff from being saved (rather than warns) and likely needs to be reviewed periodically.

iant (approved):
:+1: looks good to me @urbanwide couple of comments/suggestions

app/models/energy_tariff_price.rb
urbanwide: Good point, thanks! I'll pop it in a Trello card.

app/views/energy_tariffs/energy_tariff_flat_prices/_form.html.erb
urbanwide: Yes good point @iant, thanks! Making a change to the step too so that the up/down arrows use pence increments.

PR #3091, opened 2023-09-27
Update the special school category on the Edit school details page

iant:
See also https://github.com/Energy-Sparks/energy-sparks_analytics/pull/601
Will need to sync the analytics yml when doing the next tag release (the i18n spec will warn that this needs doing)


urbanwide (approved):
All good @iant 👍

PR #3092, opened 2023-09-27
Revise hot water analysis not enough data message

iant:


ldodds (changes requested):
Doesn't implement the `data_available_from` method which is what's needed here.

iant:
closing this as it needs some work on the analytics side

PR #3094, opened 2023-09-27
Limit chart annotations to those shown on charts and relevent fuel types

iant:


spec/services/charts/annotate_spec.rb
```diff
@@ -1,250 +1,791 @@
...
+      let!(:intervention_observation_multi_fuel) { create(:observation, :intervention, school: school, at: Date.new(2018, 6, 24), intervention_type: multi_fuel_intervention) }
+      let!(:activity_observation_multi_fuel) { create(:observation, :activity, school: school, at: Date.new(2018, 7, 8), activity: activity_multi_fuel) }
 
-      let(:x_axis_categories){ [
-        "24 Jun 2018",
-        "01 Jul 2018",
-        "08 Jul 2018",
-        "15 Jul 2018"
-      ]}
+      it 'is set true' do
```
urbanwide, line 54: This is what I was trying to explain earlier - about only putting test setup in `before` statements and putting only tests in `it` statements. I would suggest something more like this:

```
context "is true" do
  before do
    multi_fuel_intervention.update(show_on_charts: true)
    activity_type_multi_fuel.update(show_on_charts: true)
   end
   it "does whatever is expected"
end
```

spec/services/charts/annotate_spec.rb
```diff
@@ -1,250 +1,791 @@
...
+        multi_fuel_intervention.update(show_on_charts: true)
+        activity_type_multi_fuel.update(show_on_charts: true)
+        expect(subject_multi_fuel.annotate_weekly(x_axis_categories)).to eq(
+          [
```
urbanwide, line 58: Testing for large data structures like this in output can make tests difficult to read / debug / maintain. I am not sure how I would do it exactly in this case, but I'd try and replace this one test with multiple tests to check the output contains certain items (to prove it's behaving as expected). e.g. `it "has two annotations"`. `it "has an intervention observation"` etc. Maybe one to think about another time and probably would need shared examples and further contexts to achieve this. This would also have the side effect of describing what the desired output looked like without having to infer it from looking at the data structure.

urbanwide (commented):
Code looks good @iant but there are some modifications that could be made to the rspec, which we discussed today, so thought it worth bringing up while fresh.

spec/services/charts/annotate_spec.rb
iant: :+1:

spec/services/charts/annotate_spec.rb
iant: I agree that's probably good practice but I think in this case (and also where we're doing similar elsewhere) everything in the array is relevant to correct behaviour (e.g. `Observation` dates being pinned to the nearest week date, the icon type etc).

spec/services/charts/annotate_spec.rb
iant: have pushed an update for that @urbanwide

spec/services/charts/annotate_spec.rb
urbanwide: Thanks @iant. I'm always one for descriptive short tests where possible (not saying I always do it myself either), unless the test describes itself. Which it sounds like it does in this case, so, ok to leave as-is.

urbanwide (approved):

PR #3095, opened 2023-09-28
Remove simulator prototype

ldodds:
The analytics has a prototype of an energy usage simulator. This was developed a few years ago but was never finished or deployed to production. We have no plans to do that currently.

This PR removes the application side of the integration with that code: routes, models, controllers, templates, CSS and JS.

If we decide to move forward with the simulator we'll need to rebuild all the front end anyway.


iant (approved):
:+1: looks good to me @ldodds

PR #3096, opened 2023-09-28
Attempt a Rails 7 upgrade

iant:


iant:
@ldodds @urbanwide I'll close this as i don't think there's anything useful here (but closed pr will still be around for reference).  my first (naive) strategy was to upgrade the rails gem (and any dependencies) and get tests passing, commenting out lines that were breaking so as to fix and add back in later, but didn't get very far, sorry.   
Some links for reference:
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html
https://www.fastruby.io/blog/rails/upgrades/upgrade-rails-6-1-to-7-0.html
https://dev.to/thomasvanholder/how-to-upgrade-rails-61-to-rails-7-33a3

PR #3101, opened 2023-10-02
Change autovacuum delay for amr_validated_readings

ldodds:
Changes the `autovacuum_vacuum_cost_delay` parameter on the `amr_validated_readings` table to zero.

This should ensure that a vacuum runs as fast as possible, with minimal timeout. Hoping this reduces the "Timeout:VacuumDelay" waits we're seeing in the database and ensure that when a vacuum is required its done as quickly as possible. At the moment it looks like the system is timing out trying to complete timeouts, so the period when performance is slow / space has not been recovered is longer.

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/wait-event.timeoutvacuumdelay.html
https://www.cybertec-postgresql.com/en/tuning-autovacuum-postgresql/


urbanwide (approved):
👍

PR #3107, opened 2023-10-02
Allow archiving of schools

ldodds:
When we remove schools we do a soft delete, setting the `active` flag to false. The admin forms for doing this require user accounts to be locked and meters to be deactivated before the flag can be set.

This PR has some updates to the removal process to allow for "archiving" schools. If a school is archived, rather than deleted then we want to:

* retain the alert contacts for users, otherwise they can be deleted
* retain the links between the meters and the unvalidated readings, so when the school is re-enabled we can reload their data

When a school is deleted we should:

* remove alert contacts
* remove the links to the unvalidated readings

In both cases we want to:

* redirect admins to the removal page to clarify that the school has been removed
* disable use of the visibility, process data and other buttons
* indicate whether the school has been archived or removed on the existing school group and admin report
* filter those schools from the daily import emails and school group meter report

This PR has improvements to the removal page and implements the above behaviour. Further improvements can and will be made, but this provides the basic functionality to allow the admins to begin archiving schools.

### DIFF
--- a/app/controllers/admin/schools_controller.rb
+++ b/app/controllers/admin/schools_controller.rb
@@ -14,20 +14,52 @@ module Admin
       redirect_back fallback_location: root_path, notice: e.message
     end
 
-    def deactivate_meters
-      service = SchoolRemover.new(@school)
-      service.remove_meters!
-      redirect_back fallback_location: root_path, notice: "Meters have been deactivated and data removed"
+    def archive_meters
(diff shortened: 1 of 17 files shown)


iant (approved):
:+1: looks good to me @ldodds 
couple of comments on possibly using the `inactive` method in a few places

PR #3108, opened 2023-10-03
Fix missing route and add conditions around text

ldodds:
Previous PR for school removal had some bugs:

* incorrect route, which wasn't caught because of missing spec. Have fixed route and added extra spec
* the noticed were always appearing, so these are now wrapped in conditional blocks

plus a couple of text tweaks.


iant (approved):
:+1:

PR #3109, opened 2023-10-03
Change timing of import notifier email

ldodds:
Moves the import notifier email to 9.45am UTC.

Also removes a couple of commented out jobs


iant (approved):
👍

PR #3111, opened 2023-10-03
Update Energy Sparks Terms and Conditions section 4

iant:


config/locales/views/terms_and_conditions/terms_and_conditions.yml
```diff
@@ -26,6 +26,9 @@ en:
...
         <li> 4.3 Participating schools will commit to engage with Energy Sparks through a pupil eco-team or similar, using pupil-led action to drive energy reduction measures at the school. </li>
+        <li> 4.4 Funded schools receiving free access will nominate at least three Energy Sparks champions. These will include a member of the senior leadership team, a business/facilities/estates/site manager and a teacher. When staff leave the school they will commit to hand over their Energy Sparks user role to another suitable staff member. Staff users are strongly encouraged to attend at least one Energy Sparks training webinar per school year.</li>
+        <li> 4.5 Funded schools will commit to complete and record at least one Energy Sparks programme (https://energysparks.uk/programme_types) of pupil activities per school year (between 5 and 10 activities per programme); carry out and record at least one adult action (https://energysparks.uk/intervention_type_groups) per long school term based on Energy Sparks’ recommendations for the school; and nominate at least one named staff member to carry out a holiday switch off before each holiday.</li>
```
urbanwide, line 30: Should these links that are to the energysparks app be passed as parameters?

urbanwide (changes requested):
See inline comment about the links being passed as parameters. Also I've just noticed that Terms and Conditions on the website has a Revision date of 25th April 2023, will this need updating too?

config/locales/views/terms_and_conditions/terms_and_conditions.yml
iant: ^^ @ldodds ???

iant: urls look to be hardcoded in brackets but let me check. 

i'v pushed an update for the revision date.  good spot.

iant: @urbanwide I've pushed an update to pass urls as parameters

urbanwide (approved):
Looks good @iant 👍

PR #3112, opened 2023-10-03
Tweaks to loading and deleting readings

ldodds:
Reviewing the code and logs for when we are processing validated readings I noticed that:

- we order readings when loading from the database, but these are later added to an `AmrData` object which is a hash. So the ordering is unnecessary
- when we delete readings we were deleting all the dates present in the `AmrData.keys`. This isn't necessary as the `AmrData` returned by the analytics is guaranteed to have a continuous time series between its start and end dates. So we only have to delete data before the `start_date` or after the `end_date`. 

This PR fixes both of those issues. These are unlikely to have a huge impact on performance, but might make some of the data processing slightly quicker.

### DIFF
--- a/app/services/amr/analytics_validated_amr_data_factory.rb
+++ b/app/services/amr/analytics_validated_amr_data_factory.rb
@@ -5,7 +5,7 @@ module Amr
   private
 
     def build_meter_data(active_record_meter)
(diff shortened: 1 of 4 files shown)


urbanwide (approved):
Looks good @ldodds 👍

PR #3113, opened 2023-10-03
Show benchmark flag on description

ldodds:
Trivial change to add an `AlertType` flag to the show page.

Helps identify whether an alert type is used to also generate benchmark metrics. If not, and we've not set up any ratings then its safe to disable.

Adding this flag will make it easier to review the settings.


iant (approved):
👍

PR #3116, opened 2023-10-03
Optimise generation of out of hours usage benchmark

ldodds:
This PR updates the out of hours usage benchmark generator to use a new method on `AnnualUsageBreakdownService` in the analytics. It avoids generating additional variables (co2, £, £current) when we only need the kwh values for the benchmark assessment.

This PR relies on this change: https://github.com/Energy-Sparks/energy-sparks_analytics/pull/606. 

Cannot be merged until we have a new tagged version of the analytics.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,8 +22,8 @@ gem 'closed_struct'
 gem 'pg_search'
 
 # Dashboard analytics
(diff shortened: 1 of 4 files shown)


iant (approved):
:+1:

PR #3118, opened 2023-10-04
Update reports to only include active schools

iant:


app/models/school.rb
```diff
@@ -186,6 +186,7 @@ class ProcessDataError < StandardError; end
   scope :inactive,            -> { where(active: false) }
   scope :deleted,             -> { inactive.where.not(removal_date: nil)}
   scope :archived,            -> { inactive.where(removal_date: nil)}
+  scope :active_and_archived, -> { where(removal_date: nil)}
```
urbanwide, line 189: Just checking - is ok not not check the active flag here - and rely on the removal date being nil as the primary indicator of being active or archived?

urbanwide (approved):
Looks good @iant - couple of comments / questions. Thanks

app/models/school.rb
ldodds: agree, might be clearer to include active + archived more explicitly

app/models/data_source.rb
```diff
@@ -47,6 +47,10 @@ def to_csv
     end
   end
 
+  def meters_from_active_schools
```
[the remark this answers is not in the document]
iant: :+1: yep, sounds good to me.  have pushed an update.

app/models/procurement_route.rb
```diff
@@ -24,7 +24,7 @@ class ProcurementRoute < ApplicationRecord
   def to_csv
     CSV.generate(headers: true) do |csv|
       csv << csv_headers
-      meters.each do |meter|
+      meters_from_active_schools.each do |meter|
```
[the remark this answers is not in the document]
iant: :+1:

app/models/school.rb
iant: :+1:

app/models/school.rb
```diff
@@ -184,8 +184,9 @@ class ProcessDataError < StandardError; end
   #are archived, with chance of returning if we receive funding
   scope :active,              -> { where(active: true) }
   scope :inactive,            -> { where(active: false) }
-  scope :deleted,             -> { inactive.where.not(removal_date: nil)}
-  scope :archived,            -> { inactive.where(removal_date: nil)}
+  scope :deleted,             -> { inactive.where.not(removal_date: nil) }
+  scope :archived,            -> { inactive.where(removal_date: nil) }
+  scope :active_and_archived, -> { find_by_sql(active.to_sql + ' UNION ALL ' + archived.to_sql) }
```
urbanwide, line 189: Is this the same as writing active.or(archived) ? Just to simplify a bit if so...

PR #3119, opened 2023-10-04
Properly lock accounts when school is deleted

iant:
The `SchoolRemover` tries to lock user accounts using the Devise locking feature. However the current Devise config allows locks to automatically expire after an hour.  This PR

- [x] revises the config so any locked accounts must request an unlock email, rather than being timed out
- [x] ensures the admin views of user lists indicate whether an account is locked, with an option to manually unlock

### DIFF
--- a/app/views/admin/school_groups/users/show.html.erb
+++ b/app/views/admin/school_groups/users/show.html.erb
@@ -23,6 +23,7 @@
         <th>Confirmed?</th>
         <th>Last signed in</th>
         <th>Language</th>
+        <th>Locked?</th>
         <th></th>
       </tr>
     </thead>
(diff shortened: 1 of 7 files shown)


config/initializers/devise.rb
```diff
@@ -180,15 +180,12 @@
   # :time  = Re-enables login after a certain amount of time (see :unlock_in below)
   # :both  = Enables both strategies
   # :none  = No unlock strategy. You should handle unlocking by yourself.
-  config.unlock_strategy = :both
+  config.unlock_strategy = :email
```
iant, line 183: not sure if this should be `:none` or `:email`

ldodds: I think email is fine as we do want to allow users to unlock themselves if its triggered for schools that are active.

app/views/admin/school_groups/users/show.html.erb
```diff
@@ -23,6 +23,7 @@
         <th>Confirmed?</th>
         <th>Last signed in</th>
         <th>Language</th>
+        <th>Locked?</th>
```
ldodds, line 26: The locked status and buttons should only be visible to admins, and not group admins.

ldodds (changes requested):

ldodds (approved):

PR #3120, opened 2023-10-04
Reduce number of updates on amr_validated_readings

ldodds:


app/services/amr/upsert_validated_readings_for_a_meter.rb
```diff
@@ -22,6 +25,56 @@ def perform
...
+      #Calling private method here, but means we can piggy-back off all the SQL creation
+      insert = insert_all.send(:to_sql)
+      where = <<-SQL.squish
+        WHERE
+          amr_validated_readings.kwh_data_x48 IS DISTINCT FROM excluded.kwh_data_x48 OR
+          amr_validated_readings.one_day_kwh IS DISTINCT FROM excluded.one_day_kwh OR
+          amr_validated_readings.status IS DISTINCT FROM excluded.status OR
+          amr_validated_readings.substitute_date IS DISTINCT FROM excluded.substitute_date OR
+          amr_validated_readings.upload_datetime IS DISTINCT FROM excluded.upload_datetime
+        RETURNING "id"
```
iant, line 73: to avoid any ambiguity (not sure if excluded has an id?) maybe this should read `RETURNING "amr_validated_readings "."id"`

ldodds: This is the same as what the `InsertAll` action produces by default, so just kept to that. There's no ids in the inserted data, so this is the id for any new or updated records.

iant (approved):
I think this looks good to me @ldodds

urbanwide (approved):
Looks good @ldodds 👍

PR #3121, opened 2023-10-04
Do not remove user from school if school is archived

iant:
When we remove a school, if a user is linked to more than one school, then we unlink them from this school.
But we don’t want to do that if we’re archiving a school, as we want the same users to be linked in future if its reenabled.  This PR adds changes so we:

- [x] don’t remove users from schools when they are archived or deleted, so the links are retained
- [x] still only lock accounts if user is linked to only this school
- [x] relax the condition on the School Remover to allow a school to be removed if either all accounts are locked, or the users are linked to other schools
- [x] update the Switch School menu to only show active schools

note: `cluster_schools_for_switching` (used in the last item switch school menu) is already set to exclude non active schools, so there are no changes for that in this pr.  see https://github.com/Energy-Sparks/energy-sparks/blob/master/app/models/user.rb#L107

### DIFF
--- a/app/services/school_remover.rb
+++ b/app/services/school_remover.rb
@@ -15,9 +15,9 @@ class SchoolRemover
   end
 
   def users_ready?
-    # Requires all users are access locked except those users linked to another school
+    # Requires all school users are access locked except those linked to another school
     return true if @school.users.all?(&:access_locked?)
-    return true if unlocked_users_linked_to_another_school.any?
+    return true if all_unlocked_users_are_linked_to_other_schools?
 
     false
   end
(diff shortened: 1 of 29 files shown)


app/services/school_remover.rb
```diff
@@ -61,6 +68,35 @@ def remove_meters!
 
   private
 
+  def all_unlocked_users_are_linked_to_other_schools?
```
ldodds, line 71: I think the logic is correct here: any users linked ONLY to this school should be locked. Any users links to other schools do not have to be locked in order to proceed with archiving/removal. And tests cover that.

We shouldn't have any SQL in a service class though, so any db code needs to be on a model.

The `User` already has a `has_other_schools?` method, so you could probably simplify this to finding the unlocked users and rejecting those that don't have other schools.

ldodds (changes requested):

ldodds (approved):

PR #3123, opened 2023-10-06
Use chart component for all school charts

iant:


app/views/management/schools/report.html.erb
```diff
@@ -14,7 +14,12 @@
 
 <% @overview_charts.each_with_index do |(energy, chart_config), index| %>
   <h2><%= t("management.schools.overview_charts.overview.#{energy.to_s}") %></h2>
-  <%= chart_tag(@school, chart_config[:chart], no_zoom: true, chart_config: {y_axis_units: select_y_axis(@school, chart_config[:chart], chart_config[:units])}) %>
+  <%= component 'chart',
```
ldodds, line 17: Looking at the existing page, this printable version of the dashboard doesn't have the axis controls or analysis controls, so I think these options should be passed into the component?

app/components/chart_component.rb
```diff
@@ -6,11 +6,11 @@ class ChartComponent < ViewComponent::Base
   renders_one :header
   renders_one :footer
 
-  attr_reader :school, :chart_type, :analysis_controls, :no_zoom, :axis_controls, :html_class, :fuel_type
+  attr_reader :school, :chart_type, :analysis_controls, :no_zoom, :axis_controls, :html_class, :fuel_type, :show_advice, :autoload_chart, :wrap
 
   include ChartHelper
 
-  def initialize(chart_type:, school:, chart_config: nil, analysis_controls: true, no_zoom: true, axis_controls: true, html_class: 'analysis-chart', fuel_type: nil)
+  def initialize(chart_type:, school:, chart_config: nil, analysis_controls: true, no_zoom: true, axis_controls: true, html_class: 'analysis-chart', fuel_type: nil, show_advice: false, autoload_chart: true, wrap: true)
```
ldodds, line 13: Adding the autoload option is a good extension, but I'm not sure about the show_advice and wrap settings.

The "show_advice" option on the chart tag is a holdover from much earlier versions of the analytics. I'm not sure its even used anywhere at the moment, it was originally used to inject some content from the JSON response into the page. I'm not sure its actually used anywhere any more. I'd prefer to try and identify that and then remove the option rather than preserve it.

For the "wrap" option, my goal with this change is to remove the need for the chart wrapper, rather than add it to the component. What I was hoping was that we'd be able to replace use of the wrapper divs (which usually contain content before and after the chart, see e.g. custom titles on the pupil analysis controls, or the analysis controls partial ), with the existing chart component slots and options. I think you can achieve the same thing using those.

Can you revisit and see whether you can achieve the same thing with existing features?

ldodds (changes requested):

app/components/chart_component.rb
iant: looks like show advice is set true in `app/views/admin/analysis/generic_chart_template.html.erb` and potentially in `app/views/schools/charts/show.html.erb` (`!params[:control].present?`)  ok to remove for those two?

will look at removing the wrap option and chart wrapper

iant: the wrap & show_advice removed from the component options

app/views/schools/charts/show.html.erb
ldodds, line 1: If we've removed `render_html` in the controller, then can't we delete this view?

app/views/management/schools/report.html.erb
ldodds, line 17: Does this need `axis-controls: false` as I don't see those controls on the current page (bottom of the dashboard, 'print view')

app/components/chart_component/chart_component.html.erb
```diff
@@ -13,7 +13,16 @@
     <%= header %>
 
     <%= render 'shared/analysis_controls', chart_type: chart_type, axis_controls: axis_controls, analysis_controls: analysis_controls %>
-    <%= chart_tag(school, chart_type, no_zoom: no_zoom, wrap: false, html_class: html_class, chart_config: chart_config, fuel_type: fuel_type) %>
+    <% if dynamic_titles %>
```
ldodds, line 16: I don't think this option is needed as we've already got the `:header` slot on the component, see line 13.

So any templates that use a dynamic_titles section can set that to be the header slot and they will be added before the chart?

app/views/schools/usage/show.html.erb
```diff
@@ -25,10 +25,14 @@
   </div>
 
   <div class="row">
-    <div class="col-md-12 chart-wrapper" id="chart_wrapper_<%=@chart_config[@period]%>">
```
ldodds, line 28: We've lost the `col-md-12` styling here as well as the chart-wrapper is that not needed?

iant: :+1: will add back in

app/views/schools/charts/show.html.erb
iant: :+1: yep will remove

app/components/chart_component/chart_component.html.erb
iant: this pr's started to get a bit unwieldy so I think leave in as part of this pr and remove as a separate bit of work?

app/views/management/schools/report.html.erb
iant: :+1:

ldodds (approved):

PR #3125, opened 2023-10-07
Onboardings automated reminder mailer

urbanwide:
This can't be switched on until plural translations have been added from Transifex. So have popped it behind a send_automated_reminders feature flag.

Automates reminder emails to be sent either one week after the initial onboarding email was sent and then weekly after the last reminder until the onboarding is completed.

The bit of code to find onboardings with reminders due could probably be done with more sql and less rails, but because there aren't many incomplete onboarding records and it can be run anytime of night (currently 4am in cron), I am hoping it will be ok as is. But can look again if need be.


app/models/school_onboarding.rb
```diff
@@ -71,6 +71,14 @@ def has_event?(event_name)
     events.where(event: event_name).any?
   end
 
+  def last_event(event_name)
+    events.where(event: event_name).last
```
iant, line 75: probably overkill but does this need some ordering?  e.g. `events.where(event: event_name).order(created_at: :asc).last` (or `events.where(event: event_name).order(updated_at: :asc).last`)

app/controllers/admin/school_groups/school_onboardings_controller.rb
```diff
@@ -78,15 +77,20 @@ def produce_csv_row_manual(school)
         ]
       end
 
-      def for_selected(notice)
-        if (ids = params.dig(:school_group, :school_onboarding_ids))
-          @school_group.school_onboardings.find(ids).each do |onboarding|
+      def for_selected(message)
+        if onboardings.any?
+          onboardings.each do |onboarding|
             yield onboarding
```
iant, line 83: I know this isn't in your changes but I think, if possible, it would be useful to have a comment explaining what this yield is doing here.

app/models/school_onboarding.rb
urbanwide: Good shout - I did think the same at the same, then I looked at other bits of code in the same class that did the same and saw that it wasn't done there. I think it's a good idea to add it to be sure as you say though!

app/controllers/admin/school_groups/school_onboardings_controller.rb
urbanwide: Thanks - it is my code from before though 😀 Happy to document!

app/services/onboarding/reminder_mailer.rb
```diff
@@ -0,0 +1,44 @@
+module Onboarding
+  class ReminderMailer
+    THRESHOLD = 7
```
iant, line 3: maybe rename this `DAY_THRESHOLD`? (sorry, missed this last time)

iant (approved):
:+1: look good @urbanwide

PR #3126, opened 2023-10-09
Rework school regeneration process

ldodds:
This PR reworks how we regenerate the school configuration, alerts, benchmarks and update the rails cache with latest version of meter collection.

Every day, for each school we need to:

- load and validate the school meter data
- update the database with the latest version of validated meter data
- ensure that the meter collection for the school is in the rails cache, to improve performance of charts and advice pages
- update a range of school metrics and configuration with cached data to improve performance of the rest of the application

This process is run for all schools via the "daily regeneration job" and can be run manually when we "revalidate" a school.

The PR refactors the existing code to create:

- `Schools::SchoolRegenerationService` - which is now responsible for coordinating the overall process for each school
- `Schools::SchoolMetricsGeneratorService` - which coordinates the final step, invoking a range of separate services

These new classes replace the older `ContentBatch` service and `ContentRegenerator` services which have been removed, and simplifies the `DailyRegenerationJob`. 

The specs for the new services focus on checking that all the required steps are called, as there are specific tests for the individual services that generate specific metrics and bits of school configuration.

Aside from tidying the code the main motivation for the refactoring was to remove a redundant step in the overall process. 

Currently we create a `MeterCollection` using the unvalidated readings, then use that to update the database. We then throw that away and create a new `MeterCollection` object using the validated readings in the database. This is unnecessary as the meter data will be exactly the same. We just ending up using more short-term memory and run an extra set of SQL queries to load the validated readings.

The `SchoolRegenerationService` now just uses the existing meter collection to drive creation of the updates metrics and stores that directly in the Rails cache.  

Based on manual testing, for schools with more data/meters, this should improve performance of the overall process by around 10%.


lib/tasks/schools/daily_regeneration.rake
```diff
@@ -3,7 +3,7 @@ namespace :school do
   task daily_regeneration: :environment do
     puts "#{DateTime.now.utc} Run daily regeneration for all process data schools start"
     BenchmarkResultGenerationRun.create!
-    School.process_data.each do |school|
+    School.process_data.order(:name).limit(5).each do |school|
```
urbanwide, line 6: Do you mean to limit to 5 here?

ldodds: I didn't! that was left over from manual testing. I've pushed an update.

The ordering was intentional though to make it a little more predictable when different schools get executed in the queue.

spec/services/schools/school_metrics_generator_service_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  #this will create an empty meter collection as the school has no data
+  let(:meter_collection)  { Amr::AnalyticsMeterCollectionFactory.new(school).validated }
+
+  subject(:service)       { Schools::SchoolMetricsGeneratorService.new(school: school, meter_collection: meter_collection)}
+
+  let(:stub)              { double('service_stub') }
+
+  describe '#perform' do
+
+    context 'it should update school configuration' do
```
urbanwide, line 17: I appreciate this spec is taken from elsewhere and understand you might not have had time to change it. 

Context labels - I'd be labelling them more like "updating school configuration" or "when updating school configuration" and I'd be looking to put the expect statements also inside the 'it' block (and not the 'before' block with them being tests and not test setup).

urbanwide (approved):
Looks good @ldodds. Just a small comment about some of the rpsec format - think the rspec is taken from the deleted files though, so I understand why it's as it is (so all good).

spec/services/schools/school_metrics_generator_service_spec.rb
ldodds: Yes, good shout I rushed this a bit. I've reworked the specs as suggested.

PR #3127, opened 2023-10-09
Increase cache expiry times

ldodds:
The daily regeneration process for schools is taking longer to complete. Some days its hitting 3+ hours.

This means that jobs running at the end of the process may end up having to reload some of the "schedule data" (temperatures, solar pv data, calendars, etc).

This PR increases the cache period by an hour to avoid that happening.


urbanwide (approved):
Looks good @ldodds 👍🏻

PR #3128, opened 2023-10-09
Pluck fields from amr_data_feed_reading

ldodds:
Change how we load the `AmrDataFeedReadings` for the validation process. We only need 4 fields from the model, so just `.pluck` those values rather than creating a full active record model. This should be more efficient

This change was already done for the validated readings, but not the unvalidated.


urbanwide (approved):
All good @ldodds 👍

PR #3129, opened 2023-10-09
Remove unused index on amr_validated_readings

ldodds:
Checking [the statistics in `pg_stat_all_indexes` table](https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STAT-ALL-INDEXES-VIEW) shows that there is an index on the `amr_validated_readings` table on the `meter_id` column which is unused. The stats have zero reads using that index.

Looks like postgres prefers to use the compound, unique index when doing all loads.

This PR removes the index, which will reclaim space and avoid need to maintain the index during inserts, vacuums, etc.


urbanwide (approved):
👍

PR #3130, opened 2023-10-09
Create OneDayAMRReading when loading readings

ldodds:
Currently when we load the unvalidated and validated readings from the database we build a hash for each reading, for each meter.

This is passed to a second factory in the analytics (`MeterCollectionFactory`) which then turns the hash of meter data into `Dashboard::Meter` and `AMRData` objects populated with `OneDayAMRReading` values.

This PR alters this so that the application just creates the `OneDayAMRReading` objects as it reads data from the dashboard. This saves on creating potentially thousands of Hash objects which are just immediately thrown away. So should be more memory efficient.

There's no other benefits to the indirection as elsewhere we just create the analytics objects directly. Having clearer dependencies also helps understand how and where data is created. Suspect this code is a hold-over from earlier version of the application.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,8 +22,8 @@ gem 'closed_struct'
 gem 'pg_search'
 
 # Dashboard analytics
-gem 'energy-sparks_analytics', git: 'https://github.com/Energy-Sparks/energy-sparks_analytics.git', tag: '4.0.3'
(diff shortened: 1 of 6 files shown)


urbanwide (approved):
Looks good @ldodds 👍

PR #3132, opened 2023-10-10
Prevent onboarding failing if the postcode is not found

iant:


config/initializers/geocoder.rb
```diff
@@ -28,16 +28,60 @@
 if Rails.env.test?
```
ldodds, line 28: As these are tests, I think they should be in a file in `spec/support` ?

ldodds (changes requested):

config/initializers/geocoder.rb
iant: :+1: have pushed an update moving them to spec/support

ldodds (approved):

PR #3133, opened 2023-10-10
Remove use of bootstrap-email for admin mailers

ldodds:
This PR removes the use of `bootstrap-email` for the admin mailers.

Internally `bootstrap-email` delegates to `premailer` which seems to be the standard Rails gem for inlining CSS into HTML emails. Premailer has some optimisations to work around performance issues for large emails. This poor performance is particularly bad for emails with tables and manifests itself as very high CPU usage

Bootstrap uses a lot of CSS selectors that cannot be optimised by Premailer. This results in very slow performance. That performance is actually due to nokogiri/libxml2 being slow at searching larger DOMs. 

So there doesn't seem to be another workaround for our high CPU issues other than removing bootstrap-email.

This PR:

* Adds a new custom layout for the `AdminMailer` and `ImportMailer`
* Removes references to the `application_mailer.css` file from that layout and inlines a number of CSS styles from bootstrap. This includes some basic styling for tables and some utility classes.
* The mailers have then been updated to just call `mail` rather than `make_bootstrap_mail`

I've chosen to inline the CSS classes for now as it avoids adding SCSS compilation to the email generation. It at least keeps the styles where they are used.

The admin mails that are affected don't look too bad in the new styles. The most obvious change is removal of the header and footer which wasn't necessary for an internal email. We can always improve styling in later PRs.

As a side effect of this change, the specs that test these mailers will likely run faster!

### DIFF
--- a/app/mailers/admin_mailer.rb
+++ b/app/mailers/admin_mailer.rb
@@ -1,5 +1,6 @@
 class AdminMailer < ApplicationMailer
   helper :application, :issues
+  layout 'admin_mailer'
 
   def school_data_source_report
     to, data_source_id = params.values_at(:to, :data_source_id)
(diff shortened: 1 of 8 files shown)


urbanwide (approved):
Looks good @ldodds 👍

PR #3134, opened 2023-10-10
Attempt to regenerate following validation failure

ldodds:
Testing the updated school regeneration workflow highlighted that if the validation stage failed, then we end up with further downstream failures.

The previous workflow would have raised an error but would have continued to regenerate the school using the existing validated readings (if any).

This PR adds a fallback step if an exception is thrown by the validation process. In that case we build a new meter collection, using the existing validated data in the database, and then continue processing. 

This ensures that the school is properly regenerated any will see warnings about out of date data, etc. The


iant (approved):
:+1:

PR #3135, opened 2023-10-11
Add premailer-rails

ldodds:
A follow-up to https://github.com/Energy-Sparks/energy-sparks/pull/3133.

While the previous PR appeared to work fine, with emails showing up as styled in mailcatcher and rails previews, this is because the browser was rendering the styles. They weren't actually being inlined into the emails.

To make this work correctly we also need to add `premailer-rails`. This automatically processes ALL outgoing emails using premailer. This helps fix styling on the admin emails but interferes with the bootstrap-email gem, so as part of customising we force the premailer processing to be disabled there.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -114,7 +114,7 @@ gem 'delegated_type'
 gem 'roo'
 gem 'roo-xls'
 
-# Used to handle CSS style processing for
+# Used to handle mail processing for the admin mailer
 gem 'premailer-rails'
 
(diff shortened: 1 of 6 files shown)


iant (approved):
:+1:

iant (approved):
:+1: updates look good to me @ldodds

ldodds:
Third attempt @iant !

PR #3138, opened 2023-10-11
Optimise meters_running_behind query

ldodds:
The ImportNotifier is fairly slow. It does several complex queries, sorts and then loads a lot of school and meter data to format the report. This PR is part of ongoing steps towards further optimising the code.

Pushes the sort carried out on `find_meters_running_behind` into the database, avoiding need to do it in Ruby. This is the section of the report that consistently has most data, so worth doing.

PR also refactors the tests to make it more extensible/maintainable.

Further PRs will need to:

- optimise the other two queries
- use `includes` to include the key school group, school fields to reduce number of n+1 queries when formatting the report


app/services/import_notifier.rb
```diff
@@ -56,7 +49,8 @@ def find_meters_running_behind
          .joins('LEFT JOIN data_sources on data_sources.id = meters.data_source_id')
          .joins(:amr_validated_readings)
          .where(schools: { active: true })
-         .group('meters.id, data_sources.import_warning_days')
+         .joins('LEFT JOIN school_groups on schools.id = school_groups.id')
```
iant, line 52: should this be `.joins('LEFT JOIN school_groups ON schools.school_group_id = school_groups.id')`?

iant (approved):
:+1:

PR #3142, opened 2023-10-11
Reject rows when mpan_mprn contains a +

iant:


ldodds (changes requested):
The code successfully generates a warning, but there's no feedback to the user. You're adding a `:invalid_non_numeric_mpan_mprn` to the list of warnings, but you've not added it to the warnings hash.

The view does a lookup on that, see `_warnings.html.erb` which does `AmrReadingData::WARNINGS[w.to_sym]`.

Can you add a warning message about the invalid MPAN?

iant:
@ldodds just pushed an update to add the warning (and a couple of additional tests) so i think can be reviewed again

ldodds (approved):

PR #3143, opened 2023-10-11
New data config for MyCorona portal

iant:


iant:
note:  I think this can be accomplished in `SingleReadConverter` so need to write some tests for row per reading files that have date & times over two columns

iant:
closing this in favour of https://github.com/Energy-Sparks/energy-sparks/pull/3159

PR #3144, opened 2023-10-12
Make admin user search case insensitive

iant:


urbanwide (approved):
Looks good @iant - could a quick test be put in somewhere?

iant:
thanks for the review @urbanwide yep, no probs at all. I've pushed an update to add a system test for the search behaviour.
(I was hoping I could to get away with a one character change pr :D )

ldodds (approved):

PR #3147, opened 2023-10-12
Reduce meter collection sizes by removing unused schedule data

ldodds:
Some background on this PR here: https://github.com/Energy-Sparks/energy-sparks_analytics/pull/609

Changes are:

* Remove old `:use_meteostat feature` flag from schedule data manager. As this behaviour will always be used for now. A follow-up PR will tighten up some of the validations
* Remove the old `:date_bound_schedule_data` which was only ever activated on test, but not live
* Updated the analytics to add behaviour to prune unused data from meter collection post aggregation. The code being removed in the `:date_bound_schedule_data` feature implemented this slightly differently: it only created validated meter collections with limited data.

The net result of this change should be that the meter collections added to the Rails cache are much smaller, as they have fewer unused data points. This revised approach works better with the new school regeneration flow and should have the same result

This won't impact the speed of the regeneration job, but may reduce overall memory footprint / garbage collection as we load meter collections for advice pages and charts.


iant (approved):
:+1:

PR #3150, opened 2023-10-12
Rename include_examples -> it_behaves_like

urbanwide:
Branched from https://github.com/Energy-Sparks/energy-sparks/pull/3148


ldodds:
As with the other PR, if you merge #3148 I'll review this

ldodds (approved):
Looks good. I did play around with the various options for including shared examples for the tariff specs, and went with "include_examples" because it produced a more concise output (less nesting?). But happy to go with "it_behaves_like" if we're going to standardise on that everywhere.

urbanwide:
Thanks @ldodds. I totally can see that! 

One thing I've noticed by doing this refactoring is that include_examples behaves a bit like include_context - so any variables etc set up when using it are thrown in to the current context.  This can lead to unintended consequences as something done inside a shared example can affect a test that comes after it. Or if it is intended, can make tests harder to follow if the scope isn't clear.

I have left `system/schools/school_targets_spec.rb` using include_examples as it's doing a bit of a combo of including context and including examples for the tests that come after it.

I think it's ok to flex a bit on this sometimes though and just make this more of a guideline.

PR #3153, opened 2023-10-16
Improve display of data feed configs

ldodds:
This PR improves the admin page that lists the data feed configurations.

* Only show non API configurations, the others are just used internally
* Split page into two tabs, one showing the list of configurations with their notes plus action buttons, the other with more detail on the configuration to make them easier to review and compare
* Include tooltips to explain some of the configuration
* Improve the system specs a bit to confirm page displays as expected
* Improve the form by reorganising fields and adding hints


urbanwide (approved):
Looks good @ldodds 👍

PR #3156, opened 2023-10-18
Add prompt to complete programme

urbanwide:
Add prompts to complete programmes to recommendations page - now looks like this:
image

* Changed points logic to only display points remaining
* Added some styling
* Present singular text for when there is only one activity left


urbanwide:
My bonus points logic is wrong!

ldodds (approved):

PR #3157, opened 2023-10-18
Improve data load reports

ldodds:
Allow admins to page through the list of manual data load reports. 

- Add pagy support to the controller and template
- Change page size to 20 rather than 50

Also revises the recent amr import reports page to:

- Apply a 30 day window to the summary counts in the tabs
- Use 30 rather than 7 days as the window for totalling counts by identifier, but leaves summary of changes in last 48 hours
- Highlight disabled configs so we can see if we are still loading data using them


urbanwide (approved):
All good @ldodds 👍

PR #3158, opened 2023-10-18
Remove mpans reformatted by Excel

ldodds:
When the admins export from Excel to CSV, its possible for the MPANs to be accidentally reformatted based on how they are shown on screen rather than as the full string. This has lead to data in the database with data like "7.22E+13".

Delete all these records from the AmrDataFeedReading table.


urbanwide (approved):
Looks good @ldodds 👍

PR #3160, opened 2023-10-19
Halt regeneration if aggregation fails

ldodds:
Some further improvements to the service responsible for regenerating schools each day, and on demand by admins.

Reviewing Rollbar we are getting some errors from the analysis which are due to the aggregation process failing. This leaves the meter collection in an inconsistent state and we end up trying to calculate, e.g. equivalences, without the necessary data. We log each of these errors which just adds noise: the actual problem is the failure to aggregate properly.

This PR updates the service so that if the aggregation step throws an exception then we don't continue with the final steps. The school is just removed from the rails cache.

So the new process is:

- validate the school's meter data and update the database if sucessful
- if the validation step fails and we cant update their data, then continue using any previously saved data, so the school can at least see any existing analysis
- if the aggregation step fails then we can't continue any further, so just end the process
- if both validation and aggregation steps complete, then update equivalences, saved metrics, etc

### DIFF
--- a/app/services/schools/school_regeneration_service.rb
+++ b/app/services/schools/school_regeneration_service.rb
@@ -7,9 +7,16 @@ module Schools
 
     def perform
       meter_collection = validate_and_persist_readings
-      run_aggregation(meter_collection)
-      cache_meter_collection(meter_collection)
-      regenerate_school_metrics(meter_collection)
+      continue = run_aggregation(meter_collection)
+      if continue
+        cache_meter_collection(meter_collection)
+        regenerate_school_metrics(meter_collection)
+        return true
+      else
+        @logger.error("Unable to continue with regeneration")
+        remove_meter_collection_from_cache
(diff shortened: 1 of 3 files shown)


app/services/schools/school_regeneration_service.rb
```diff
@@ -7,9 +7,16 @@ def initialize(school:, logger: Rails.logger)
...
-      regenerate_school_metrics(meter_collection)
+      continue = run_aggregation(meter_collection)
+      if continue
+        cache_meter_collection(meter_collection)
+        regenerate_school_metrics(meter_collection)
+        return true
+      else
+        @logger.error("Unable to continue with regeneration")
+        remove_meter_collection_from_cache
+        return false
```
iant, line 18: i think these `return`s can be removed

iant (approved):
looks good to me @ldodds

PR #3162, opened 2023-10-19
Correct truncation of schedule data and only produce estimates when required

ldodds:
For schools with less than a year of data, we offer the option of them providing an estimate of their annual usage so that they can still set a target even if we have limited historical data.

If we have some data for the school (e.g. a few months) we try and help them produce a decent estimate by providing a calculation of our own. The school `Configuration` class stores whether an estimate is necessary for a fuel type and our best guess.

Debugging a problem regenerating some schools I found to issues with this code.

- there was a bug in the truncation of the schedule data, which meant that calculating some estimates was failing. This is fixed in the analytics (https://github.com/Energy-Sparks/energy-sparks_analytics/pull/613)
- we are currently producing estimates for all fuel types, regardless of whether they're actually needed. e.g. we are still estimating electricity usage even if the school has >1 year of data

This PR bumps the analytics tag to include the bug fix and updates the `GenerateEstimatedUsage` service to check whether an estimate is actually needed before attempting to generate one.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,7 +22,7 @@ gem 'closed_struct'
 gem 'pg_search'
 
 # Dashboard analytics
-gem 'energy-sparks_analytics', git: 'https://github.com/Energy-Sparks/energy-sparks_analytics.git', tag: '4.0.5'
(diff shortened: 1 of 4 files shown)


urbanwide (approved):
Looks good @ldodds 👍

PR #3163, opened 2023-10-19
Tune vacuum analyse for reading tables

ldodds:
Some time ago we introduced the `VacuumService` to manually VACUUM and ANALYSE on the `amr_data_feed_readings` table as running `count` on the table was becoming very slow after updates. The analyse process recalculates the table statistics and avoids need to do a full scan to do the count.

Since then we've learned some more about how to tune the autovacuum process that should be running the VACUUM and ANALYSE process automatically. 

It turns out that due to the fact that the `amr_data_feed_readings` and `amr_validated_readings` are largely append-only, with lots of updates, the default postgres settings in RDS won't ensure that the autovacuum runs as frequently as needed.

This PR is an experiment in tuning [the autovacuum settings](https://www.postgresql.org/docs/current/runtime-config-autovacuum.html) for these two tables to avoid the need for manual vacuuming.

The changes are:

* Update the `VacuumService` so that it has the option of doing an ANALYSE separately to the VACUUM. For statistics we only need an ANALYSE which can be quicker. If we need to continue to run manual vacuums, then we'll want to do only ANALYSE
* Removes use of the service where its currently called after the daily data loads and via the manual data uploads run via admins
* Applies per-table settings to tune the vacuuming. The settings switch from triggering the processes based on a % of the table being changed to using a simple record count threshold. For tables that just grow over time a % approach isn't the best option. The initial values have been guessed based on looking at the current insert/update counts over a 24 hour period. They may need tweaking if the processes run too frequently.

The plan is to release this to test to try and assess impacts before pushing to live. If the settings make things worse we can roll back to manual vacuuming.

### DIFF
--- a/app/jobs/manual_data_load_run_job.rb
+++ b/app/jobs/manual_data_load_run_job.rb
@@ -24,7 +24,6 @@ class ManualDataLoadRunJob < ApplicationJob
       amr_uploaded_reading.update!(imported: true)
       manual_data_load_run.info("Inserted: #{amr_data_feed_import_log.records_imported}")
       manual_data_load_run.info("Updated: #{amr_data_feed_import_log.records_updated}")
-      vaccuum_for_large_imports(amr_data_feed_import_log.records_imported, amr_data_feed_import_log.records_updated)
       manual_data_load_run.info("SUCCESS")
       status = :done
     rescue => e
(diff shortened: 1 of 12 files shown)


iant (approved):
:+1:

PR #3166, opened 2023-10-20
Lock users from archived schools

iant:


lib/tasks/deployment/20231020160616_lock_users_for_archived_schools.rake
```diff
@@ -0,0 +1,37 @@
...
+          user.contacts.for_school(school).first&.destroy
+          user.lock_access!(send_instructions: false)
+        end
+      end
+    end
+
+    School.deleted.each do |school|
+      school.transaction do
+        school.users.each do |user|
+          next if user.has_other_schools?
```
ldodds, line 22: We don't skip here if the we're archiving?

lib/tasks/deployment/20231020160616_lock_users_for_archived_schools.rake
```diff
@@ -0,0 +1,37 @@
...
+    puts "Running deploy task 'lock_users_for_archived_schools'"
+
+    School.archived.each do |school|
+      school.transaction do
+        school.users.each do |user|
+          next if user.has_other_schools?
+          next if user.locked_at?
+
+          # Lock account if user is linked to only this school
+          user.contacts.for_school(school).first&.destroy
```
ldodds, line 13: We don't do this if we're archiving?

ldodds (changes requested):
I don't think the logic quite matches what is in the SchoolRemover, can you double check?

PR #3167, opened 2023-10-23
Recommendations page: prompt to complete audit

urbanwide:
As well as the prompt, this changes the current audit prompt to only consider published audits (currently can return unpublished ones too).

Please can you also review the text below - with added bonus points:

Screenshot 2023-10-23 at 16 45 52

Thank you!


app/models/audit.rb
```diff
@@ -41,6 +41,14 @@ class Audit < ApplicationRecord
   scope :published, -> { where(published: true) }
   scope :by_date,   -> { order(created_at: :desc) }
 
+  def completed_activity_types
+    activity_types.where(id: school.activities.where(happened_on: created_at..).pluck(:activity_type_id))
```
ldodds, line 45: I didn't know about this syntax!

app/services/audits/audit_service.rb
```diff
@@ -6,8 +6,12 @@ def initialize(school)
       @school = school
     end
 
+    def last_audit
+      @school.audits.published.order(created_at: :desc).first
+    end
+
     def recent_audit
-      @school.audits.where("created_at >= ?", 90.days.ago).order(created_at: :desc).first
+      @school.audits.published.where("created_at >= ?", 90.days.ago).order(created_at: :desc).first
```
ldodds, line 14: Now I've seen the other syntax, should we be using: `.where(created_at: 90.days.ago...)`? Although easy to miss the double dots I guess.

spec/services/audits/progress_spec.rb
```diff
@@ -0,0 +1,84 @@
+require 'rails_helper'
+
+describe Audits::Progress do
```
ldodds, line 3: Minor thing, I've been tagging by service system specs with `type: :service` in case we later want to be able to filter them, or run just specific types of tests. Maybe something for the style guide?

ldodds (approved):

ldodds:
Looks good @urbanwide left a couple of minor comments. I learned a bit of syntax, and then wondered whether we should be using it more elsewhere.

app/models/audit.rb
urbanwide: It's fairly new to me too, only found it a few weeks ago! I quite like it.

app/services/audits/audit_service.rb
urbanwide: Yes - I'll change it now. Might just be 2 dots though for >= as 3 dots would be >

PR #3169, opened 2023-10-24
Re-enable overcommit forbidden branch hook

iant:
This PR enables the forbidden branche(s) pre-commit hook to prevent committing to master.
It can be overidden by `SKIP=ForbiddenBranches git commit -m "my commit message" `


ldodds (approved):
OK, lets try this. Do we need to do anything to enable it in our local configs?

iant:
@ldodds no, shouldn't need to do anything to enable it

PR #3170, opened 2023-10-24
Fix flickering scorable test

iant:
See https://github.com/Energy-Sparks/energy-sparks/actions/runs/6624236158/job/17992797183?pr=3166

### DIFF
--- a/spec/support/shared_examples/scorable.rb
+++ b/spec/support/shared_examples/scorable.rb
@@ -3,8 +3,10 @@ RSpec.shared_examples "a scorable" do
     let!(:schools) { (1..5).collect { |n| create :school, :with_points, score_points: 6 - n, scoreboard: scoreboard, school_group: school_group, activities_happened_on: 6.months.ago, template_calendar: subject.scorable_calendar}}
 
     it 'returns schools in points order' do
-      expect(subject.scored_schools.map(&:sum_points)).to eq([5, 4, 3, 2, 1])
-      expect(subject.scored_schools.map(&:id)).to match_array(schools.map(&:id))
+      travel_to subject.previous_academic_year.start_date do
+        expect(subject.scored_schools.map(&:sum_points)).to eq([5, 4, 3, 2, 1])
+        expect(subject.scored_schools.map(&:id)).to eq(schools.map(&:id))
+      end
     end
 
     context 'with academic years' do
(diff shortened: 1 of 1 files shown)


spec/support/shared_examples/scorable.rb
```diff
@@ -3,7 +3,7 @@
     let!(:schools) { (1..5).collect { |n| create :school, :with_points, score_points: 6 - n, scoreboard: scoreboard, school_group: school_group, activities_happened_on: 6.months.ago, template_calendar: subject.scorable_calendar}}
 
     it 'returns schools in points order' do
-      expect(subject.scored_schools.map(&:id)).to eq(schools.map(&:id))
+      expect(subject.scored_schools.map(&:id)).to match_array(schools.map(&:id))
```
urbanwide, line 6: I have just read that match_array doesn't care about the order of the passed array - but this is a test that is checking for things being in a certain order (points order?) so should care about the order of things returned? So is it the right thing to use here?

ldodds (changes requested):
See discussion

spec/support/shared_examples/scorable.rb
iant: ok looks like the shared example test is (was?) returning only `nil` `sum_points` for all schools (see failing tests in [ebea34c] https://github.com/Energy-Sparks/energy-sparks/actions/runs/6626093231/job/17998325162?pr=3170 where i've added a new expectation for the sum points ordering) 

adding a travel to for the previous academic year gets it passing again (see latest commit) - so related to the scoreboard `academic_year_calendar` used in `scoreboard_spec.rb` and `school_group_spec.rb` - but unclear if this is the correct fix here  @ldodds ??

ldodds: Yes I think this is OK. 

It looks like the scorable calendar is created with the `:template_calendar` factory which creates an academic year with term times in the future. The scorable shared examples create activities for the schools that have happened in the past.

It looks like there's a time/date issue with the mix of calendars, calendar event types, academic years and the timestamps for the activities which ends up with the activities being part of the previous academic year, rather than the one associated with the calendar. This might only affect the test at particular times of the year?

The travel to looks reasonable, but I'm also wondering if that might also have issues at different times. We could try this and revisit.

An alternative is to setup the activities differently to `activities_happened_on: 6.months.ago` and ensure they're in the current academic year for the template calendar?

ldodds (approved):

PR #3172, opened 2023-10-24
Update rubocop config

urbanwide:
Rubocop automatically disables cops in the todo file by default if there are too many files affected. Have regenerated the todo with an exclude-limit of 1000 which means the todo list should exclude specific files with offences rather than turning the cop off all together.

Have also switched on the it_behaves_like cop so we only use it_behaves_like rather than it_should_behave_like. There is no cop to prefer it_behaves_like over shared_examples (as they are both slightly different I suppose).


ldodds (approved):

.rubocop.yml
```diff
@@ -373,10 +379,37 @@ RSpec/ScatteredLet:
...
-    - 'spec/system/schools/advice_pages/hot_water_spec.rb'
+  AllowedPatterns:
+    - _£_
 RSpec/ScatteredSetup:
   Enabled: false
+RSpec/ItBehavesLike:
+  EnforcedStyle: 'it_behaves_like' # rather than 'it_should_behave_like'
+RSpec/MultipleExpectations:
+  Max: 12 # max expectations per test
+  Exclude:
```
ldodds, line 390: Are these manually added, rather than generated via `auto-gen-config`?

urbanwide: Yes - because of wanting a config of Max: 12 (for now), and .rubocop.yml inheriting from  .rubocop_todo.yml, it seemed the only way to do it as far as I can tell.

PR #3174, opened 2023-10-25
Change loading period for sheffield solar

ldodds:
Every day we load data from the PV Live API. Currently we (re)load a 10 day period that starts 2 days before the job runs (`Date.yesterday - 1`). This gives us a 1 day lag in our data, but the API has data up until yesterday.

This PR changes the loading window to be the previous 7 days.


iant (approved):
:+1: looks good to me @ldodds

PR #3176, opened 2023-10-25
3829 switch on onboarding reminders

ldodds:
Branched from #3175 so merge that first.

Remove feature flag around sending onboarding reminder emails.


urbanwide (approved):
Thanks for doing this @ldodds. Just noticed the Welsh also has translations for other plurals i.e. .two as well as the usual ones such as .other and .one, so I have changed the view to use the count / plural functionality (which it probably should have done in the first place). Not sure if you want to review again just to check?

PR #3177, opened 2023-10-25
Remove scheduled upload of unvalidated meter collections to S3

ldodds:
We have a rake tasks that pushes a subset of school data to S3. The meter collections have been used for doing bulk testing of the analytics. We are no longer doing this on a regular basis, as we're moving towards more rspec based testing. 

As the job involves a big spike in CPU and memory usage to generate and then upload the data to S3 I'm disabling the job.

I've retained the rake task so we can re-run on demand if we do find we need access to meter collection files in bulk. It's still possible to manually download data for specific schools from the test and production servers.


iant (approved):
:+1:

PR #3178, opened 2023-10-25
Add new data config for Bryt

iant:


lib/tasks/deployment/20231025142201_create_bryt_data_feed_config.rake
```diff
@@ -0,0 +1,24 @@
+namespace :after_party do
+  desc 'Deployment task: create_bryt_data_feed_config'
+  task create_bryt_data_feed_config: :environment do
+    puts "Running deploy task 'create_bryt_data_feed_config'"
+
+    config = {}
```
ldodds, line 6: Looking at the example the file row in the worksheet is a "Grand Total" row, so I think we need to configure the `column_row_filters` to exclude that, to avoid errors/warnings.

ldodds (changes requested):

lib/tasks/deployment/20231025142201_create_bryt_data_feed_config.rake
iant: :+1: 
I've pushed an update to add a column row filter added to ignore the 'Grand Total' row.  Have tested via `http://localhost:3000/admin/amr_data_feed_configs` and all looks ok to me (no warning and first data row showing in the preview)
    `config['column_row_filters'] = { 'Mpan' => 'Grand Total' }`

ldodds (approved):

PR #3179, opened 2023-10-25
Remove ability to manage dark sky areas

ldodds:
Tidying up a bit of technical debt.

The Dark Sky weather service shut down earlier this year. We've kept the data as it is still used by some schools.

This PR:

- removes the ability to create, edit and add new areas
- adds a note that this is old data to the admin page and move it off the main admin homepage
- move around some of the links to reports and CSV downloads for solar areas, weather stations to standardise on where these can be found and reduce the size of the admin/reports page


iant (approved):
:+1:

PR #3187, opened 2023-10-27
Convert strings to integers in SingleReadConverter

ldodds:
Fix bug in single read converter: the period needs to be converted to string. The specs had integer values which is why it was missed.

[Testing this has highlighted a local test failure in the same spec, which I'll investigate separately]


urbanwide (approved):
Looks good @ldodds 👍

PR #3201, opened 2023-11-01
Initial version of report for identifying engaged schools

ldodds:
An initial attempt at producing an admin report that lists "engaged schools".

The definition of an engaged school is one that has recently logged in users, or has used features of the site to record activities, actions, etc. The `:engaged` scope on the School model encompasses this. A number of other scopes have been added to other models to make it easier to build the required queries.

A new admin report page with a CSV download has been added, but depending on performance this might need to be turned into an emailed report. This prototype will be deployed to test for further review before merging.

Still to do:

- [x] Get internal feedback
- [x] Test performance as a page vs emailed report
- [x] Include cluster school users to recently active users? - no

### DIFF
--- a/app/controllers/admin/reports/engaged_schools_controller.rb
+++ b/app/controllers/admin/reports/engaged_schools_controller.rb
@@ -2,7 +2,7 @@ module Admin
   module Reports
     class EngagedSchoolsController < AdminController
       def index
-        @engaged_schools = School.engaged.by_name
+        @engaged_schools = School.engaged.joins(:school_group).order('school_groups.name asc, name asc')
       end
     end
   end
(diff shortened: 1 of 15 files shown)


app/controllers/admin/reports/engaged_schools_controller.rb
```diff
@@ -0,0 +1,49 @@
+module Admin
+  module Reports
+    class EngagedSchoolsController < AdminController
+      def index
+        respond_to do |format|
+          format.html do
+            @engaged_schools = ::Schools::EngagedSchoolService.list_engaged_schools
```
urbanwide, line 7: `` @engaged_schools = ::Schools::EngagedSchoolService.list_engaged_schools``
 is done twice in index and could just be done once at the top of the method?

urbanwide (approved):
Looks good @ldodds - just one small comment

PR #3203, opened 2023-11-07
Fix column alignment on meters table

ldodds:
Fix the colspan for admins, otherwise headings are misaligned.


urbanwide (approved):
Looks good @ldodds 👍

PR #3207, opened 2023-11-08
Allow filtering by funder on school comparison pages

ldodds:
Adds a new admin-only option on the "Choose country" tab of the school comparison page.

The new select box and filter allows an admin to further refine the list of schools to just those associated with a specific funder.

As the admins need to also narrow by country and school type it seemed better to add this as a new option rather than a new tab.

### DIFF
--- a/spec/system/compare_spec.rb
+++ b/spec/system/compare_spec.rb
@@ -68,7 +68,7 @@ describe 'compare pages', :compare, type: :system do
     it { expect(page).to have_link('Change benchmark') }
   end
 
-  shared_examples "a form filter" do |id:, school_types_excluding: nil, school_type: nil, country: nil, school_groups: nil|
+  shared_examples "a form filter" do |id:, school_types_excluding: nil, school_type: nil, country: nil, funder: nil, school_groups: nil|
     let(:all_school_types) { School.school_types.keys }
 
     it 'has school_type checkbox fields', if: school_types_excluding do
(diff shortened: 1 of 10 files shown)


spec/system/compare_spec.rb
```diff
@@ -390,6 +401,65 @@
...
+        it_behaves_like "a benchmark list page"
+        it_behaves_like "a filter summary", country: "Scotland", school_types_excluding: ['middle']
+        it_behaves_like "a filter summary", funder: "Grant Funder"
+
+        context "Changing options" do
+          before { click_on "Change options" }
+
+          it_behaves_like "an index page", tab: 'Choose country', show_your_group_tab: false
+          it_behaves_like "a form filter", id: '#country', country: 'scotland', school_types_excluding: ['middle']
+          it_behaves_like "a form filter", id: '#country', funder: "Funded by Grant Funder"
```
urbanwide, line 441: Should the id for this be '#funder'?

urbanwide (approved):
Looks good @ldodds - just one small comment about a test

spec/system/compare_spec.rb
ldodds: It's checking within the countries tab (`#country`) for the expected state of the funder select box, so this is correct (and passes).

urbanwide: Sorry yes I can see that now 👍 Was a bit of a swift review, apologies

PR #3208, opened 2023-11-08
Additional recommendations page style updates

urbanwide:
Thanks for the feedback @ldodds. Have centred the cards for the recommendations panel anyway, so we can use it that way in the future. Also added the new link and with icon, as I think it might look odd to have 4 cards at the bottom if there are 5 shown in the other two panels. Also refactored a couple of other things. Summary of the changes in this PR:

* Changes labels to pupil activities and adult actions
* Switches to using 5 recommendations / cards wide
* Centres the recommendations cards
* Add recent alerts link with magnifying glass icon
* rename recommendations component max_lg to limit_lg.
* pass standard display options through to recommendations component rather than individually each time


ldodds:
I think this looks betters with the 5 items on each row.

If you can't get the CSS wrangled to use 4 items on final row, you could add make the fifth tile a link to their "recent alerts" list, e.g "View recent alerts for your school to identify some immediate priorities" Maybe use the magnifiying glass icon from the "Detective" category?

ldodds (approved):

ldodds (approved):

PR #3210, opened 2023-11-08
Ensure activation date is set

ldodds:
Schools have an "activation date" attribute which can be edited by admins to set the date the school became active on energy sparks. This date is used by the analytics as the start date for some of the school comparison reports (specifically, the report looking at change in energy use since activation).

If the activation date is nil then the analytics defaults to the school record creation date.

In some cases this can be quite different. A few schools have taken a number of months before they've become data enabled.

This PR does 2 things:

- set the activation date when a school first becomes data enabled
- retrospectively sets the activation dates for existing schools to the date they first became data enabled, where we have that information

This will improve some of the historical reporting.


urbanwide (approved):
Looks good @ldodds 👍

PR #3211, opened 2023-11-09
Fix redkit calendars

ldodds:
Redkit school group schools have been setup with the wrong calendar.

This PR does a database update to:

- update the template calendar linked to the school
- change the schools calendar to be based on the correct one (Yorkshire)
- reset all the school calendars to match the Yorkshire one

I've run this locally and it runs OK and applies the required changes. There are already tests around the calendar reset service.


urbanwide (approved):
Looks good @ldodds 👍

PR #3214, opened 2023-11-10
Allow manual load for solar edge

ldodds:
Provides an option for admins to request an immediate load from a SolarEdge API. This makes it quicker to set-up, test and then finish configuration of the data feed. Otherwise they have to wait until the next day for data to be available and they can review the school configuration.

With this change they can run the loader immediately. Its run as a background job as there could be several years of data to fetch and update.

This PR:

- Adds a new `SolarEdgeLoaderJob` that will run as a background job, trigger the data load and then email the admin with the results
- A new mailer for reporting on the success or failure of manually run data solar data loads
- A new "Run Loader" button on the list of solar edge installations to trigger the load
- A set of tests for the job that confirms that the mails are generated as expected (the loading code has its own tests)
- A basic system spec to test the button triggers creation of a job

I've extracted most of the code that does the running of the job into a base class. In a later PR I'll extend the ability to run these jobs to the 2 other types of feed we have configured.

The mailer is also generic and can cope with any of the "installation" models as they have a similar set of properties and relationships. The PR adds `#display_name` for use in the emails.


app/jobs/solar/base_solar_loader_job.rb
```diff
@@ -0,0 +1,47 @@
...
+      @installation = installation
+      @upserter = upserter(start_date, end_date)
+      @upserter.perform
+      send_notification(notify_email)
+    rescue => e
+      Rollbar.error(e, job: :import_solar_edge_readings)
+      send_failure_notification(notify_email, e)
+      false
+    end
+
```
urbanwide, line 22: I'd be tempted to do add something like this for upserter and solar_feed_type methods to ensure they are implemented in subclasses:

  ```
  private

  def upserter(start_date, end_date)
    raise "Not implemented"
  end

  def solar_feed_type
    raise "Not implemented"
  end
```

urbanwide (approved):
Looks good @ldodds - small comment about overriding methods - would be tempted to add "interface" methods to the base class to ensure subclasses implement expected methods used in the same class... (see inline comment).

app/jobs/solar/base_solar_loader_job.rb
ldodds: Good call. I've never been sure about the best way to do this in Ruby, but having some kind of stub makes sense.

urbanwide: Looks good @ldodds 👍

PR #3215, opened 2023-11-10
Move Transport Survey classes to their own directory

urbanwide:
Have had a go at moving the transport survey classes in to their own directory as a test case - it's up for discussion really!
* This moves TransportSurveyResponse to TransportSurvey/Response. This feels right as a response belongs to transport survey.
* Also moved TransportType to TransportSurvey/TransportType - not entirely convinced this bit feels right though as transport type does not belong to transport survey in terms of the relationship but it is definitely exclusive to transport surveys so maybe it is right. See what you think @ldodds. 
* Have kept TransportSurvey in the root as it wouldn't feel right scoped to TransportSurveys::TransportSurvey or similar
* I am going to leave moving interventions/activities for another time / branch. 

Some reading: https://blog.makandra.com/2014/12/organizing-large-rails-projects-with-namespaces/


ldodds:
This looks good to me @urbanwide. Useful to see that you need to update the references in the controller and ability for it all to work.

ldodds (approved):

PR #3216, opened 2023-11-10
Allow manual loads for all solar feeds

ldodds:
Extends the features added in #3214 to the other types of solar feeds.

So each of the three sections on the manage Solar API feeds page for a school now has:

- check API button for testing API access
- "Run Loader" button for running a job to load data for that data source

I've refactored the tests for the solar edge job to allow reuse of examples and contexts to make it easier to add the specs for the 2 new jobs. I've also added system specs for the new buttons


urbanwide (approved):
Looks good @ldodds 👍

PR #3219, opened 2023-11-13
Adult and pupil dashboards: add a prompt to complete recommended pupil or adult led activities

urbanwide:
The main point of this PR is to a link to the recommendations page from both adult and pupil dashboards. It does more though...

**Adult Dashboard**

* Adds a single prompt to the recommendations page
* Reworks the prompt spec to make it usable for the pupil dashboard prompts spec

**Pupil Dashboard:**

* Adds a single prompt to the recommendations page with 3 variations
* Also adds the prompt to complete a programme as discussed
* Have added a bunch of spec to check the prompt behaviour. Worth noting that current behaviour on this page is that all the prompts are visible regardless of if a user is logged in or not. Some of the prompts redirect to a login page when clicked, so is not ideal. For now, I have added these two prompts in the same way but we ought to revisit this when we discuss the adult dashboard (or sooner).

Also note I used include_examples rather than it_behaves_like as the test descriptions didn't make sense / were too long otherwise.

### DIFF
--- a/app/controllers/pupils/schools_controller.rb
+++ b/app/controllers/pupils/schools_controller.rb
@@ -35,6 +35,7 @@ module Pupils
       @show_temperature_observations = show_temperature_observations?
       @observations = setup_timeline(@school.observations)
       @default_equivalences = default_equivalences
+      @programmes_to_prompt = @school.programmes.last_started
     end
 
     def setup_data_enabled_features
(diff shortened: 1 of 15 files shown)


urbanwide:
Thanks @ldodds - do you mind just checking over the last commit - I forgot to add a test to check the pupil prompt switched correctly between the 3 options.

ldodds (approved):

PR #3222, opened 2023-11-15
Add additional British Gas portal configs

ldodds:
Two new configurations one for gas meters and one for electricity meters.

The basic spreadsheet formats are the same, the only significant difference for us is the use of a column called "MPAN" for electricity and "MPRN" for gas. These are the correct labels for the identifiers, but means we have to have 2 configurations.

There are some other differences in the Excel files but not that impacts our data loading currently.


urbanwide (approved):
👍

PR #3223, opened 2023-11-15
Remove :event Observation

ldodds:
We were creating an `:event` observation type for schools when their activation date was set. This was originally used to add a message "X school has joined Energy Sparks" and later "X became an active user of Energy Sparks" to the school timeline.

But this hasn't been working properly as activation dates weren't being routinely set until this PR was completed: https://github.com/Energy-Sparks/energy-sparks/pull/3210.

This has lead to a number of schools having recent entries in their timeline, which is incorrect. The callback to add the Observation in the School class should have been using the activation date, not the current date.

As this feature of the timeline hasn't been working correctly for some time, and because its not adding much value, we've decided to just remove the events, rather than fix the data and the School model callback.

So this PR:

- alters the Observation class to use the Hash syntax for the enum, to remove `:event`
- adds an AfterParty task to remove the records with an observation_type of "3" (the events)
- removes the template to show the timeline entry

### DIFF
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -47,8 +47,10 @@ class Observation < ApplicationRecord
   belongs_to :programme, optional: true
   belongs_to :audit, optional: true
   belongs_to :school_target, optional: true
+
   # If adding a new observation type remember to also add a timelime template in app/views/schools/observations/timeline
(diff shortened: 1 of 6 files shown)


urbanwide (approved):
👍

PR #3224, opened 2023-11-15
Add task to load funders

ldodds:
Adds an after party task to load contents of a CSV file to populate list of funders.

This is a one-off, if we do this again we'll provide a better starting point for matching to avoid looking for schools/funders by name.


urbanwide (approved):
Looks good @ldodds 👍

PR #3228, opened 2023-11-16
Replace prompt on activity & action categories pages

urbanwide:


ldodds:
Code changes all look good, but testing this locally the notice on the activities page doesn't fit the full width of the page. Although the one on the action page does:

Screenshot from 2023-11-17 15-50-24

The difference is that its wrapped in a bootstrap row on the activities page but not the actions.

urbanwide:
I added the bootstap row around on the activities page notice as a quick way of aligning it with the blocks below with the bg-light background. It did work on my machine 🤔

I've reworked it now though and put a container around the other blocks instead. I did try other more-elegant ways but alas to no avail.

Hopefully this is ok!

ldodds (approved):
looks good :+1:

PR #3233, opened 2023-11-20
Fix tariff editor page calendar language

urbanwide:
Looks like window.onload overwrites any previous settings of itself - this meant that when the new panel switcher component was added, which also used window.onload, it was overwriting the window.onload for the date picker form component - which is also where it's locale was set.

Fortunately we can use window.addEventListeners instead which appends rather than overwrites
(Jquery is not available at the point this is called, so we use full fat JS)


ldodds (approved):
Nice catch here. Using an event listener looks like a better approach anyway.

PR #3240, opened 2023-11-22
Ensure climate change levy displays on cost table

ldodds:
Pulls in the changes from: https://github.com/Energy-Sparks/energy-sparks_analytics/pull/636.

Also updated the meter table component and translation keys so that the Climate Change Levy costs display correctly.


urbanwide (approved):
All good @ldodds 👍

PR #3241, opened 2023-11-22
Add another special case for unvalidated reading report

ldodds:
Add another format specific check to the unvalidated readings report.

I'm not sure if the cause here is the date format being updated and so historical data is now problematic. Or recent data was loaded with an unexpected date format.

Need to tackle the longer term fix of actually storing the parsed dates on the unvalidated readings table, not reparsing later. That warrants a separate discussion and PR.


urbanwide (approved):
👍

PR #3245, opened 2023-11-24
Attempt at fixing the fixed navbar issue

urbanwide:


ldodds:
Tidying up the partial solutions is a good idea.

I'm just wondering if there's a pure CSS fix. For example would it be possible to use a media selector to increase the margins on those elements but only for smaller screen widths?

urbanwide:
See what you're saying, I'll have a go now 👍

urbanwide:
@ldodds - see updated description

urbanwide:
Going to add a different breakpoint for when in Welsh as the menu text is way longer. I'll update in the morning 👍

urbanwide:
More changes @ldodds - have updated the description too. Leaving it as draft for now.

ldodds:
@urbanwide I've tested this locally and looks good to me. 

I did notice that the top nav bar does overflow the page width at times, but that might be due to there being 2 locale buttons in my test environment which makes that worse.

Otherwise I didn't notice any breakage across the portions of the site I've looked at (activity, schools  + forms, advice pages, etc). I also checked the Welsh version and seemed OK to me.

I didn't notice any visual differences in the login/out buttons so might have missed something.

I think this is an improvement and it's probably worth getting this released before we decide to do anything else. We can go back to the user who originally had the issue to ask for feedback. Tidying up the navigation first might be the best option.

ldodds (approved):

urbanwide:
This is the difference in login buttons between a blue bar top menu bar vs a black bar top menu bar. The blue bar version has an enrol button too. I've maintained how it worked before so all good but just so you know what the difference is:

Screenshot 2023-11-30 at 10 29 58
Screenshot 2023-11-30 at 10 29 48

PR #3252, opened 2023-11-30
Fix link in alert emails

ldodds:
The email alerts still have a link to the old advice page. This PR updates the alert template to fix the link and add explicit assertions for the expected urls to the spec, not just checks for the link text.


urbanwide (approved):
Looks good @ldodds 👍

PR #3253, opened 2023-11-30
try ruby 3.0

tbhi:
- use ruby 3 compatible analytics
- upgrade cancancan for ruby 3 compatibility
- function argument changes
- BigDecimal.round(0) difference
- update to work with latest selenium-webdriver


tbhi:
replaced by https://github.com/Energy-Sparks/energy-sparks/pull/3258

PR #3259, opened 2023-12-04
test flake at spec/services/programmes/creator_spec.rb:75

tbhi:
looks like `happened_on` defaults to yesterday so would create 2 activities with the same date and then sometimes pick the wrong one for the test


urbanwide (approved):
Looks good @tbhi 👍

PR #3261, opened 2023-12-05
Switch to font-awesome gem

ldodds:
As part of the Ruby 3 upgrades we've switched to the font-awesome-sass gem to work around some issues with webpacker and compiling assets.

This PR applies those changes to the main project ahead of the release.

- adds the gem and references it in the application scss file
- removes some unneeded files and imports for the node version, inlining some styles to simplify the configuration


tbhi (approved):
:+1:

PR #3267, opened 2023-12-06
Remove axis choices for thermostatic analysis charts

ldodds:
The thermostatic analysis charts plot consumption (kWh) against temperature. This is to show how the linear regression has associated usage across different temperatures with periods when the heating is on or off.

The charts don't support viewing the data as cost or co2 emissions vs temperature.

This PR ensures that the chart axis controls are disabled for these charts on the gas and storage heater analysis pages.


tbhi (approved):
makes sense, hesitant about negative expectation but think makes sense in this case

PR #3269, opened 2023-12-06
Add correct variables to solar pv page

ldodds:
When a school has solar pv we show the benefits of having the solar panels.

There's currently a bug in this version of the page. The table summarising the benefits if the school owns the panels should include how much energy the panels produced, as well as an estimate of the export value of that energy they've exported.

Currently the page just has hard-coded figures: a mistake from when the pages were rebuilt.

This PR replaces the hard-coded text with the actual values and updates the description of the Solar Export Guarantee with new text. We don't know the actual price that a school may be getting, so we now just show our estimate along with an indication of the likely range.

I've updated the system specs to check for these values.


tbhi (approved):
makes sense

PR #3272, opened 2023-12-07
Knapsack tweaks

tbhi:
- use knapsack for all tests
- add new job for generating knapsack timings
- simplify existing workflow
- use node caching
- increase workers to 16
- compatibility fix for new chrome versions


ldodds (approved):
Looks good.

If I understand the changes correctly, the knapsack report can be generated via a manually triggered workflow, then the updated file can be added to the project. Can you add a note about doing that somewhere?

PR #3274, opened 2023-12-07
Try autocommiting knapsack report updates

tbhi:


tbhi:
replaced with https://github.com/Energy-Sparks/energy-sparks/pull/3282

PR #3275, opened 2023-12-07
Ensure the check for axis controls does something

ldodds:
Adds a check for hidden element, as the axis controls have display: none initially.


tbhi (approved):
:+1:

PR #3286, opened 2023-12-11
Move programme observation

urbanwide:
Move observations for programmes to use the new polymorphic relationship.
Keeps the old programme_id column in the table for now, will remove all these when they are all moved.

Worth noting that I don't think there are any observations of these types in the database - has nobody ever completed a programme? Or maybe it's not working (there are tests which indicate the code should be working)?

### DIFF
--- a/db/migrate/20231207104401_move_programme_observation.rb
+++ b/db/migrate/20231207104401_move_programme_observation.rb
@@ -5,13 +5,9 @@ class MoveProgrammeObservation < ActiveRecord::Migration[6.0]
       observation.observation_type = :observable
       observation.save!
     end
-
-    remove_column :observations, :programme_id
   end
 
   def down
-    add_reference :observations, :programme, index: true
-
     Observation.where(observable_type: 'Programme').find_each do |observation|
       observation.programme = observation.observable
       observation.observable = nil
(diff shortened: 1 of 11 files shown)


db/migrate/20231207104401_move_programme_observation.rb
```diff
@@ -0,0 +1,18 @@
+class MoveProgrammeObservation < ActiveRecord::Migration[6.0]
+  def up
+    Observation.where(observation_type: :programme).find_each do |observation|
+      observation.obserable = observation.programme
```
ldodds, line 4: There's a typo here I think?

ldodds (changes requested):
Looks good apart from the migration

ldodds:
There's one Programme with a `:completed` status, so there should be an observation for it?

db/migrate/20231207104401_move_programme_observation.rb
urbanwide: Oh gosh, good spot, thank you. Strange that it ran (or perhaps I changed it afterwards).

urbanwide:
I couldn't see one - however it looks like the programme observation feature was added this July, after the programme was completed. I couldn't find an after party task to add retrospective programme observations in, want me to do this?

ldodds:
This was obviously missed! If you can add a task for that as a separate PR, that would be great. Would also then ensure your migration gets exercised

ldodds (approved):

PR #3287, opened 2023-12-11
Make a component for displaying podium and scoreboard activity on dashboards

urbanwide:
Previews available at:
http://localhost:3000/admin/components/previews/scoreboard_summary_component/school_with_points
http://localhost:3000/admin/components/previews/scoreboard_summary_component/school_without_points

Latest screen grab showing a school without points on the podium:
Screenshot 2023-12-13 at 17 05 22

Not sure if they should have a crown though if they have no points @ldodds ?


ldodds:
I can't see the schools in the preview (just the timeline). I've got a newer dump of production and the test school you're using has been archived as they've left. I wonder if you can just find the first active school with or without some points this year for the preview?

Otherwise I think this looks good. 

Looking at an actual school with zero points, I see the actual podium positions for the bottom 2 schools. I assume your preview school just happened to be on a scoreboard with only 2 scoring schools.

I agree I don't think they should get a crown if they don't have any point.

On the scoreboard page we don't show a podium position for schools without scores, they are all unranked. But here we are showing them one rank lower than the lowest. I think we should remove both their position and the crown in the case where there's no points for consistency.

Not reviewed the code in detail, but looks good to me.

urbanwide:
Thanks Leigh

Have removed the crown and points for schools without points on a podium - that makes sense, thanks!
Also have hopefully made the previews more useful
Have written a load more spec too
Have moved observations from timeline component in to their own component and tested compact view
It's a bit of a beast of a PR sorry!

app/components/podium_component/podium_component.html.erb
```diff
@@ -1,15 +1,28 @@
...
+  <% if podium.includes_school? && podium.school_has_points? %>
+    <% if EnergySparks::FeatureFlags.active?(:activities_2023) %>
+      <h4 class="text-center"><%= t('components.podium.full_position_html', position_ordinal: podium.school_position.ordinal, scoreboard: podium.scoreboard.name, scoreboard_path: scoreboard_path(podium.scoreboard)) %>.</h4>
+    <% else %>
+      <h4 class="text-center"><%= t('components.podium.position_html', position_ordinal: podium.school_position.ordinal) %>.</h4>
+    <% end %>
+  <% else %>
+    <h4 class="text-center"><%= t('components.podium.no_points_this_year') %></h4>
+
+    <p class="text-center"><%= t('components.podium.complete_an_activity') %>.
```
ldodds, line 11: Suggestion: link the "complete an activity" text to the school recommendations page, so there's a direct link to that page from the prompt.

app/components/scoreboard_summary_component.rb
```diff
@@ -0,0 +1,44 @@
...
+  def scoreboard
+    podium.scoreboard
+  end
+
+  def limit
+    5
+  end
+
+  def observations
+    scope = other_schools? ? scoreboard.observations : Observation
```
ldodds, line 29: Defaulting to any activity across EnergySparks is fine, but just noting we might want to review whether we want to fall back to what other schools in their group are doing as a more directly relevant set of activities. That can come in the new year though after we've got some feedback.

ldodds (approved):
This looks great. Have tested out with a range of schools and all looks OK. Nice use of components and thorough set of tests. I left one minor suggestion to add a link, otherwise good to merge.

app/components/podium_component/podium_component.html.erb
urbanwide: Good suggestion, thank you.

app/components/scoreboard_summary_component.rb
urbanwide: OK, sounds like a plan

PR #3291, opened 2023-12-12
Ensure we show correct progress summary

ldodds:
There's a bug on the school target progress report page. The school targets controller was using the ProgressService to create the progress summary to display for the target.

For a current target this is fine, as the ProgressService loads that. But for an expired target it means we show the current progress, not the summary for the expired target.

The functional change in this PR is to update the controller to just call `.to_progress_summary` on the target.

I've also updated the specs to:

- add a test for this situation to ensure the bug is fixed
- improve the tests for the progress reporting so that it checks explicitly for the expected summary for gas and electricity data
- improve the tests for the fallback when we have limited data so its checking what data is shown and not just whether some text is displayed
- improved the names of some of the contexts and checks

### DIFF
--- a/spec/system/schools/school_targets_spec.rb
+++ b/spec/system/schools/school_targets_spec.rb
@@ -205,9 +205,9 @@ RSpec.shared_examples "managing targets", include_application_helper: true do
         it { expect(page).not_to have_link("View results") }
 
         context 'and there are progress reports for both targets' do
-          let!(:expired_electricity_progress) { build(:fuel_progress, fuel_type: :electricity, progress: 0.99, target: 20, usage: 15) }
-          let!(:expired_gas_progress)         { build(:fuel_progress, fuel_type: :gas, progress: 0.59, target: 19, usage: 17) }
-          let!(:expired_last_generated)       { Date.yesterday }
+          let(:expired_electricity_progress) { build(:fuel_progress, fuel_type: :electricity, progress: 0.99, target: 20, usage: 15) }
+          let(:expired_gas_progress)         { build(:fuel_progress, fuel_type: :gas, progress: 0.59, target: 19, usage: 17) }
+          let(:expired_last_generated)       { Date.yesterday }
 
           let(:expired_target) do
             create(:school_target, school: test_school,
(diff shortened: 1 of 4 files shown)


spec/system/schools/school_targets_spec.rb
```diff
@@ -201,6 +204,49 @@
...
+          it "links to expired progress report" do
+            #Extra check for debugging flickering test
+            expect(Schools::Configuration.count).to be 1
+            expect(School.first.has_electricity?).to be true
+            expect(page).to have_link("View monthly report", href: electricity_school_school_target_progress_index_path(test_school, expired_target))
+          end
+
+          it 'shows progress summary' do
+            within('#electricity-row') do
+              expect(page).to have_content("-2.0%") #target
```
tbhi, line 240: Could be helpful to reference expired_electricity_progress.progress and expired_target rather than strings?

spec/system/schools/school_targets_spec.rb
```diff
@@ -201,6 +204,49 @@
 
         it { expect(page).not_to have_link("View results") }
 
+        context 'and there are progress reports for both targets' do
+          let!(:expired_electricity_progress) { build(:fuel_progress, fuel_type: :electricity, progress: 0.99, target: 20, usage: 15) }
+          let!(:expired_gas_progress)         { build(:fuel_progress, fuel_type: :gas, progress: 0.59, target: 19, usage: 17) }
+          let!(:expired_last_generated)       { Date.yesterday }
```
tbhi, line 210: could be within the :expired_target let? should it be the same as target_date?

tbhi (approved):

spec/system/schools/school_targets_spec.rb
```diff
@@ -146,10 +146,13 @@
       expect(page).to have_title("Your current energy saving targets")
     end
 
-    it 'links to help page if there is one' do
-      create(:help_page, title: "Targets", feature: :school_targets, published: true)
-      refresh
-      expect(page).to have_link("Help")
+    context 'when there is a help page of the right type' do
```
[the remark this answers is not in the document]
ldodds: no, I was just tidying up a bit

spec/system/schools/school_targets_spec.rb
```diff
@@ -201,6 +204,49 @@
...
+              electricity_progress: expired_electricity_progress, gas_progress: expired_gas_progress,
+              report_last_generated: expired_last_generated)
+          end
+
+          let!(:electricity_progress) { build(:fuel_progress, fuel_type: :electricity, progress: 0.90, target: 15, usage: 15) }
+          let!(:gas_progress)         { build(:fuel_progress, fuel_type: :gas, progress: 0.50, target: 7, usage: 17) }
+          let!(:last_generated)       { Time.zone.today }
+
+          let!(:target) do
+              create(:school_target, school: test_school,
```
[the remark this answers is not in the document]
ldodds: I've switched most of the `let!` to `let`. That wasn't needed for the other objects.  The target needs to be created before the visit to page, so that should be a `let!`.

I've manually re-run rubocop on the file and its not giving me any warnings.

spec/system/schools/school_targets_spec.rb
ldodds: I kept them separate as I think its easier to check the values that way, then have multiple objects built in one larger let block.

PR #3298, opened 2023-12-14
Reinstate webdrivers

ldodds:
Add the webdriver gem back in to work around chromedriver install issues.

May need to be reverted in Ruby 3.0 branch.

Previous issues:

- having to pin driver: https://github.com/Energy-Sparks/energy-sparks/pull/2922
- them update to webdrivers, removing need to pin: https://github.com/Energy-Sparks/energy-sparks/pull/2992

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -164,11 +164,11 @@ group :test do
   gem 'capybara'
   gem 'capybara-email'
   gem 'database_cleaner'
-  gem "webdrivers", '>= 5.3.0'
   gem 'selenium-webdriver'
   gem 'shoulda-matchers'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -164,6 +164,7 @@ group :test do
   gem 'capybara'
   gem 'capybara-email'
   gem 'database_cleaner'
+  gem "webdrivers", '>= 5.3.0'
```
tbhi, line 167: can we keep the same quoting and order? also a comment about removing for when merging into the ruby 3 would be useful

tbhi (approved):

PR #3301, opened 2023-12-14
Change data check to use BaseloadAnalysis class

ldodds:
If a school has just short of a years data for some meters, then the seasonal and intraweek variation analysis for those meters breaks. There's a discrepancy between the date checking between the services.

To fix this now, I've changed the BaseloadService to call the analysis Baseload::BaseloadAnalysis class directly for the individual meter breakdown. This fixes a broken page for one school so they can properly access their baseload analysis.

### DIFF
--- a/app/services/schools/advice/baseload_service.rb
+++ b/app/services/schools/advice/baseload_service.rb
@@ -199,7 +199,7 @@ module Schools
         meter = load_meter ? meter_for_mpan(analytics_meter.mpan_mprn) : nil
         seasonal_baseload_service = Baseload::SeasonalBaseloadService.new(analytics_meter, date)
         #return if there's not enough data, then return limited object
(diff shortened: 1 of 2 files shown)


app/services/schools/advice/baseload_service.rb
```diff
@@ -227,8 +227,9 @@ def seasonal_variation_rating(percentage)
 
       def calculate_intraweek_variation(analytics_meter = aggregate_meter, date = asof_date, load_meter = false)
         intraweek_baseload_service = Baseload::IntraweekBaseloadService.new(analytics_meter, date)
+        baseload_analysis = Baseload::BaseloadAnalysis.new(analytics_meter)
```
tbhi, line 230: not needed?

tbhi (approved):

PR #3306, opened 2023-12-18
Update tx config for new components

urbanwide:
Updated Transifex config after running:

`rake i18n:generate_tx_config`


tbhi (approved):
not familiar with the file but changes looks inline with everything else

PR #3310, opened 2023-12-20
National scoreboard

tbhi:


app/services/scoreboard_all.rb
```diff
@@ -0,0 +1,30 @@
...
+# scoreboard for all schools
+class ScoreboardAll
+  include Scorable
+
+  SLUG = 'all'
+
+  attr_reader :academic_year_calendar
+
+  def initialize
+    @academic_year_calendar = Calendar.find_by(title: 'England and Wales', calendar_type: 'national')
```
ldodds, line 12: This isn't ideal but I guess the alternative would be to have an attribute on the Calendar model to identify the default calendar to use for the national scoreboard, which doesn't seem much better.

Currently the national calendars can't be deleted or renamed, so this seems fairly safe.

ldodds (changes requested):
Changes look good to me, but there's a bug on the national scoreboard page.

There's some text: "These are the current scores for this academic year. [View the final scores from last year]" which links to the scores for the previous year, looks like the template needs to be updated to use the slug for the ScoreboardAll service:

`http://localhost:3000/scoreboards/%23%3CScoreboardAll:0x000055ffc98c4c30%3E?academic_year=33`

ldodds (approved):

PR #3311, opened 2023-12-21
Improve activity & action completed pages

urbanwide:
New "completed" pages for activities and actions. See what you think. 

Have put this behind a activities_2024 feature flag - we need to remove the activites_2023 one though soon!

Think the notice with the link to the action or activity could be improved (think the link should be inline with the text at least, but this is not standard behaviour for this component, so I thought I would wait to discuss this with you).

I noticed there is some text on the old page that mentions the scoreboard prizes. It is behind the scoreboard_prizes feature flag and currently switched off in production so have left it out on the new page for now.

Also - it is possible to record an action or activity and not score points (when the completed date is set in a previous academic year), so in this situation, instead of writing "Congratulations! You've just scored X points", I have put "Congratulations! We've recorded your activity" (or action).

I added an icon to the "View all suggestions" prompt on these pages, so it matched better with the other prompts above it.


ldodds:
Have taken a look at this, my initial feedback after seeing it implemented:

- I think now that the "Share what you've done..." notice ought to go below the scoreboard and other activity, to focus on the progress / activity feedback rather than the sharing
- Suggest that this notice goes under the "What do you want to do next?" heading so that section now covers all both sharing of progress as well as picking what to do next
- I agree that the notice text and link needs to be better aligned, on this page it might look better with the link as a button rather than a plain hyperlink

Other suggested changes to text, etc look good to me too.

urbanwide:
Thanks @ldodds - here's a screen grab of the bottom of the page based on the changes suggested.

I have added an icon to the share text so it matches the other prompts, see what you think:
image

ldodds (approved):

ldodds:
I think that looks good and like the icon.

PR #3316, opened 2023-12-22
Change style of disabled button on record activity / action pages

urbanwide:
This does a few things.

- Changes the style of the disabled button on record activity and action pages.
- Makes the notice that contains this form full width on the activity pages to match the style of the action pages
- Adds a tooltip to the disabled button when disabled (and removes when not)
- Changes the alignment a bit of this form

All of which is up for discussion... see what you think!


ldodds (approved):
Looks good to me. Styling is more consistent and tool tip helps clarify what is going on.

PR #3317, opened 2024-01-03
Add trait for creating previous and next academic years to calendars factory

ldodds:
Some recently added tests around the scorable concern rely on creating a calendar for the `Scorable`. The calendar needs to have some terms and a correct set of calendar years.

Currently the academic years were being created as a side effect of adding the terms. And the logic for that was just creating academic years for the previous and next _calendar_ year.

This works fine between 1st September and 31st December because the previous school year is in the previous calendar year.

But the test data is incorrect from 1st January as the previous school year begins two calendar years before that. E.g. on 1st January 2023 the school year is the Sept 2023-August 2024 period and the previous school year is Sept 2022-August 2023.

The incorrectly setup data causes three sets of specs to fail.

I've added a trait that correctly creates previous, current and latest academic years regardless of the time of year and updated the specs to use it. This should resolve the problems

### DIFF
--- a/spec/factories/calendars_factory.rb
+++ b/spec/factories/calendars_factory.rb
@@ -45,15 +45,11 @@ FactoryBot.define do
     #Ensures there is a previous, current and upcoming AcademicYear
     trait :with_previous_and_next_academic_years do
       after(:create) do |calendar, _evaluator|
-        current_year = Time.zone.today.year
+        today = Time.zone.today
         #At the end of the year, school academic year started in previous year
         #But otherwise, the previous academic year started 2 years ago
-        if Time.zone.today >= Date.new(current_year, 9, 1)
-          start_year = current_year - 1
-        else
-          start_year = current_year - 2
-        end
-        end_year = current_year + 1
+        start_year = today.year - (today.month < 9 ? 2 : 1)
+        end_year = today.year + 1
         (start_year..end_year).each do |year|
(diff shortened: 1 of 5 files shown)


spec/support/shared_examples/scorable.rb
```diff
@@ -3,6 +3,7 @@
     let!(:schools) { (1..5).collect { |n| create :school, :with_points, score_points: 6 - n, scoreboard: scoreboard, school_group: school_group, activities_happened_on: 6.months.ago, template_calendar: subject.scorable_calendar}}
 
     it 'returns schools in points order' do
+      #pry
```
tbhi, line 6: guess don't want this

spec/factories/calendars_factory.rb
```diff
@@ -42,6 +42,24 @@
...
     end
 
+    #Ensures there is a previous, current and upcoming AcademicYear
+    trait :with_previous_and_next_academic_years do
+      after(:create) do |calendar, _evaluator|
+        current_year = Time.zone.today.year
+        #At the end of the year, school academic year started in previous year
+        #But otherwise, the previous academic year started 2 years ago
+        if Time.zone.today >= Date.new(current_year, 9, 1)
+          start_year = current_year - 1
```
tbhi, line 52: `start_year = today.year - (today.month < 9 ? 2 : 1)`?

tbhi (approved):

spec/factories/calendars_factory.rb
ldodds: Yes, thats better :sweat_smile:

PR #3321, opened 2024-01-03
use composite github action to reduce duplication  

tbhi:


ldodds (approved):
Nice, I wondered whether actions allowed us to share more of the config.

PR #3322, opened 2024-01-04
Migrate to Cloudwatch agent

ldodds:
We currently push some instance metrics to CloudWatch using the deprecated Cloudwatch perl scripts. This PR removes that in favour of using the CloudWatch Agent.

The EB extension has been altered to create a suitable configuration and start the agent.

We also had a couple of scripts pushing some GoodJob metrics which are broken. One is not needed as it was just reporting number of processes, I think that was a hold-over from early testing of GoodJob, the other was reporting memory usage. I've replaced that one with a `procstat` entry in the CloudWatch agent config.

The old cronjobs are removed.

I've tested the config and the command to (re)start the agent manually, but will need to confirm that the whole release process works correctly.


.ebextensions/cloudwatch.config
```diff
@@ -1,26 +1,47 @@
...
-    perl-LWP-Protocol-https: []
-    perl-Switch: []
-    perl-URI: []
-    perl-Digest-SHA: []
-
-sources:
-  /opt/cloudwatch: https://aws-cloudwatch.s3.amazonaws.com/downloads/CloudWatchMonitoringScripts-1.2.1.zip
-
+files:
+  "/opt/aws/amazon-cloudwatch-agent/bin/config.json":
```
tbhi, line 2: I wonder about the config file location, I see things in `/opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d` - perhaps better around there? Or perhaps better to keep custom files outside aws managed directories and go in /var/app/current?

tbhi (approved):

.ebextensions/cloudwatch.config
ldodds: Yes I wondered about alternate locations. I used the current file location as it was used in this specific example:

https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/customize-containers-cw.html

Running with the current command it validates the configuration, writes to the validation log, then copies it to the amazon-cloudwatch-agent.d directory. So it could potentially start elsewhere, I guess.

The fetch-config and `-s` does a restart of the agent after installing new config, I think if we didn't use fetch-config we'd need to stop/start?

PR #3329, opened 2024-01-05
try caching of assets

tbhi:


tbhi:
don't think this is as relevant now we don't have webpacker

PR #3338, opened 2024-01-09
good job migrations

tbhi:


ldodds (approved):
Looks good. Seems odd that the schema ends up with 2 ways to reference `gen_random_uuid()`. Once like that and once as `public.gen_random_uuid()` but both seem to work on test db.

PR #3344, opened 2024-01-11
national scoreboard revisions

tbhi:


app/controllers/scoreboards_controller.rb
```diff
@@ -5,13 +5,17 @@ class ScoreboardsController < ApplicationController
...
-    @scoreboard_all = ScoreboardAll.new
+    @national_scoreboard = NationalScoreboard.new
     @scoreboards = ComparisonService.new(current_user).list_scoreboards
   end
 
   def show
-    if params[:id] == ScoreboardAll::SLUG
-      @scoreboard = ScoreboardAll.new
+    case params[:id]
+    when 'all'
```
ldodds, line 14: minor thing but the redirect could be done in the routes instead?

ldodds (approved):
Looks good. Can you do a push to transifex after merging so we can get the text translated?

app/controllers/scoreboards_controller.rb
tbhi: guess the advantage is keeping logic out of the controller but it also seems convenient to see it together here too

PR #3346, opened 2024-01-15
Remove validation on fields removed from alert type rating forms

ldodds:
A while ago we removed the "find out more" fields from the editable fields on the alert type rating forms. But the validation associated with those fields was accidentally left in place. This means it's not possible to edit the content of some ratings as the update fails validation for fields not shown to the user.

This PR removes those validation rules. 

I will separately do another, larger PR to properly retire the old "find out more" feature which is otherwise inactive.


tbhi (approved):
seems like a test would be useful to cover the not being able to edit issue

ldodds:
OK, I've added a spec on the model for the validation code, including one that confirms that setting the "find_out_more_active" flag doesn't trigger validation failure for an unrelated field.

PR #3355, opened 2024-01-16
Fix solar charts

ldodds:
* Update analytics to include fix for solar charts: https://github.com/Energy-Sparks/energy-sparks_analytics/pull/661
* Small change to the Annotate service so that it doesn't throw an error if the series are missing. This can happen if the chart being annotated fails to load. Reduces some noise in Rollbar.


urbanwide (approved):
👍

PR #3358, opened 2024-01-17
use i18n-tasks on welsh files too

tbhi:


tbhi:
replaced with https://github.com/Energy-Sparks/energy-sparks/pull/3434

PR #3359, opened 2024-01-17
Allow long term changes pages to work with limited data

tbhi:
- logic changes per card
- switch OpenStruct to ActiveSupport::OrderedOption for rubocop performance warning (probably not much of a concern here but seems good general practice)
- reuse advice base controller's analysis_dates in base long term controller
- rename analysis_dates.one_years_data to one_years_data?
- combine gas/electricity long term partials to reduce duplication
- refactor spec to use factories rather than mocking


app/views/schools/advice/electricity_long_term/_comparison.html.erb
```diff
@@ -1,15 +1,26 @@
...
+    <% end %>
+  </div>
+<% else %>
+  <p>
+    <%= t('advice_pages.electricity_long_term.insights.comparison.less_than_a_year',
+          available_date: analysis_dates.date_when_one_years_data.to_s(:es_short)) %>
+  </p>
+  <p>
+    <%= t('advice_pages.electricity_long_term.insights.comparison.less_than_a_year_well_managed',
+          usage: format_unit(benchmarked_usage.benchmark_value, :kwh)) %>
```
tbhi, line 21: benchmark_value or exemplar_value?

ldodds: Should be `benchmark_value` as these "benchmark schools" are what are referred to as "well managed" on the website.

spec/system/schools/advice_pages/electricity_long_term_spec.rb
ldodds, line 1: Refactoring to use factory bot makes sense, and we could do that for some of the other pages too. Will give a better end-to-end check for some of the processing. 

Just noting that for other pages, and particularly for those gas pages that use heating models, the synthetic data will end needing to be more realistic and cover a longer period. It will also require temperature and calendar data for the required periods. There's a trade-off against testing pages vs the underlying services I think. We'll need the same in the analytics.

ldodds (approved):
Looks good, tested locally with school with limited data

PR #3360, opened 2024-01-19
Join programme prompts

urbanwide:
This is in 3 parts:
1. Adds a message to complete activity page if programme is completed by completing an activity:
Screenshot 2024-01-19 at 13 28 51

2. Adds "join programme prompt" to task (activity and action) completed page & recommendations pages
image

3. To programme_types index, adds this prompt if no programmes are user has no active / started programmes. This was not needed on the recommendations or task completed pages as the "prompt to complete programmes" prompt already has a similar message when there are no active programmes.

Screenshot 2024-01-19 at 13 28 33

4. Just remembered I also refactored some image partials for activities / actions etc.

I have not tested the last 2 points as yet pending review @ldodds


config/locales/views/schools/schools.yml
```diff
@@ -326,6 +326,17 @@ en:
       audit:
         message_html: You have completed <strong>%{completed_activities_count}/%{total_activities_count}</strong> of the activities and <strong>%{completed_actions_count}/%{total_actions_count}</strong> of the actions from your recent energy audit
         summary_html: Complete the others to score <strong>%{remaining_points}</strong> points and <strong>%{bonus_points}</strong> bonus points for completing all audit tasks
+      completed_programme:
+        button: View programme
```
ldodds, line 330: Maybe just "View"? Could then reuse the translation we're using on the programme index page, so the buttons are consistently labelled.

ldodds (approved):

config/locales/views/schools/schools.yml
urbanwide: Yes no probs! Makes sense

config/locales/views/schools/schools.yml
```diff
@@ -326,6 +326,17 @@ en:
...
+        message_html: Well done, you've just completed the <strong>%{title}</strong> programme and have earned <strong>%{bonus_points}</strong> bonus points!
```
ldodds, line 331: One more comment: what if they've completed the programme but are not eligible for the bonus as it wasn't done in same academic year. Should we have different text to avoid saying "earned 0 bonus points"

ldodds (approved):

PR #3368, opened 2024-01-24
Refine recommendations algorithms

urbanwide:
* Now makes activity recommendations based on a school's key stages
* Now swaps out the the old suggesters in the targets pages to use the new algorithms. There are a few places left that use the old suggesters, which are mostly not used as they are behind feature flags that are permanently on (e.g. activitities_2023). 
* Updated panel switcher component not to render panels that are empty
* Now has spec


app/models/alert_type_rating.rb
```diff
@@ -38,6 +38,7 @@ class AlertTypeRating < ApplicationRecord
   has_many :intervention_types, through: :alert_type_rating_intervention_types
 
   scope :for_rating, ->(rating) { where('rating_from <= ? AND rating_to >= ?', rating, rating) }
+  scope :with_scope, ->(scope) { where("#{scope}_active": true) } # need to use this?!
```
ldodds, line 41: I don't think you need to do this. The active flags indicate whether the rating will trigger something being visible on say the pupil or school dashboard or weekly emails. 

I don't think those contexts are relevant to recommendation, as if they've been assigned a rating and that rating has associated activities we should just suggest they do them. We use the same set of activities/interventions regardless of which scopes are active.

So just because we've decided to show something as a management priority and/or on the dashboards doesn't change what we recommend they do.

app/services/recommendations/base.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    def suggested_for(task, excluding: [])
+      task.suggested_types.not_including(excluding)
+    end
+
+    def suggest_random(count, excluding: [])
+      count > 0 ? all(excluding: excluding).sample(count) : []
+    end
+
+    def suggest_from_audits
```
ldodds, line 91: should there be a base class definition of `suggest_from_alerts` too?

ldodds (commented):
Left couple of comment but otherwise this looks OK to me so far. You're discovering the list of activity_types/intervention_types correctly.

app/models/alert_type_rating.rb
urbanwide: Ahh that's great to know, thanks Leigh

app/services/recommendations/base.rb
urbanwide: Yes there should be - well spotted, thank you

ldodds (approved):

PR #3369, opened 2024-01-24
Make out of hours usage page work with less than a year of data

ldodds:
The functional change for this PR is to allow the electricity and gas out of hours usage pages to work with limited data. 

The new `Usage::UsageBreakdownService` service in the analytics will allow a minimum of 7 days of data.

The application side changes involve:

- Revising some text on the insights page so we say which period is being reported on instead of just "last year"
- Omit the school comparison section which requires a year of data to do a proper comparison
- Tweaks to the analysis page to use alternative charts in the holiday section when we have less than a year of data
- Removing rows from the holiday usage comparison table where we don't have data for the current or previous holiday

As there was a lot of duplication across the electricity and gas templates I've:

- Extracted out common partials to cover the common sections, these are the new files in `views/schools/advice/out_of_hours`
- Extracted out some other partials for each section in the analysis pages, to make them easier to amend
- Created a new `HolidayUsageTableComponent` for the holiday comparison table. This has some custom formatting and logic that was worth encapsulating in a component. This makes it easier to test. As a result I've been able to remove other partials that were common to the electricity and gas out of hours pages, and the storage heater analysis
- Run erb-lint over them to tidy them up

A lot of the deleted/new/changed files in the PR are due to the above.

I've added `i18n-tasks` hints to a number of the shared partials otherwise it thought the keys weren't being used. 

I've also reworked the system specs to not use mocks. This means we're now doing end-to-end test against the analytics. This makes the tests more robust, but slightly slower to run.


app/views/schools/advice/electricity_out_of_hours/_by_day.html.erb
```diff
@@ -0,0 +1,24 @@
+<%= render 'schools/advice/section_title',
+           section_id: 'usage_by_day_of_week',
+           section_title:
+           t('advice_pages.electricity_out_of_hours.analysis.usage_by_day_of_week.title') %>
+<%= component 'chart', chart_type: :electricity_by_day_of_week_tolerant, school: school do |c| %>
+  <% c.with_title do %>
+    <%# i18n-tasks-use t('advice_pages.electricity_out_of_hours.analysis.usage_by_day_of_week.electricity_by_day_of_week_tolerant_chart.title') %>
+    <%= t('advice_pages.electricity_out_of_hours.'\
+          'analysis.usage_by_day_of_week.electricity_by_day_of_week_tolerant_chart.title') %>
```
ldodds, line 9: Splitting lines to keep ERB happy, but then having to add hints without split strings to keep i18n-tasks happy. Even if we factored out a prefix, we'd still need the hints. Maybe we should just allow longer lines? Or longer lines in ERB files?

spec/components/holiday_usage_table_component_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+    context 'when comparison cant be made as only partial data for holiday' do
+      let(:data_end_date) { final_holiday_end - 2.days }
+
+      it 'returns nil' do
+        expect(component.comparison_row(holiday_4)).to be false
+      end
+    end
+  end
+
+  context 'when rendering' do
```
tbhi, line 198: wonder if it would be useful check the value of something like `html.css('tr').map { |row| row.css('th, td').map { |td| td.text.strip }` to check the whole table

tbhi (approved):

PR #3371, opened 2024-01-24
try erb-lint

tbhi:
- add erb_lint and config
- run via overcommit on changed files but not on CI as too many errors
- try on app/views/schools/advice/*long_term

### DIFF
--- a/.erb-lint.yml
+++ b/.erb-lint.yml
@@ -1,0 +1,8 @@
+---
+EnableDefaultLinters: true
+linters:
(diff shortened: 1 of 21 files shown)


urbanwide (approved):
Looks good to me @tbhi 👍

ldodds (approved):
Looks good to me. The templates seem to be all whitespace related, will be interesting to see if it flags up other errors.

Assume we should just try running manually locally until we can re-enable in Github Actions.

PR #3377, opened 2024-01-29
show peak usage data on electricity intraday page with 60 days of data

tbhi:


app/services/schools/advice/peak_usage_service.rb
```diff
@@ -8,19 +8,16 @@ def initialize(school, meter_collection)
         @meter_collection = meter_collection
       end
 
-      #Not yet implemented in underlying services
       def enough_data?
-        meter_data_checker.one_years_data?
+        meter_data_checker.at_least_x_days_data?(AlertElectricityPeakKWVersusBenchmark::DAYS_REQUIRED)
```
tbhi, line 12: the card mentions a link to AlertElectricityPeakKWVersusBenchmark so seemed to make sense to reference that - not sure of the actual link though so maybe a comment about that makes sense

ldodds: Using same constant makes sense to me, you could maybe add a comment.

ldodds (approved):

app/services/schools/advice/peak_usage_service.rb
ldodds: See comments here though: https://github.com/Energy-Sparks/energy-sparks_analytics/pull/668#pullrequestreview-1850934811

ldodds:
I think this was failing because of other changes I'd made to analytics. I've merged down the application side now, so if you merge in master it'll probably build now.

PR #3381, opened 2024-01-30
rails 7.0

tbhi:
https://github.com/ViewComponent/view_component/issues/1064


app/services/schools/advice/peak_usage_service.rb
tbhi, line 1: not sure this file should be included

tbhi:
replaced with https://github.com/Energy-Sparks/energy-sparks/pull/3676

PR #3383, opened 2024-02-01
Remove activities_2024 feature flag

urbanwide:
This removes the activities_2024 feature flag.

You'll notice that this removes some of the scoreboard_prizes feature flag content, as that was on the old activities / actions completed page behind the activities_2024 flag. However scoreboard_prizes still feature on the main scoreboard index and individual scoreboard view. Is this ok @ldodds?


ldodds:
I think its OK to remove that. Lets tidy up and re-add if necessary.

PR #3385, opened 2024-02-02
Should solar nav have this switch?

urbanwide:


app/views/schools/advice/_nav.html.erb
```diff
@@ -1,23 +1,39 @@
...
+                    visible: @school.has_storage_heaters?) do |s| %>
     <% sort_by_label(advice_pages.where(fuel_type: :storage_heater)).each do |ap| %>
       <% s.with_item(name: translated_label(ap), href: advice_page_path(@school, ap), classes: 'storage-item') %>
     <% end %>
   <% end %>
-  <% c.with_section(name: t('advice_pages.nav.sections.solar_pv'), icon: 'sun', classes: 'solar-section', visible: @school.has_electricity?) do |s| %>
+  <% c.with_section(name: t('advice_pages.nav.sections.solar_pv'),
+                    icon: 'sun',
+                    classes: 'solar-section',
+                    visible: @school.has_solar_pv?) do |s| %>
```
urbanwide, line 36: The actual change here is changing this:
`visible: @school.has_electricity?`
to
`visible: @school.has_solar_pv?`

ldodds: No, because there are actually two version of the solar page: If a school has solar we report on how much they've generated. If they have electricity but no solar, then we estimate impacts of adding solar.

urbanwide: Ahh didn't realise (and that's a good feature), thanks @ldodds 👍

PR #3387, opened 2024-02-02
Add school URN search

tbhi:


app/views/admin/find_school_by_urn/index.html.erb
```diff
@@ -0,0 +1,33 @@
+<h1>Find schools by URN</h1>
+
+<p>
+Use this form to find the school based on a URN. You can include a partial URN to do a wildcard search. A maximum of 20 meters will be turned
```
ldodds, line 4: maximum of 20 schools?

app/views/admin/find_school_by_mpxn/index.html.erb
```diff
@@ -4,9 +4,9 @@
 Use this form to find the school based on an MPAN or MPRN. You can include a partial MPXN to do a wildcard search. A maximum of 20 meters will be turned
```
ldodds, line 4: I know this was older text, but the form is finding meters rather than schools.

ldodds (approved):

PR #3395, opened 2024-02-05
Create new data model for Metrics and Metric Types

urbanwide:
Adds MetricType, Metric, Period and Report models, scoped to Comparison


app/models/concerns/enum_fuel_typeable.rb
```diff
@@ -0,0 +1,16 @@
+module EnumFuelTypeable
```
ldodds, line 1: worth adding to advice page as it has the same enums in same order?

urbanwide: Will do 👍

app/models/comparison/metric.rb
```diff
@@ -0,0 +1,49 @@
...
+#  fk_rails_...  (custom_previous_period_id => comparison_periods.id) ON DELETE => cascade
+#  fk_rails_...  (metric_type_id => comparison_metrics.id) ON DELETE => cascade
+#  fk_rails_...  (school_id => schools.id) ON DELETE => cascade
+#
+class Comparison::Metric < ApplicationRecord
+  self.table_name = 'comparison_metrics'
+
+  # Add reporting periods as required?
+  # [:last_12_months, :financial_year, :academic_year]
+  enum reporting_period: { custom: 0 }
```
ldodds, line 40: We could add `last_12_months` now, as it'll be needed to start populating some actual values. The `custom` periods will be an additional new feature. The others will get added as we do them. At the moment its looking like `school_year` is the one most used by schools. 

If the period is `custom` then we should have a `custom_period` defined. So could add a validation for that.

app/models/comparison/metric.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  belongs_to :school
+  belongs_to :metric_type, class_name: 'Comparison::MetricType'
+  belongs_to :alert_type
+  belongs_to :custom_period, class_name: 'Comparison::Period', optional: true
+  belongs_to :custom_previous_period, class_name: 'Comparison::Period', optional: true
```
ldodds, line 46: Slightly different design than in the doc. I think it makes more sense to just refer to a custom period which then defines its own current and previous periods, but happy to discuss.

app/models/comparison/metric_type.rb
```diff
@@ -0,0 +1,29 @@
...
+class Comparison::MetricType < ApplicationRecord
+  self.table_name = 'comparison_metric_types'
+
+  include EnumFuelTypeable
+  extend Mobility
+
+  translates :label, type: :string, fallbacks: { cy: :en }
+  translates :description, type: :string, fallbacks: { cy: :en }
+
+  validates :key, presence: true, uniqueness: true
```
ldodds, line 21: Needs to be scoped to the fuel type? E.g. we'll have say `annual_kwh` for gas, electricity, etc.

app/models/comparison/period.rb
```diff
@@ -0,0 +1,16 @@
...
+# Table name: comparison_periods
+#
+#  created_at :datetime         not null
+#  end_date   :date             not null
+#  id         :bigint(8)        not null, primary key
+#  label      :string           not null
+#  start_date :date             not null
+#  updated_at :datetime         not null
+#
+class Comparison::Period < ApplicationRecord
```
ldodds, line 12: I'd originally thought this would define 2 date ranges for current and previous period along with labels for both.

app/models/comparison/report.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  translates :title, type: :string, fallbacks: { cy: :en }
+  translates :introduction, backend: :action_text
+  translates :notes, backend: :action_text
+
+  # Add reporting periods as required?
+  # [:last_12_months, :financial_year, :academic_year]
+  enum reporting_period: { custom: 0 }
+
+  belongs_to :custom_period, class_name: 'Comparison::Period', optional: true
```
ldodds, line 37: See above comments

db/migrate/20240206150348_create_comparison_metric_models.rb
```diff
@@ -0,0 +1,80 @@
...
+      t.references :metric_type, null: false, foreign_key: { to_table: 'comparison_metrics', on_delete: :cascade }
+      t.references :alert_type, null: false
+
+      t.string :value # type can vary, how do we do this?????
+
+      t.integer :reporting_period # enum
+      # thought splitting the periods in to two might provide more flexibility. I might be wrong but is
+      # it true that we're not always comparing to a previous period? and that a metric can just be relevant to
+      # just a single period?
+      # is there the option to store exact date information for every metric rather than just for custom ones?
```
ldodds, line 22: this is largely covered by the additional columns in the metric, as you can derive the dates from the asof_date for the fixed period.

db/migrate/20240206150348_create_comparison_metric_models.rb
```diff
@@ -0,0 +1,80 @@
...
+      t.boolean :public, default: false
+      t.integer :reporting_period # enum
+      # for custom reports
+      t.references :custom_period, null: false, foreign_key: { to_table: 'comparison_periods', on_delete: :cascade }
+      t.references :custom_previous_period, foreign_key: { to_table: 'comparison_periods', on_delete: :cascade }
+      t.timestamps
+    end
+
+=begin
+    # How do we tie metrics to reports?
```
ldodds, line 63: I'd assumed this would be in the code, rather than configured. A report may end up with multiple tables, which a range of metrics. So not explicit association in the database.

app/models/comparison/metric_type.rb
```diff
@@ -0,0 +1,30 @@
...
+#  index_comparison_metric_types_on_key_and_fuel_type  (key,fuel_type) UNIQUE
+#
+class Comparison::MetricType < ApplicationRecord
+  self.table_name = 'comparison_metric_types'
+
+  extend Mobility
+  include EnumFuelType
+
+  enum units: [:float, :date, :percent, :relative_percent]
+  # are there more? [:£, :co2, :kwh, :time, :string, :kw]
```
ldodds, line 23: Yes, I'll send you the list of types that are currently declared, but don't necessarily have to include here

ldodds (approved):

app/models/comparison/metric_type.rb
urbanwide: Thank you @ldodds 👍

PR #3397, opened 2024-02-06
Use fixed dates in specs

ldodds:
Resolves some date issues in the recently updated out of hours specs. Using relative dates (e.g. `30.days.ago`) introduced some date dependencies. Have switched the failing tests over to use fixed dates.

I think the others are OK as they are less reliant on having specific meter data and holiday dates aligning correctly.


urbanwide (approved):
Looks good @ldodds 👍

PR #3400, opened 2024-02-06
Temporarily require 2 years data for monthly comparison chart

ldodds:
Temporarily change the long term usage page so that a broken chart only displays when there is 2 years data.

This avoids a bug in the chart whilst working on a proper fix in the analytics.

### DIFF
--- a/spec/system/schools/advice_pages/electricity_long_term_spec.rb
+++ b/spec/system/schools/advice_pages/electricity_long_term_spec.rb
@@ -111,7 +111,8 @@ RSpec.describe 'electricity long term advice page', :aggregate_failures do
 
         it 'includes expected charts' do
           expect(page).to have_css('#chart_management_dashboard_group_by_week_electricity')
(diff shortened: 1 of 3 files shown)


spec/system/schools/advice_pages/electricity_long_term_spec.rb
```diff
@@ -135,11 +136,36 @@
...
-          expect(page).to have_css('#chart_wrapper_electricity_by_month_year_0_1')
+          # TEMPORARY
+          expect(page).not_to have_css('#chart_wrapper_electricity_by_month_year_0_1')
           # not enough data for this
           expect(page).to have_no_css('#chart_wrapper_electricity_longterm_trend')
         end
       end
+
+      context 'with more than 800 days of meter data' do
+        let(:reading_start_date) { 800.days.ago }
```
tbhi, line 147: would put this on the boundary usually

tbhi (approved):

PR #3402, opened 2024-02-09
Support content management for footnotes

urbanwide:


app/controllers/admin/comparisons/footnotes_controller.rb
```diff
@@ -0,0 +1,38 @@
+module Admin::Comparisons
+  class FootnotesController < AdminController
+    include LocaleHelper
+
+    load_and_authorize_resource :footnote, class: 'Comparison::Footnote'
+
+    def index
```
ldodds, line 7: Suggest having an explicit ordering here to keep index tidy. Sort by the key?

ldodds (approved):

app/controllers/admin/comparisons/footnotes_controller.rb
urbanwide: Good idea, thanks

PR #3403, opened 2024-02-09
Fix some additional text and chart issues with the advice pages

ldodds:
Fixes several issues with the advice pages:

- Revise section heading on out of hours page so its says "Recent trend" instead of "Last 12 months" if there's less than a year of data. Have reused translation key to keep consistent with long term trend page
- Restores the chart title and subtitle to the first chart on the long term usage page. We switch charts if there's <1 year but all charts should have a title and subtitle. As this chart is used on another page I've reused the translation key again
- Hide the heating use by day chart on the gas out of hours page if there's less than a month of data. Otherwise the chart fails to run. The analytics is not very clear on minimum data required for a fit, but 30 days seems to be a key threshold

### DIFF
--- a/app/views/schools/advice/gas_out_of_hours/_by_day.html.erb
+++ b/app/views/schools/advice/gas_out_of_hours/_by_day.html.erb
@@ -29,29 +29,32 @@
   <% end %>
 <% end %>
 
-<%= render 'schools/advice/section_title',
-           section_id: 'usage_through_the_school_day',
-           section_title: t('advice_pages.gas_out_of_hours.analysis.usage_through_the_school_day.title') %>
+<% if analysis_dates.months_of_data >= 1 %>
 
-<%= t('advice_pages.gas_out_of_hours.analysis.usage_through_the_school_day.subtitle') %>
+  <%= render 'schools/advice/section_title',
+             section_id: 'usage_through_the_school_day',
+             section_title: t('advice_pages.gas_out_of_hours.analysis.usage_through_the_school_day.title') %>
 
-<%= component 'chart', chart_type: :gas_heating_season_intraday_up_to_1_year, school: school do |c| %>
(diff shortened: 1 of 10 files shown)


app/views/schools/advice/gas_out_of_hours/_by_day.html.erb
```diff
@@ -29,29 +29,32 @@
...
-     end %>
-  <%# i18n-tasks-use t('advice_pages.gas_out_of_hours.analysis.usage_through_the_school_day.gas_heating_season_intraday_up_to_1_year.footer_html') %>
-  <% c.with_footer do %>
-    <%= t('advice_pages.gas_out_of_hours.'\
-          'analysis.usage_through_the_school_day.gas_heating_season_intraday_up_to_1_year.footer_html',
-          insights_school_advice_heating_control_path: insights_school_advice_heating_control_path(school)) %>
+  <%= t('advice_pages.gas_out_of_hours.analysis.usage_through_the_school_day.subtitle') %>
+
+  <%= component 'chart', chart_type: :gas_heating_season_intraday_up_to_1_year, school: school do |c| %>
+    <%# i18n-tasks-use t('advice_pages.gas_out_of_hours.analysis.usage_through_the_school_day.gas_heating_season_intraday_up_to_1_year.title') %>
```
urbanwide, line 41: I've not seen this in use before but it looks like it's for declaring the use of dynamic keys? However the key doesn't look dynamic? Just curious really.

ldodds: The key isn't dynamic but i18n-tasks doesn't like the line wrapped string just below:

```
t('advice_pages.gas_out_of_hours.'\
          'analysis.usage_through_the_school_day.gas_heating_season_intraday_up_to_1_year.title')
```

I had to break the strings to keep erb lint happy, but this made i18n-tasks unhappy, so had to add explicit hints...

urbanwide (approved):

PR #3420, opened 2024-02-13
Write migration to populate `Comparison::MetricType` with existing benchmark variable definitions

ldodds:
Adds a service that reads the existing benchmark metrics defined in the analytics alert classes and then creates new `MetricType` records for the new school comparison tool.

The mapping:

- handles the variety of different units supported by the analytics, mapping them into a new simpler range.
- ignores some existing metrics that are just copies of attributes from other models, e.g. the school name, that can be easily added to templates
- identifies the fuel type from the metric based on the alert type configuration or, failing that, the name or configuration of the metric. Some values are for multiple fuel types, e.g. annual total energy consumption. A new `:multiple` fuel type is added to support that, but this required some tweaks (and erblint tidying) to the advice pages

In addition to the spec, the code has been run on all configured alerts. Currently outputs 458 MetricType records.

Some further work is required to finalise some of the unit types, and possibly the metric definitions used in the mapping (e.g. if we want to rename some metrics) but this can come in further PRs. 

Current expectation is that this may be run as required during development, but prior to release of the newer code we'll run this via an afterparty task to populate the database.

### DIFF
--- a/app/services/comparison/metric_migration_service.rb
+++ b/app/services/comparison/metric_migration_service.rb
@@ -1,0 +1,79 @@
+module Comparison
+  # Populate the MetricType model based on the definitions in the analytics
+  class MetricMigrationService
+    def migrate
+      AlertType.enabled.each do |alert_type|
+        create_metrics(alert_type) if alert_type.benchmark_variables.present?
+      end
+      true
+    rescue => e
+      puts e
+      puts e.backtrace
+      false
+    end
+
+    # Maps the units defined by the alert classes to what we will be storing
+    # in the database
+    def units_for_metric_type(unit)
+      # fallback if not declared by analytics
+      return :float if unit.nil?
+      return :boolean if unit == TrueClass
+      # String => :string, Integer => :integer
+      return unit.to_s.downcase.to_sym if unit.is_a?(Class)
+      # { kw: :electricity}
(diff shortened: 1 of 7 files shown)


spec/services/comparison/metric_migration_service_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+      end
+
+      it 'does not create anything' do
+        expect(Comparison::MetricType.count).to be(0)
+      end
+    end
+
+    context 'with an test alert' do
+      # rubocop:disable RSpec/LeakyConstantDeclaration
+      # rubocop:disable Lint/ConstantDefinitionInBlock
```
tbhi, line 79: think rubocop:disable RSpec/LeakyConstantDeclaration, Lint/ConstantDefinitionInBlock works

app/services/comparison/metric_migration_service.rb
```diff
@@ -0,0 +1,79 @@
...
+      case key.to_s
+      when /electricity/
+        return :electricity
+      when /gas/
+        return :gas
+      when /storage_heater/
+        return :storage_heater
+      when /solar/
+        return :solar_pv
+      end
```
tbhi, line 53: else clause then can drop the returns?

tbhi (approved):

PR #3421, opened 2024-02-13
Adapt benchmark generation to populate new metrics tables

ldodds:
Initial version of a service to populate the new `Comparison::Metric` model when a school is regenerated.

- fixes an incorrect foreign key in the `Comparison::Metric` table, as it had a circular reference rather than a reference to `Comparison::MetricType` this was stopping data from being inserted
- adds a reference from `Comparison::Metric` to `BenchmarkResultSchoolGenerationRun` so we can track metrics generated for a run
- adds additional reporting periods to cover all the alerts
- adds a custom ActiveRecord Type to allow serialising and deserialising different metric types to the database. This avoids need for e.g. template code to keep casting the values
- revises how Metric Types are generated for ratings, so that they are prefixed by their class name
- adds `Comparison::MetricCreationService` which does the work of creating the records using the results of running an alert
- integrates the above with the `GenerateAndSaveAlertsAndBenchmarks` service (behind a feature flag) to allow metrics to be generated alongside the existing `Alert` and `BenchmarkResult` records.
- removes some old services that ran alerts and benchmarks that are no longer used
- adds an admin only view listing the generated metrics for a school in the benchmark results report. Just shows a simple table of all metrics for a run. This probably needs tests, but its internal only and pretty basic. 
  
This is sufficient to allow us to start working on the rest of the UX for creating comparison reports.

The feature flag to activate this is `FEATURED_FLAG_NEW_SCHOOL_COMPARISONS`. You will need to have run: `Comparison::MetricMigrationService.migrate` to populate the `Comparison::MetricType` table. Without that the Comparison::MetricCreationService won't create any data.

The Comparison::MetricCreationService service does not yet handle custom reporting periods. This is planned for a later iteration.


ldodds:
One thing I'm not sure about here is the handling of `rating`. I've ended up scoping them to the class as I was concerned about ambiguity. But realise now that you can disambiguate via the AlertType in the metric. So considering reverting that. Any thoughts?

I had to change the variables names from using £ signs as a suffix as this was causing encoding errors in the Rails console. To avoid any issues I've just rewritten them.

tbhi (approved):

ldodds:
Replaced by: https://github.com/Energy-Sparks/energy-sparks/pull/3423

PR #3422, opened 2024-02-15
Implement basic framework for producing a school comparison report

ldodds:
Based on https://github.com/Energy-Sparks/energy-sparks/pull/3421 which needs reviewing and merging first. Actual changes in this PR are smaller.

The key changes in this PR are:

* Adds scopes to `Metric` and `BenchmarkResultSchoolGenerationRun` to support querying for latest metrics for schools. 
* Adds some simple types `Comparison::ReportDefinition` and `Comparison::ReportResult` for the description of a report (as a list of metric types, etc) and results (as a hash of schools to metric values)
* A `Comparison::ReportService` that accepts a `ReportDefinition` and returns a `ReportResult`
* A initial base controller for running reports
* The implementation of the `baseload_per_pupil` report, using a definition which matches the existing comparison page. Has both a chart and a table

The PR currently uses the existing I18n translation keys for the report, although some of this will be moving into the database.

Expecting that specific implementation to be refactored as we extract/add components for tables and charts.

Some work is required to:

- adds some more tests for edge cases, e.g. missing metrics required for ordering
- finalising table layout, e.g. how do footnotes
- integration with the `Report` and `Footnote` model

...but I think all of that can probably be done in other PRs, as this is big enough already.

It'd be good to get some feedback on this so we can then move on to doing any necessary components for the tables/charts and start on porting over the rest of the reports.

It turned out to be quite easy to actually build the report page with all this in place.

I wasn't sure about the final URL structure, its currently using `/comparisons` but the existing code is at `/compare`. Should we move it?

### DIFF
--- a/app/classes/comparison/report_definition.rb
+++ b/app/classes/comparison/report_definition.rb
@@ -1,8 +1,8 @@
 module Comparison
   class ReportDefinition
-    attr_reader :report_key, :advice_page, :schools, :metric_type_keys, :alert_types, :fuel_types, :order_key
+    attr_reader :report_key, :advice_page, :schools, :metric_type_keys, :alert_types, :fuel_types, :order_key, :order
 
-    def initialize(report_key:, advice_page:, schools:, metric_type_keys:, alert_types:, fuel_types:, order_key:)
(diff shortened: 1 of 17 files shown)


ldodds:
Replaced by: https://github.com/Energy-Sparks/energy-sparks/pull/3423

PR #3423, opened 2024-02-18
Alternate comparison framework

ldodds:
This PR implements an alternative approach for storing and querying metrics for the new school comparison pages.

Rather than use a new `Comparison::Metric` table this approach does the following:

- stores the raw variable data in the Alert table, keeping all the generated metrics in a single location
- uses the [Scenic](https://github.com/scenic-views/scenic) gem to create a database view which has the necessary data for the report
- uses basic model-view-controller setup to implement the report, making the query code much simpler

The view relies on using the Postgres [json_t_record](https://www.postgresql.org/docs/9.5/functions-json.html) function to extract specific keys from the variables stored in the `Alert` table. The function has to cast these to a type known by the database. This means that we can rely on ActiveRecord to handle the type conversions. 

The Postgres [floating point types](https://www.postgresql.org/docs/14/datatype-numeric.html#DATATYPE-FLOAT) support handling of Nan/Infinitity/-Infinity values so that removes the need to add extra round-tripping over those values.

Overall I think this approach is actually better than what I had originally designed and implemented, so am proposing we use this approach instead although still need to work through a few things.

Advantages:

- there's less code involved
- much of the work of migrating the existing reports will be writing the SQL for the views which will follow a fairly predictable pattern.
- for those reports that have more complex sorting or selection criteria these will be easier to implement as we'll just be using standard SELECT/ORDER by clauses in the queries over the views
- we can more easily calculate, e.g. sum/min/max/average values in the queries over the views, doing more work in the database
- we can increase performance by using Scenic support for materialised views (ensuring they are refreshed after each daily benchmark run) and then adding indexes

Disadvantages:

- Requires database migrations every time we change the report? But this seems very minor
- Involves some custom Postgres functions, so we're more closely tied to that as a database. But we have dependencies on it elsewhere
- No others...?
- 
Still to be worked through:

- ensuring that the stored JSON can be correctly cast to the native postgres type. Postgres already handles casting of iso 8601 formatted data, and I've added support for Float values. But need to ensure that booleans (which are stored at `t`/`f` are mapped to [a supported text value](https://www.postgresql.org/docs/14/datatype-boolean.html)

This PR still needs tidying up of some of the code and additional of test cases.

### DIFF
--- a/app/views/comparisons/baseload_per_pupil/index.html.erb
+++ b/app/views/comparisons/baseload_per_pupil/index.html.erb
@@ -1,0 +1,95 @@
+<% content_for :page_title, 'Baseload Per Pupil' %>
+
+<h1><%= t('analytics.benchmarking.chart_table_config.baseload_per_pupil') %></h1>
+
+<%= render 'compare/summary' %>
+
+<%= t('analytics.benchmarking.content.baseload_per_pupil.introduction_text_html') %>
+
+<%= t('analytics.benchmarking.caveat_text.es_exclude_storage_heaters_and_solar_pv').html_safe %>
+
+<ul class="nav nav-tabs locales url-aware" id="compare-results-tabs" role="tablist">
+    <li class="nav-item">
+      <a class="nav-link active" id="tables-tab" data-toggle="tab" href="#tables"
+        role="tab" aria-controls="tables-content"><%= t('compare.show.tables') %></a>
+    </li>
+    <li class="nav-item">
+      <a class="nav-link" id="charts-tab" data-toggle="tab" href="#charts"
+        role="tab" aria-controls="charts-content"><%= t('compare.show.charts') %></a>
+    </li>
+</ul>
+
+<div class="tab-content" id="compare-results-content">
+  <div class="tab-pane fade show active" id="tables" role="tabpanel" aria-labelledby="tables-tab">
+    <div class="row">
+      <div class="col-sm-12">
+        <table id="comparison-table" class="table advice-table table-sorted">
+          <thead>
+            <tr>
+              <th><%= t('analytics.benchmarking.configuration.column_headings.school') %></th>
+              <th class="text-right">
+                <%= t('analytics.benchmarking.configuration.column_headings.baseload_per_pupil_w') %>
+              </th>
+              <th class="text-right">
+                <%= t('analytics.benchmarking.configuration.column_headings.last_year_cost_of_baseload') %>
+              </th>
+              <th class="text-right">
+                <%= t('analytics.benchmarking.configuration.column_headings.average_baseload_kw') %>
+              </th>
+              <th class="text-right">
+                <%= t('analytics.benchmarking.configuration.column_headings.baseload_percent') %>
+              </th>
+              <th class="text-right">
+                <%= t('analytics.benchmarking.configuration.column_headings.saving_if_matched_exemplar_school') %>
+              </th>
+            </tr>
+          </thead>
+          <tbody>
+            <% @results.each do |baseload_per_pupil| %>
+              <tr>
+                <td>
+                  <%= link_to baseload_per_pupil.school.name,
+                              advice_page_path(baseload_per_pupil.school, @advice_page, :insights) %>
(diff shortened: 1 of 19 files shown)


db/migrate/20240218141409_add_alert_columns.rb
```diff
@@ -0,0 +1,6 @@
+class AddAlertColumns < ActiveRecord::Migration[6.1]
+  def change
+    add_column :alerts, :variables, :json
```
tbhi, line 3: would jsonb be better?

app/views/comparisons/baseload_per_pupil/index.html.erb
```diff
@@ -0,0 +1,95 @@
...
+              </th>
+            </tr>
+          </thead>
+          <tbody>
+            <% @results.each do |baseload_per_pupil| %>
+              <tr>
+                <td>
+                  <%= link_to baseload_per_pupil.school.name,
+                              advice_page_path(baseload_per_pupil.school, @advice_page, :insights) %>
+                  <% if baseload_per_pupil.electricity_economic_tariff_changed_this_year == true %>
```
tbhi, line 53: `== true` not required?

db/views/baseload_per_pupils_v01.sql
```diff
@@ -0,0 +1,29 @@
...
+      one_year_baseload_per_pupil_kw float,
+      annual_baseload_percent float,
+      one_year_saving_versus_exemplar_gbp float
+    )
+    WHERE alert_type_id=1
+  ) AS baseload,
+  (
+    SELECT alert_generation_run_id, school_id, data.*
+    FROM alerts, json_to_record(variables) AS data(electricity_economic_tariff_changed_this_year boolean)
+    WHERE alert_type_id=59
```
tbhi, line 20: would be nice to clarify what the fix ids are

db/migrate/20240218141409_add_alert_columns.rb
ldodds: I'm not sure. I think jsonb is slower to write, which could impact regeneration times. Otherwise with this approach it'll be down to whether postgres implementation of json_to_record vs jsonb_to_record has much of a performance difference

db/views/baseload_per_pupils_v01.sql
ldodds: Yep, this was a quick prototype. The view should be joining to alert type here by `class_name`.

urbanwide (approved):

ldodds:
I've updated this PR to tidy the code and add some specs around the first report.

I have:

- changed the `variables` column in the alert table to be jsonb
- adding the reporting period column to alerts, although this is not yet used
- added a new base class for the view models
- refactored the baseload_per_pupil report to push more common elements into the base class and some shared templates
- added some shared examples which might be useful for testing other reports, we can probably integrate some of the existing compare_spec.rb examples too, but have not done that here
- added a system spec for the report that just checks it renders with expected data, it could be more comprehensive but good enough for a start

I've removed the Welsh version of the variables. I think for the school comparisons there shouldn't be any text columns that are created when the variables are generated. If there are we should change those.

Things we discussed but I've not done:

- I've left the `BaseloadPerPupil` model in its own file. Using a single shared file breaks the `annotate --models` command. We might be able to filter that, but its quite useful to see the annotations in the file? Maybe there's a work around I missed?
- I've not generalised the controller. There's some elements which are bespoke to a report, e.g. the query it runs, how it builds its chart as well as some text and linking options. The latter might be generalisable through the `Comparison::Report` model but the rest needs to be defined somewhere. I wasn't sure that moving this to a service class rather than just leaving in the controller added much. Happy to change, but maybe we merge this first?
- Removed the models and code from the previous version, I'll do that in a separate PR.

Before we merge this I'll need to get this analytics PR merged and the Gemfile updated to a tagged version: https://github.com/Energy-Sparks/energy-sparks_analytics/pull/673

urbanwide (approved):
Looks good - think it works keeping the view models in separate files in order to have the annotations and the front end stuff looks good too 👍

PR #3429, opened 2024-02-19
Revise when out gas use through the school day chart is displayed

ldodds:
There is a chart on the out of hours usage page that requires a heating model to work correctly.

At the moment the page is just checking for >1 month of data as the analytics by default requires at least 30 days. But it still might not generate a valid heating model from the available data, there needs to be sufficient data to calculate usage for heating on/off. These don't quite align with "summer" and "winter" date periods, so it's difficult to check without actually calculating the regression models.

To handle this I've I've revised the gas out of hours page so that it applies the same check to see if there's enough data as the heating control page. If there isn't then the chart isn't shown.

### DIFF
--- a/app/controllers/schools/advice/gas_out_of_hours_controller.rb
+++ b/app/controllers/schools/advice/gas_out_of_hours_controller.rb
@@ -2,6 +2,7 @@ module Schools
   module Advice
     class GasOutOfHoursController < BaseOutOfHoursController
       before_action :load_dashboard_alerts, only: [:insights, :analysis]
+      before_action :set_heating_model_available, only: [:analysis]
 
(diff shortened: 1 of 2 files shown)


urbanwide (commented):
Wonder if a test could be added for this?

ldodds:
There is a test around this already, the spec checks to see if the chart is inserted or not based on the available data. Previously it was relying on the date specific check, but the tests also pass. The analytics is generating a default heating model with the synthetic data produced from the test.

I could maybe make that a bit more explicit but will need to figure out how to generate some representative gas data across a year.

urbanwide (approved):

PR #3431, opened 2024-02-20
test flake - try ordering search

tbhi:


tbhi:
ordering change done as part of https://github.com/Energy-Sparks/energy-sparks/pull/3676

PR #3433, opened 2024-02-21
Fix argument name

ldodds:
Found another key that was renamed but not in the Welsh translations.


urbanwide (approved):

tbhi:
is this another one https://github.com/Energy-Sparks/energy-sparks/blob/c6c9011a19f78f8d9eb490b7045e653fd75c9428/config/locales/cy/views/advice_pages/gas_costs.yml#L8 ?

ldodds:
The additional ones are typos/mistakes introduced by the translators. The definitive source of text for the cy files is the transifex system, rather than our repo. I've fixed those there.

PR #3434, opened 2024-02-21
update i18n-tasks config, make it work on cy files

tbhi:


ldodds:
It would be helpful to be able to identify missing/incorrect interpolations in the Welsh files within the existing structure. But moving the files around will cause issues with synching requiring every key and file to be reviewed. Missing translations to be expected while we are waiting for translation work to be done. We shouldn't be creating our own Welsh text via google translate, etc.

PR #3437, opened 2024-02-21
Annual change in electricity since last year report

ldodds:
Implements the `change_in_electricity_since_last_year` (Annual change in electricity) report.

# Changes to common code:

- adjusted base templates to make tabs option is no charts are provided by the controller, as this comparison doesn't have a chart currently
- added a `sticky-heading` style that can be applied to `thead` this [makes the table header sticky](https://btxx.org/posts/Please_Make_Your_Table_Headings_Sticky/) so if a user scrolls with a long table the heading stays visible. it doesn't work so well for tables with multiple header rows but have left in place here, will be fine with others.

# Things to note for other comparisons:

- I've pushed the sorting into the database by calculating the percentage change in the order by clause
- I've center aligned the column groups so they display better
- replace the values in the "Estimated" column with yes/no values that are translated rather than the hard-coded Y/N we use currently, not sure if other tables have hard-coded text, but be aware of those
- I've copied some helper functions from `Benchmarking::BenchmarkManager` into `ComparisonsHelper`. I've reworked them slightly but otherwise left unchanged. Suggest we copy over what we need then tidy and add tests.

# Change to report behaviour

The current where condition for the report does this:

```
!enba_ken.nil? && enba_peap != ManagementSummaryTable::NO_RECENT_DATA_MESSAGE
```

This equates to schools that have a previous year electricity value but the text "no recent data" for the value of their `activationyear_electricity_gbp_relative_percent` attribute. This didn't seem right to me as I think the report is trying to remove schools that don't have recent data, not no recent data for their first year on the site. We also don't display that column. 

So I'll be changing the condition to confirm we have current and previous year instead.

### DIFF
--- a/db/views/change_in_electricity_since_last_years_v01.sql
+++ b/db/views/change_in_electricity_since_last_years_v01.sql
@@ -1,0 +1,30 @@
+SELECT latest_runs.id,
+       enba.school_id,
+       enba.previous_year_electricity_kwh,
+       enba.current_year_electricity_kwh,
+       enba.previous_year_electricity_co2,
+       enba.current_year_electricity_co2,
+       enba.previous_year_electricity_gbp,
+       enba.current_year_electricity_gbp,
+       enba.solar_type
+FROM
+  (
+    SELECT alert_generation_run_id, school_id, data.*
+    FROM alerts, alert_types, jsonb_to_record(variables) AS data(
+      previous_year_electricity_kwh float,
+      current_year_electricity_kwh float,
+      previous_year_electricity_co2 float,
+      current_year_electricity_co2 float,
+      previous_year_electricity_gbp float,
+      current_year_electricity_gbp float,
+      solar_type text
+    )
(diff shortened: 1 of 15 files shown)


db/views/change_in_electricity_since_last_years_v01.sql
```diff
@@ -0,0 +1,30 @@
...
+      previous_year_electricity_kwh float,
+      current_year_electricity_kwh float,
+      previous_year_electricity_co2 float,
+      current_year_electricity_co2 float,
+      previous_year_electricity_gbp float,
+      current_year_electricity_gbp float,
+      solar_type text
+    )
+    WHERE alerts.alert_type_id = alert_types.id and alert_types.class_name='AlertEnergyAnnualVersusBenchmark'
+  ) AS enba,
```
tbhi, line 23: should we avoid these old names?

tbhi (approved):

db/views/change_in_electricity_since_last_years_v01.sql
ldodds: I was leaving them in to help backtrack during development. I'm not wedded to keeping them.

PR #3438, opened 2024-02-22
Peak school day electricity use

tbhi:


spec/system/comparisons/electricity_peak_kw_per_pupil_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+describe 'electricity_peak_kw_per_pupil' do
+  let(:school) { create(:school) }
+
+  before do
+    create(:advice_page, key: :electricity_intraday)
+
+    alert_run = create(:alert_generation_run, school: school)
+
+    baseload_alert = create(:alert_type, class_name: 'AlertElectricityPeakKWVersusBenchmark')
```
tbhi, line 13: concerned that if the class name is wrong here and the code then the test will pass when it shouldn't

ldodds: As a small improvement, what about adding a factory that accepts a Class rather than a name and returns an alert_type. That at least confirms that the class exists in the code base? 

Similarly for the alert we could add a factory or helper that confirms that an alert class exposes the variables we want to store the in the `Alert`. 

Provides a bit of consistency checking without us having to run every analysis?

Gemfile
```diff
@@ -26,8 +26,8 @@ gem 'pg'
 gem 'scenic'
 
 # Dashboard analytics
-gem 'energy-sparks_analytics', github: 'Energy-Sparks/energy-sparks_analytics', tag: '5.1.0'
-# gem 'energy-sparks_analytics', github: 'Energy-Sparks/energy-sparks_analytics', branch: 'alert-interface-change'
+# gem 'energy-sparks_analytics', github: 'Energy-Sparks/energy-sparks_analytics', tag: '5.1.1'
+gem 'energy-sparks_analytics', github: 'Energy-Sparks/energy-sparks_analytics', branch: '4110-peak-school-day-electricity-use'
```
ldodds, line 30: Maybe pull in the rename in a separate PR?

app/controllers/comparisons/base_controller.rb
```diff
@@ -43,6 +45,31 @@ def create_charts(_results)
...
+      #
+      # This could be done by clipping values to 100.0 if the metric has a
+      # unit of percentage/relative_percent
+      results.each do |result|
+        metric = result.send(name)
+        next if metric.nil? || metric.nan? || metric.infinite?
+
+        # for a percentage metric we'd multiply * 100.0
+        # here we're converting from kW to W
+        chart_data[result.school.name] = metric * 1000.0
```
ldodds, line 62: This will need revising as its only baseload and maybe peak usage where we convert to watts. As noted in the comments in other places we convert percentages.

Gemfile
tbhi: I think `spec/i18n_spec.rb` fails without this

config/locales/analytics/benchmarking/content_base.yml
```diff
@@ -301,7 +301,7 @@ en:
           unadjusted_kwh: Unadjusted change (kWh)
           urn: URN
           variation_in_baseload_between_days_of_week: Variation in baseload between days of week
-          w_floor_area: w/floor area
+          w_floor_area: Watt/floor area
```
ldodds, line 304: Not sure if you've used that here, but there's a rake task to "Reinstall" the YML from the analytics. So suggest you merge you analytics branch, bump the tag and refresh the YML here.

ldodds (approved):

app/controllers/comparisons/electricity_peak_kw_per_pupil_controller.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Comparisons
+  class ElectricityPeakKwPerPupilController < BaseController
+    private
+
+    def title_key
```
ldodds, line 7: With the changes Deb has made to integrate the report, you'll need to define the report_key method, then you can remove the _introduction.html.erb file as the content will be loaded from the database.

Spec will also need the report created.

ldodds (approved):

PR #3448, opened 2024-02-26
Electricity use during current holiday

tbhi:


db/schema.rb
tbhi, line 1: not sure why this has changed so much

app/controllers/comparisons/electricity_consumption_during_holiday_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Comparisons
+  class ElectricityConsumptionDuringHolidayController < BaseController
+    private
+
+    def title_key
```
ldodds, line 7: report key

app/views/comparisons/electricity_consumption_during_holiday/_introduction.html.erb
ldodds, line 1: Can remove this now

ldodds (changes requested):

ldodds (approved):

PR #3449, opened 2024-02-26
Fix link on national scoreboard

urbanwide:
* Also fixes scoreboard_prizes feature flag spec failing locally

### DIFF
--- a/app/views/scoreboards/_table.html.erb
+++ b/app/views/scoreboards/_table.html.erb
@@ -20,8 +20,7 @@
             <h2 class="nowrap">
               <%= link_to school.sum_points,
                           school_timeline_path(
-                            school, (
-                              { academic_year: @academic_year } unless @scoreboard.national?)
+                            school, ({ academic_year: @academic_year } unless @scoreboard.national?)
                           ),
                           class: 'badge badge-success' %>
               &nbsp;
--- a/app/views/scoreboards/_table.html.erb
+++ b/app/views/scoreboards/_table.html.erb
@@ -19,7 +19,10 @@
           <td>
             <h2 class="nowrap">
               <%= link_to school.sum_points,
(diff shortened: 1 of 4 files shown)


app/views/scoreboards/_table.html.erb
```diff
@@ -18,7 +18,12 @@
           </td>
           <td>
             <h2 class="nowrap">
-              <%= link_to school.sum_points, school_timeline_path(school, academic_year: @academic_year), class: 'badge badge-success' %>
+              <%= link_to school.sum_points,
+                          school_timeline_path(
+                            school, (
+                              { academic_year: @academic_year } unless @scoreboard.national?)
```
tbhi, line 24: wonder if end bracket should be on next line? or everything on the same line?

spec/system/scoreboard_spec.rb
```diff
@@ -40,6 +40,7 @@
       visit scoreboard_path(scoreboard)
       expect(page).to have_content('Super scoreboard')
       expect(page).to have_content(school_with_points.name)
+      expect(page).to have_link(points.to_s, href: school_timeline_path(school_with_points, academic_year: scoreboard.this_academic_year))
```
tbhi, line 43: to_s not needed?

tbhi (approved):

app/views/scoreboards/_table.html.erb
urbanwide: I will have another go but this is what I ended up with after a small fight with erblint :-)

spec/system/scoreboard_spec.rb
urbanwide: Without it, I get the following warning: 

`Integer:123 for selector :link must be an instance of String or Symbol.`

PR #3451, opened 2024-02-28
Integrate benchmark reports to use new comparison report model

urbanwide:


app/views/compare/_summary.html.erb
```diff
@@ -1,16 +1,46 @@
 <div class="p-3 mb-3 bg-light rounded border clearfix">
-  <%= link_to t('compare.filter.change_benchmark'), benchmarks_compare_index_path(index_params.except(:benchmark)), class: 'btn btn-sm float-right ml-2' if action_name == 'show' %>
+  <%= if action_name != 'benchmarks'
```
urbanwide, line 2: The only thing that has changed here technically is as follows. All other stuff is erblint related! From
`action_name == 'show'
`
to:
`action_name != 'benchmarks'
`

ldodds (approved):
Looks good.

Having a link to the old report would be helpful for development and testing if there's an easy way to incorporate that.

urbanwide:
Have added another feature flag - COMPARISON_REPORTS_LINK_TO_OLD, which adds a link in brackets after the new benchmark.

PR #3452, opened 2024-02-28
Change in electricity use between this holiday and the same holiday last year

tbhi:


app/views/comparisons/change_in_electricity_holiday_consumption_previous_years_holiday/_introduction.html.erb
ldodds, line 1: as before can remove this and add report key

app/views/comparisons/change_in_electricity_holiday_consumption_previous_years_holiday/_tables.html.erb
```diff
@@ -0,0 +1,100 @@
...
+          <%= up_downify(format_unit(result.difference_percent, :relative_percent_0dp, true, :benchmark)) %>
+        </td>
+        <td class="text-right">
+          <%= up_downify(format_unit(result.difference_gbpcurrent, :£_0dp, true, :benchmark)) %>
+        </td>
+        <td class="text-right">
+          <%= up_downify(format_unit(result.difference_kwh, :kw, true, :benchmark)) %>
+        </td>
+        <td class="text-right">
+          <%= result.name_of_current_period %>
```
ldodds, line 58: We're going to need to make some changes to the alert class here for `name_of_current_period` and `name_of_previous_period`. These values will be text and we are only storing values that have been generated with locale of English. So Welsh uses will see English period names.

The `period_name` code looks something like:

```
  protected def period_name(period)
    I18nHelper.holiday(period.type)
  end
```

So we'll need to change the alert to declare a variable and add a method that returns the `period.type`. We can then we can use that to display the correct text in the table.

We only need to do this for any text to be displayed. Any constants can be used as is.

ldodds (changes requested):

app/views/comparisons/change_in_electricity_holiday_consumption_previous_years_holiday/_tables.html.erb
ldodds: I've not looked at the values recently. If they can be easily mapped back to existing keys that might be simpler, but if we can fix at source that's probably for the best.

app/views/comparisons/change_in_electricity_holiday_consumption_previous_years_holiday/_tables.html.erb
```diff
@@ -0,0 +1,100 @@
...
+    <% tariff_has_changed_footnote = false %>
+    <% infinite_increase_footnote = false %>
+    <% infinite_decrease_footnote = false %>
+    <% results.each do |result| %>
+      <tr>
+        <td>
+          <%= result.school.name %>
+          <% if result.pupils_changed %>
+            <% pupils_changed_footnote = true %>
+            <a href="#footnotes">(*1)</a>
```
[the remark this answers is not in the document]
tbhi: makes sense to leave until footnotes are refactored?

app/views/comparisons/change_in_electricity_holiday_consumption_previous_years_holiday/_table.html.erb
```diff
@@ -0,0 +1,80 @@
...
+      <% end %>
+      <% if result.tariff_has_changed %>
+        <% tariff_has_changed_footnote = true %>
+        <% row.with_reference do %>
+          <a href="#footnotes">(*6)</a>
+        <% end %>
+      <% end %>
+      <%= row.with_var val: result.difference_percent, unit: :relative_percent_0dp, change: true %>
+      <%= row.with_var val: result.difference_gbpcurrent, unit: :£_0dp, change: true %>
+      <%= row.with_var val: result.difference_kwh, unit: :kw %>
```
ldodds, line 37: `:kwh` I think?

ldodds (approved):

PR #3454, opened 2024-02-29
Change when consent is withdrawn via n3rgy API

ldodds:
Currently when we deactivate a meter we issue a "withdraw consent" request to the n3rgy API.

It turns out that n3rgy actually handle consent slightly differently to what we expected. Issuing consent for a given MPxN actually results in consent being granted for all meters at that address, not the specific meter used in the API request. So, for example, using the electricity meter MPAN will also grant consent to access the gas data. There's not a more granular way to grant/withdraw consent that for the whole address (or more accurately the HAN to which the MPxN is connected).

This means our current approach of removing consent if the electricity meter is disabled may end up removing access to the gas data, which isn't what was intended.

This PR changes the meter management service so that:

- if a meter is deleted, we withdraw consent
- if we remove all data for a meter, we withdraw consent
- if we deactivate a meter, then consent is not withdrawn
- if we activate a meter, and we can grant consent then we do so. This ensures that any outstanding requests are still made

The PR also changes the rake tasks that load readings and tariffs to only process active meters. Previously because consent was removed when a meter was made inactive it wouldn't have been picked up.

### DIFF
--- a/app/services/meter_management.rb
+++ b/app/services/meter_management.rb
@@ -64,6 +64,9 @@ class MeterManagement
   def delete_meter!
     @meter.transaction do
       AggregateSchoolService.new(@meter.school).invalidate_cache
+      if @meter.can_withdraw_consent?
+        Meters::DccWithdrawTrustedConsents.new([@meter]).perform
+      end
       @meter.destroy
     end
   end
--- a/app/services/meter_management.rb
+++ b/app/services/meter_management.rb
@@ -94,9 +94,7 @@ class MeterManagement
 
   def remove_data!(archive: false)
     result = true
-    if @meter.can_withdraw_consent?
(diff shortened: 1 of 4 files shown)


app/services/meter_management.rb
```diff
@@ -84,16 +84,16 @@ def deactivate_meter!
...
-        result = Meters::DccWithdrawTrustedConsents.new([@meter]).perform
-      end
     end
     broadcast(:meter_deactivated, @meter)
     result
   end
 
   def remove_data!(archive: false)
     result = true
+    if @meter.can_withdraw_consent?
```
tbhi, line 94: one liner syntax?

tbhi (approved):

tbhi (approved):

PR #3458, opened 2024-03-01
Improve error handling in metering service

ldodds:
Better handle issue where API doesn't return any cached dates.


tbhi (approved):
a test would be good?

PR #3459, opened 2024-03-01
Add Rollbar warning for multiple devices in HAN

ldodds:
The new n3rgy API is designed to support a wider range of metering scenarios than v1. Specifically it allows for a scenario where a HAN (Home Area Network) has multiple devices for the same type. E.g. multiple electricity meters.

In a domestic setting this is very unlikely to happen and, so far, it doesn't seem to be the case with schools. All currently schools only have one meter of each type (gas, electricity).

This PR adds a Rollbar warning if, after retrieving data, we find readings for multiple meters. It still continues to return readings but warns us so we can investigate.

This seems an unlikely scenario. It's more likely that any additional meters will have their own MPANs and, presumably, networks. But this should reduce the likelihood of us incorrectly loading data if it does happen.


tbhi (approved):
expect before the call still looks odd to me

PR #3461, opened 2024-03-01
Change in electricity use between the last two holidays

tbhi:


app/views/comparisons/change_in_electricity_holiday_consumption_previous_holiday/_tables.html.erb
```diff
@@ -0,0 +1 @@
+<%= render 'comparisons/change_in_electricity_holiday_consumption_previous_years_holiday/tables', results: results %>
```
ldodds, line 1: If there are tables with same layout, can we move the shared partials to a common directory. E.g. `comparisons/tables/change_in_electricity_holiday_consumption`? We're likely to have the exact same report for gas and storage heater too, so worth extracting a single common table if we can.

ldodds (changes requested):

app/views/comparisons/shared/change_in_electricity_holiday_consumption/_table.html.erb
```diff
@@ -0,0 +1,80 @@
...
+      <% end %>
+      <% if result.tariff_has_changed %>
+        <% tariff_has_changed_footnote = true %>
+        <% row.with_reference do %>
+          <a href="#footnotes">(*6)</a>
+        <% end %>
+      <% end %>
+      <%= row.with_var val: result.difference_percent, unit: :relative_percent_0dp, change: true %>
+      <%= row.with_var val: result.difference_gbpcurrent, unit: :£_0dp, change: true %>
+      <%= row.with_var val: result.difference_kwh, unit: :kw %>
```
ldodds, line 37: `:kwh`

ldodds (approved):

PR #3468, opened 2024-03-04
Fix benchmark links

urbanwide:
Rails route helpers append _index only for singular named routes to the collection resource. This does not apply for plural named routes which was assumed when linking to benchmarks. This changes the link to work for both kinds.

### DIFF
--- a/app/helpers/comparisons_helper.rb
+++ b/app/helpers/comparisons_helper.rb
@@ -13,4 +13,12 @@ module ComparisonsHelper
     val = data.sum
     to_nil_if_sum_zero && val == 0.0 ? nil : val
   end
+
+  def comparison_page_exists?(key)
+    begin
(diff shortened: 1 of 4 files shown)


app/helpers/comparisons_helper.rb
```diff
@@ -13,4 +13,12 @@ def sum_data(data, to_nil_if_sum_zero = false)
     val = data.sum
     to_nil_if_sum_zero && val == 0.0 ? nil : val
   end
+
+  def comparison_page_exists?(key)
+    begin
+      "Comparisons::#{key.to_s.camelcase}Controller".constantize
```
tbhi, line 19: would `Object.const_defined?` work?

urbanwide: Yes, looks like it does, thank you :-)

PR #3469, opened 2024-03-05
Annual electricity use per pupil with savings potential

urbanwide:


config/routes.rb
```diff
@@ -60,6 +60,7 @@
   end
 
   namespace :comparisons do
+    resources :annual_electricity_costs_per_pupil, only: [:index]
```
tbhi, line 63: yeah we should probably sort these, I've just been adding to the bottom :grimacing:

tbhi (approved):
looks good to me, don't know if @ldodds also wants to review?

ldodds (approved):

config/routes.rb
urbanwide: The route generator seems to put these at the top of the namespace by default.. (now it is working - see https://github.com/Energy-Sparks/energy-sparks/pull/3472)

app/models/comparison/annual_electricity_costs_per_pupil.rb
```diff
@@ -0,0 +1,15 @@
...
+#
+#  alert_generation_run_id                    :bigint(8)
+#  id                                         :bigint(8)
+#  last_year_gbp                              :float
+#  one_year_electricity_per_pupil_gbp         :float
+#  one_year_saving_versus_exemplar_gbpcurrent :float
+#  school_id                                  :bigint(8)
+#
+class Comparison::AnnualElectricityCostsPerPupil < Comparison::View
+  scope :with_data, -> { where.not(one_year_electricity_per_pupil_gbp: nil) }
```
urbanwide, line 13: The original doesn't have this condition but it seems like a good idea to make sure there is some data before displaying the school?

ldodds: Yes, seems sensible

ldodds (approved):

PR #3470, opened 2024-03-05
Add CSV download support to school comparison pages

ldodds:
This PR makes some changes to the core comparison framework to enable CSV downloads of tables. 

Some reports include multiple tables and we want to be able to support separate downloads for each of them. So I've revised the framework to better support multiple tables and rendering them as either HTML or CSV.

Summary of the changes:

- a report controller now defines one or more table names. These map to partials that should be in the view directory. By default every report defines a single table called `table` which is now defined in `_table.html.erb`
- the base controller automatically invokes the HTML partial for every table defined by its `table_name` method, passing in a common set of parameters (this include a table number which I'm going to use in a later PR)
- the base controller supports CSV download of a named table, if a request is made with `format: :csv` and a `table_name` parameter, then it will invoke the CSV partial for that table. E.g. `_table.csv.ruby` to match the default table name
- the `download_link` helper accepts a report, table name and existing filter parameters and builds a link to the controller that will trigger the download of the right table, with the same filters as the user currently has specified
- the shared examples for the school comparison pages have been refactored to split out specs for the basic page, the presence of  a chart, and expected contents of a named table and expected contents of the CSV file linked from the table. This is to allow us to test contents of multiple charts and add some specs around filters (I've not done the latter yet)

So, when this merged, if a report has a single table, then just need to fill in content for `_table.html.erb` and `_table.csv.ruby`. The generator has been updated to produce those files and will include the right download link. So for simple reports everything should just work without any changes.

If a report has multiple tables, then you can override `table_names` method in the controller, to do, e.g:

```
def table_names
 [:table, :additional_table]
end
```

And then add `_table.html.erb`, `_table.csv.ruby`, `_additional_table.html.erb` and `_additional_table.csv.ruby`. 

We may need to tweak this to support shared partials for tables, but I think that's easy to add.

I've also been through an updated all of the current comparisons to revise the specs and implement the CSV downloads.


tbhi:
Does it benefit from the table csv being in erb?
Looking around a see it done as a render - https://api.rubyonrails.org/classes/ActionController/Renderers.html
or a handler https://www.lugolabs.com/articles/handling-csv-views-in-ruby-on-rails or just plain ruby in the controller

tbhi:
Also wondering about the overlap between the html and csv tables - are they always the same data?  Would having @headers shared between the 2 make senses?

ldodds:
> Does it benefit from the table csv being in erb?

Only to the extent that I thought it would be easier to maintain the two versions like this as they're very similar in terms of contents.

I'll have a look at renderers/handlers. Elsewhere in the code base we've used a variety of methods, including code in the controller or separate services.

> Also wondering about the overlap between the html and csv tables - are they always the same data? Would having @headers shared between the 2 make senses?

They will always be the same data, so headers will be consistent and could be shared, yes.

ldodds:
The [handler option](https://www.lugolabs.com/articles/handling-csv-views-in-ruby-on-rails) looks like it would work within the framework for handling downloads and keeps the two versions of the files together. So I could try that. We'd need to agree on the custom suffix for the file names how about `.data`. So: `table.html.erb` and `table.csv.data`. Any thoughts?

tbhi:
Another idea, wondering about a small class for each column with header, csv, and html methods as they seem all related.

tbhi:
> The [handler option](https://www.lugolabs.com/articles/handling-csv-views-in-ruby-on-rails) looks like it would work within the framework for handling downloads and keeps the two versions of the files together. So I could try that. We'd need to agree on the custom suffix for the file names how about `.data`. So: `table.html.erb` and `table.csv.data`. Any thoughts?

think the file is basically ruby - so perhaps csvrb?

ldodds:
Using a custom format caused some encoding problems with translation keys. It also wasn't adding much over using `.ruby` files which Rails already supports. So I've switched to that. The different is just these extra lines in the file:

```
CSV.generate do |csv|
  ...
end.html_safe
```

spec/system/comparisons/electricity_targets_spec.rb
```diff
@@ -51,6 +54,27 @@
          ["Notes\nIn school comparisons 'last year' is defined as this year to date."]
         ]
       end
+      let(:expected_csv) do
+        [['School',
```
tbhi, line 58: headers could be a shared variable for most of these?

app/helpers/comparisons_helper.rb
```diff
@@ -1,4 +1,28 @@
 module ComparisonsHelper
+  def comparison_table_id(report, table_name)
+    "#{report.key}-#{table_name}"
+  end
+
+  def download_link(report, table_name, params)
+    # If the key is plural then rails routing works slightly
+    # differently, so exclude :index component
+    if report.key.singularize == report.key
```
tbhi, line 9: perhaps
```    
path = [:comparisons, report.key.to_sym]
path << :index if report.key.singularize == report.key
polymorphic_path(path, params: params.merge(table_name: table_name, format: :csv))
```

tbhi (approved):

PR #3473, opened 2024-03-06
Exclude generator from autoloader

ldodds:


urbanwide (approved):
Thanks @ldodds, looks good 👍

PR #3474, opened 2024-03-06
Solar generation summary

ldodds:


spec/system/comparisons/solar_generation_summary_spec.rb
tbhi, line 1: couple of bracket format issues

tbhi (approved):

PR #3478, opened 2024-03-06
Benefits of solar pv installation

ldodds:


db/schema.rb
```diff
@@ -2295,4 +2295,59 @@
             ORDER BY alert_generation_runs.school_id, alert_generation_runs.created_at DESC) latest_runs
     WHERE (data.alert_generation_run_id = latest_runs.id);
   SQL
+  create_view "solar_generation_summaries", sql_definition: <<-SQL
```
urbanwide, line 2298: I think this could be from another PR?

urbanwide (approved):
Looks good - just one comment about schema.rb

db/schema.rb
ldodds: Yep, forgot to rollback when jumping between PRs

PR #3479, opened 2024-03-07
Initial version of comparison table component

ldodds:
This is a draft to illustrate an alternate approach, to see if we prefer the syntax/setup. We don't have to use this, but I thought a concrete example would be helpful.

Adds a view component for handling formatting of comparison tables.

- headers are defined in the controller and passed to the component and reused across both the HTML and CSV views
- for tables with multi-line headers, controllers can implement `col_group` to provide an array of column group configuration. Used in the component only currently, but could be used in CSV too
- supports a `school` slot for providing the school name and footnote text in the first column
- supports a `footer` slot for content to be added to the table footer
- supports row and column formatting, with cell contents specified using a `var` slot. This is flexible enough to format content, optionally wrap it in `updownify` or just pass the provided content through without further processing

I've reworked 2 views one simpler one and one with multi-line headers, alternate cell formatting and plain text columns.

The pros:

- views are a bit easier to read?
- easier to manage common table elements and styling as now centralised in the component
- helps to enforce common formatting conventions, ensure cells and headers are aligned correctly
- seems relatively easy to extend, although the `VarComponent` will potentially end up doing a lot of things

The cons:

- headers are now defined in separate class so need to cross-check between files to ensure things are ordered correctly
- ...?

### DIFF
--- a/app/components/comparison_table_component.rb
+++ b/app/components/comparison_table_component.rb
@@ -11,15 +11,22 @@ class ComparisonTableComponent < ViewComponent::Base
   include AdvicePageHelper
   include ComparisonsHelper
 
-  def initialize(report:, table_name:, index_params:, headers: [], colgroups: [])
+  def initialize(report:, advice_page:, table_name:, index_params:, headers: [], colgroups: [], advice_page_tab: :insights)
     @report = report
-    @colgroups = colgroups
-    @headers = headers
+    @advice_page = advice_page
     @table_name = table_name
     @index_params = index_params
+    @headers = headers
+    @colgroups = colgroups
+    @advice_page_tab = advice_page_tab
   end
 
-  renders_many :rows, 'RowComponent'
+  renders_many :rows, ->(**args) do
+    args[:advice_page] = @advice_page
+    args[:advice_page_tab] = @advice_page_tab
+    RowComponent.new(**args)
+  end
+
   renders_one :footer
 
   # For providing information for each row in the comparison table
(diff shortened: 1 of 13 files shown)


app/components/comparison_table_component.rb
```diff
@@ -0,0 +1,97 @@
...
+# A row may refer to footnotes
+#
+# A footnote section that provides the detail for individual footnotes
+class ComparisonTableComponent < ViewComponent::Base
+  include AdvicePageHelper
+  include ComparisonsHelper
+
+  def initialize(report:, table_name:, index_params:, headers: [], colgroups: [])
+    @report = report
+    @colgroups = colgroups
```
urbanwide, line 16: Very minor but can these be in the same order they are passed as parameters?

app/views/comparisons/change_in_electricity_since_last_year/_table.html.erb
```diff
@@ -1,106 +1,32 @@
...
-        </td>
-        <td class="text-right">
-          <% if change_in_electricity_since_last_year.solar_type == 'synthetic' %>
-            <%= t('common.labels.yes_label') %>
-          <% else %>
-            <%= t('common.labels.no_label') %>
-          <% end %>
-        </td>
-      </tr>
+      <%= r.with_var { result.solar_type == 'synthetic' ? t('common.labels.yes_label') : t('common.labels.no_label') } %>
```
urbanwide, line 26: I think we ought to use the y_n(boolean) helper here - although I appreciate it was already like this :-)

app/views/comparisons/baseload_per_pupil/_table.html.erb
```diff
@@ -1,58 +1,22 @@
-<div class='text-right mt-2'>
-  <%= download_link(report, table_name, index_params) %>
-</div>
+<%= component 'comparison_table', headers: @headers, report: report, table_name: table_name, index_params: index_params do |c| %>
+  <% @results.each do |result| %>
+    <% c.with_row do |r| %>
+      <% r.with_school school: result.school, advice_page: advice_page %>
```
urbanwide, line 4: No biggie but would the advice_page key be applicable to the whole table rather than just a school? i.e. should it be passed as a parameter to the comparison_table component, rather than to every school? Might be a bit fiddly to make it available to sub-components though.

urbanwide (approved):
Looks good - just a couple of minor comments 👍

app/views/comparisons/baseload_per_pupil/_table.html.erb
urbanwide: The PageNavComponent does some passing of parameters through to slots if you need an example

app/views/comparisons/change_in_electricity_since_last_year/_table.html.erb
ldodds: Yes, that's better!

app/views/comparisons/baseload_per_pupil/_table.html.erb
ldodds: Yes, that's probably better. I've reworked.

PR #3483, opened 2024-03-07
Annual change in solar PV production and resulting CO2 savings

ldodds:
Implements the annual change in solar production table, but also sorts a couple of other issues:

* There was a bug in the generator template, so the specs had the wrong shared context name, this was masking test failures
* Fixed failing spec for other report 
* Adds a shared scope to `Comparison::View` for doing percentage changes, allowing for NULL values. Should be consistent with the function currently used to sort values

### DIFF
--- a/app/views/comparisons/change_in_solar_pv_since_last_year/_table.html.erb
+++ b/app/views/comparisons/change_in_solar_pv_since_last_year/_table.html.erb
@@ -1,0 +1,33 @@
+<%= component 'comparison_table',
+              report: report, advice_page: advice_page, table_name: table_name, index_params: index_params,
+              headers: headers, colgroups: colgroups, advice_page_tab: advice_page_tab do |c| %>
+  <% @results.each do |result| %>
+    <% c.with_row do |r| %>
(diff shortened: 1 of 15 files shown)


app/views/comparisons/change_in_solar_pv_since_last_year/_table.html.erb
```diff
@@ -0,0 +1,33 @@
+<%= component 'comparison_table',
+              report: report, advice_page: advice_page, table_name: table_name, index_params: index_params,
+              headers: headers, colgroups: colgroups, advice_page_tab: advice_page_tab do |c| %>
+  <% @results.each do |result| %>
+    <% c.with_row do |r| %>
+      <% r.with_school school: result.school %>
+
+      <%= r.with_var val: result.previous_year_solar_pv_kwh, unit: :kwh %>
```
urbanwide, line 8: Not that this really matters but don't think these need to be `<%=`, just `<%` should work.

urbanwide (approved):
👍

PR #3492, opened 2024-03-11
Add support for m3 conversion and cns file upload

ldodds:
Adds a flag to the data feed config model to indicate if values should be converted from m3 to kwh. Then updates the data feed translator to apply it.

Currently will just apply it to all readings if set and there is no reading type defined in the CSV. Or, if there is, then only to those marked as m3/M3.

Also adds a new config that uses this support and adds the `.cns` file type to the manual upload form.

CNS seems to be a pseudo CSV format used by gas suppliers.

### DIFF
--- a/app/services/amr/data_feed_translator.rb
+++ b/app/services/amr/data_feed_translator.rb
@@ -9,6 +9,7 @@ module Amr
 
     def perform
       hash_rows = @array_of_rows.map { |row| translate_row_to_hash(row) }
+      puts hash_rows.inspect
       check_units(hash_rows)
     end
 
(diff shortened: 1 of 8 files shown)


app/services/amr/data_feed_translator.rb
```diff
@@ -9,6 +9,7 @@ def initialize(config, array_of_rows)
 
     def perform
       hash_rows = @array_of_rows.map { |row| translate_row_to_hash(row) }
+      puts hash_rows.inspect
```
urbanwide, line 12: Probably don't need the puts here?

PR #3497, opened 2024-03-12
Recent change in baseload

urbanwide:
Had a wee issue with up_downify sanitizing formatted values containing the `"&percnt;"` html symbol and them becoming double encoded (e.g. `"&amp;percnt;"`)

Analytics is inconsistent in how it presents it's percent symbols for :html formatted values:
```
> FormatEnergyUnit.format(:percent, 0.123, :html)
=> "12&percnt;"
> FormatEnergyUnit.format(:relative_percent, 0.123, :html)
=> "+12%"
> FormatEnergyUnit.format(:relative_percent_0dp, 0.123, :html)
=> "+12%"
> FormatEnergyUnit.format(:percent_0dp, 0.123, :html)
=> "12%"
> FormatEnergyUnit.format(:comparison_percent, 0.123, :html)
=> "+12&percnt;"

```
This looks to be an issue with how FormatEnergyUnit.format does it's translations paired with how our translations are configured. When asking for a :html formatted value, FormatEnergyUnit.format will first look for a translation key ending in _html and if there isn't one, fall back to the normal (non _html suffixed key). Only :percent has a key for percent_html, the others do not (ignore comparison_percent, the :comparison_percent format actually uses the :percent and :percent_html translation keys):

```
      percent: "%{value}%"
      percent_0dp: "%{value}%"
      percent_html: "%{value}&percnt;"
      comparison_percent: "%"
      relative_percent: "%{sign}%{value}%"
      relative_percent_0dp: "%{sign}%{value}%"
```
So to be consistent, we should have _html keys for all types of percent (or change analytics to have less of them).

In the short term, I've added an option to up_downify so we can opt not to sanitize a value when calling up_downify inside the comparison_table component for 'change' values.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -2117,6 +2117,46 @@ ActiveRecord::Schema.define(version: 2024_03_07_181846) do
   add_foreign_key "users", "staff_roles", on_delete: :restrict
   add_foreign_key "weather_observations", "weather_stations", on_delete: :cascade
 
+  create_view "annual_change_in_electricity_out_of_hours_uses", sql_definition: <<-SQL
+      SELECT latest_runs.id,
+      usage.alert_generation_run_id,
+      usage.school_id,
+      usage.out_of_hours_kwh,
+      usage.out_of_hours_co2,
+      usage.out_of_hours_gbpcurrent,
+      usage_previous_year.previous_out_of_hours_kwh,
+      usage_previous_year.previous_out_of_hours_co2,
+      usage_previous_year.previous_out_of_hours_gbpcurrent,
+      additional.electricity_economic_tariff_changed_this_year
+     FROM ( SELECT alerts.alert_generation_run_id,
+              alerts.school_id,
+              data.out_of_hours_kwh,
+              data.out_of_hours_co2,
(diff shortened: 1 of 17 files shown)


lib/generators/comparison_report/templates/_table.csv.ruby.tt
```diff
@@ -1,11 +1,9 @@
 CSV.generate do |csv|
   # headers
-  csv << [
-    t('analytics.benchmarking.configuration.column_headings.school')
-  ]
-  @results.each do |row|
+  csv << @headers
+  @results.each do |result|
```
urbanwide, line 4: Changed from row to result to match the html.erb template

db/schema.rb
urbanwide, line 1: Lots of changes in this file - but this is the schema that comes out of migrating from the latest database dump...

ldodds:
Is it also worth doing an analytics PR to add the `_html` keys?

db/schema.rb
ldodds: Yes I had this too.

ldodds (approved):

PR #3499, opened 2024-03-13
 Migrate Gas and Storage Heater Benchmarks - Annual change in gas/storage heater use 

tbhi:


app/controllers/comparisons/change_in_heater_since_last_year.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Comparisons
+  module ChangeInHeaterSinceLastYear
```
ldodds, line 4: Minor: wondering about changing the name here slightly. As its about either gas use (which is both hot water and heating) or storage heater use. The name seems to relate to the latter. Maybe `ChangeInHeatingSinceLastYear` and similar tweaks to prefix for partials?

db/views/change_in_storage_heaters_since_last_years_v01.sql
```diff
@@ -0,0 +1,35 @@
+SELECT latest_runs.id,
+       energy.*,
+       storage.temperature_adjusted_previous_year_kwh,
+       storage.temperature_adjusted_percent
+
+FROM
+  (
+    SELECT alert_generation_run_id, school_id, data.*
+    FROM alerts, alert_types, jsonb_to_record(variables) AS data(
+      previous_year_gas_kwh float,
```
ldodds, line 10: Should these not be the storage heater variables generated by this alert, rather than the gas variables? AlertEnergyAnnualVersusBenchmark generates separate sets of variables for electricity, gas and storage heaters. See `mallaig-high-school` as an example

ldodds (changes requested):
I think some use of incorrect variables which need changing

app/controllers/comparisons/change_in_heater_since_last_year.rb
tbhi: changed

ldodds (approved):

PR #3500, opened 2024-03-13
Annual energy use per pupil and floor area comparisons

ldodds:
Implements the two comparisons, but also:

- fixes a bug on the school dashboard (went to check the translation key for school type and realised it wasn't being used)
- fixes table heading alignment in the comparison table component
- fixes for specs to change % to `&percnt;` following analytics changes

The per pupil report diverges slightly from analytics in that:

The analytics uses a custom function (`sort_energy_costs`) to order the rows. This ends up with a counter-intuitive sort order depending on whether schools have one or more fuel types. Here's the current version showing schools with electricity + gas, just gas, electricity plus storage heaters. The overall ordering really isn't clear:

Screenshot from 2024-03-13 18-12-11

I've just changed this to sort based on total kwh which is much clearer. If a user wants to sort the table by a fuel type, then they can do that. I've done that by just implemented a scope that allows the results to be ordered by some named columns (`by_total`).

New view:

Screenshot from 2024-03-13 18-14-44

The filtering in the analytics is done by referencing a total kwh value from another alert (`AlertEnergyAnnualVersusBenchmark`). I've simplified that to just adding a scope that allows the results to be filtered to schools that have values in at least one of the specified columns (`.where_any_present`).

There were some pending changes for the floor area report, to align it with the pupil report, so I've implemented that too, using a shared view. As those revisions included some new columns I've added keys for those in the main project. Avoids unnecessary changes to the analytics. I've updated the base controller to allow different base keys to be used for the charts.


Gemfile.lock
```diff
@@ -829,4 +829,4 @@ DEPENDENCIES
   wisper-rspec
 
 BUNDLED WITH
-   2.5.6
+   2.5.4
```
tbhi, line 832: probably don't want to use an older version?

tbhi (approved):

PR #3503, opened 2024-03-14
Migrate Gas and Storage Heater Benchmarks - Annual change in gas/storage heater used out of school hours

tbhi:


app/views/comparisons/annual_change_in_electricity_out_of_hours_use/_table.html.erb
```diff
@@ -1,42 +1,10 @@
...
-  <tr>
-    <td colspan="<%= headers.count %>">
-      <a name="footnotes"><strong><%= t('analytics.benchmarking.content.footnotes.notes') %></strong></a>
-      <%= t('analytics.benchmarking.configuration.the_tariff_has_changed_during_the_last_year_html') %>
-      <%= t('analytics.benchmarking.configuration.column_heading_explanation.last_year_definition_html') %>
-    </td>
-  </tr>
-  <% end %>
-<% end %>
+<%= render 'comparisons/shared/annual_change_in_out_of_hours_use/table',
```
urbanwide, line 1: This could be changed to pass all the variables through to: `locals: local_assigns`

app/models/comparison/annual_change_in_gas_out_of_hours_use.rb
```diff
@@ -0,0 +1,17 @@
+# == Schema Information
+#
+# Table name: annual_change_in_gas_out_of_hours_uses
+#
+#  alert_generation_run_id                       :bigint(8)
+#  electricity_economic_tariff_changed_this_year :boolean
```
urbanwide, line 6: I think the tariff changed column is different for gas?

urbanwide: I think gas_economic_tariff_changed_this_year see from analytics:

```
    def self.tariff_changed_school_name(content_class = nil)
      if content_class.nil?
        { data: ->{ tariff_change_reference(addp_name, addp_etch || addp_gtch)}, name: :name, units: String, chart_data: true }
      else
        { data: ->{ tariff_change_reference(addp_name, addp_etch || addp_gtch)}, name: :name, units: String, chart_data: true, content_class: content_class }
      end
    end
```

app/views/comparisons/shared/annual_change_in_out_of_hours_use/_table.html.erb
```diff
@@ -0,0 +1,42 @@
+<%= component 'comparison_table',
+              report: report, advice_page: advice_page, table_name: table_name, index_params: index_params,
+              headers: headers, colgroups: colgroups, advice_page_tab: advice_page_tab do |c| %>
+  <% results.each do |result| %>
+    <% c.with_row do |r| %>
+      <% r.with_school school: result.school %>
+
+      <% if result.electricity_economic_tariff_changed_this_year %>
```
urbanwide, line 8: Same comment about it possibly being a different column for gas. Not sure how storage heaters works though...

app/views/comparisons/annual_change_in_gas_out_of_hours_use/_table.csv.ruby
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+render 'comparisons/shared/annual_change_in_out_of_hours_use/table'
```
urbanwide, line 3: I have passed locals: local_assigns through here when doing a similar thing, which makes me wonder if I've done it right. I will check

app/models/comparison/annual_change_in_gas_out_of_hours_use.rb
ldodds: Yes, that's right its a different field.

app/views/comparisons/shared/annual_change_in_out_of_hours_use/_table.html.erb
ldodds: Storage heaters use the electricity tariffs, gas has its own tariffs, so you need to use the electricity var for storage heater footnotes and the gas var for gas footnotes.

urbanwide (changes requested):
Looks good - main comment is the field used for the footnote needs changing for gas

app/views/comparisons/annual_change_in_electricity_out_of_hours_use/_table.html.erb
tbhi: changed to inherited controller so template not needed

app/views/comparisons/shared/annual_change_in_out_of_hours_use/_table.html.erb
tbhi: corrected

ldodds (approved):

app/controllers/comparisons/annual_change_in_electricity_out_of_hours_use_controller.rb
```diff
@@ -1,33 +1,9 @@
 # frozen_string_literal: true
 
 module Comparisons
-  class AnnualChangeInElectricityOutOfHoursUseController < BaseController
+  class AnnualChangeInElectricityOutOfHoursUseController < Shared::AnnualChangeInOutOfHoursUse
```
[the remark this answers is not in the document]
ldodds: Controller suffix might be clearer?

PR #3504, opened 2024-03-15
Annual saving through 1C reduction in thermostat temperature

urbanwide:
The original report has up_downify arrows on the column, which results in green down arrows for a negative number. I don't think down is positive from reading the report description but I could be wrong: 

> A negative value is indicative of very poor thermostatic control, where Energy Sparks’ modelling cannot make sense of the relationship between the school's gas consumption and outside temperature.

Have left it as the original for now, but wondered if we need to update up_downify to have the option to switch negative to be red if need be?

image


tbhi (approved):
separate card for the red negative?

PR #3505, opened 2024-03-15
Heating used in warm weather

urbanwide:


db/schema.rb
```diff
@@ -2218,31 +2218,6 @@
             ORDER BY alert_generation_runs.school_id, alert_generation_runs.created_at DESC) latest_runs
     WHERE (enba.alert_generation_run_id = latest_runs.id);
   SQL
-  create_view "electricity_consumption_during_holidays", sql_definition: <<-SQL
```
urbanwide, line 2221: This has moved again - if it's just me I'll move it back :-)

spec/system/comparisons/heating_in_warm_weather_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'rails_helper'
+
+describe 'heating_in_warm_weather' do
+  let!(:school) { create(:school) }
+  let(:key) { :heating_in_warm_weather }
+  let(:advice_page_key) { :heating_control }
+
+  let(:gas_variables) do
```
tbhi, line 8: I'd put these in the before block since they don't need to be lets but personal style I guess

tbhi (approved):

PR #3508, opened 2024-03-15
Hot Water Efficiency

urbanwide:
This is another one where a column (the last one) was using up_downify but I think a green down arrow was possibly wrong - should it be red as down I think is bad in this case...


spec/system/comparisons/hot_water_efficiency_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'rails_helper'
+
+describe 'hot_water_efficiency' do
+  let!(:school) { create(:school) }
+  let(:key) { :hot_water_efficiency }
+  let(:advice_page_key) { :hot_water }
+
+  # change to your variables
```
tbhi, line 8: guess don't want this

tbhi (approved):

PR #3515, opened 2024-03-18
Progress against gas target

urbanwide:
Adds gas_targets but also refactors electricity_targets (in the same way Tristan has done for other reports) - as they both share the same functionality.


app/views/comparisons/electricity_targets/_table.html.erb
```diff
@@ -1,27 +1 @@
...
-    <td colspan="<%= headers.count %>">
-      <p>
-        <strong><%= t('analytics.benchmarking.content.footnotes.notes') %></strong>
-      </p>
-      <%= t('analytics.benchmarking.configuration.column_heading_explanation.last_year_definition_html') %>
-    </td>
-  </tr>
-  <% end %>
-<% end %>
+<%= render partial: 'comparisons/shared/targets/table', locals: local_assigns %>
```
tbhi, line 1: like the local_assigns, would be good to do this on the others

tbhi (approved):

app/views/comparisons/electricity_targets/_table.html.erb
urbanwide: Good point, I'll add them to the PR 👍

ldodds (approved):

PR #3518, opened 2024-03-18
Annual cost of electricity, gas, storage heaters and combined energy

urbanwide:


app/models/comparison/annual_energy_costs.rb
```diff
@@ -0,0 +1,21 @@
...
+#  last_year_gbp                 :float
+#  last_year_kwh                 :float
+#  last_year_storage_heaters     :float
+#  one_year_energy_per_pupil_gbp :float
+#  pupils                        :float
+#  school_id                     :bigint(8)
+#  school_type_name              :text
+#
+class Comparison::AnnualEnergyCosts < Comparison::View
+  scope :sort_default, -> { order('last_year_gbp DESC NULLS LAST') }
```
urbanwide, line 20: Added NULLS LAST otherwise unlike the old report, the schools with no data were being displayed first...

db/views/annual_energy_costs_v01.sql
```diff
@@ -0,0 +1,55 @@
...
+    last_year_gbp float
+  )
+  WHERE alerts.alert_type_id = alert_types.id and alert_types.class_name='AlertGasAnnualVersusBenchmark'
+), storage_heaters AS (
+  SELECT alert_generation_run_id, data.*
+  FROM alerts, alert_types, jsonb_to_record(variables) AS data(
+    last_year_gbp float
+  )
+    WHERE alerts.alert_type_id = alert_types.id and alert_types.class_name='AlertStorageHeaterAnnualVersusBenchmark'
+), total AS (
```
tbhi, line 19: energy vs total?

tbhi (approved):

db/views/annual_energy_costs_v01.sql
urbanwide: Yes - you're right, that would be better 👍

PR #3519, opened 2024-03-19
Cost of energy used in upcoming holiday last year

urbanwide:


db/schema.rb
```diff
@@ -2526,6 +2526,38 @@
             ORDER BY alert_generation_runs.school_id, alert_generation_runs.created_at DESC) latest_runs
     WHERE ((benefit_estimate.alert_generation_run_id = latest_runs.id) AND (additional.alert_generation_run_id = latest_runs.id));
   SQL
+  create_view "recent_change_in_baseloads", sql_definition: <<-SQL
```
urbanwide, line 2529: schema.rb is all messed up in terms of ordering again - have left as is from my local machine - don't know if to try and "fix" it though or just merge this in?

db/views/holiday_usage_last_years_v01.sql
```diff
@@ -0,0 +1,23 @@
...
+  (
+    SELECT alert_generation_run_id, school_id, data.*
+    FROM alerts, alert_types, jsonb_to_record(variables) AS data(
+      last_year_holiday_gas_gbp float,
+      last_year_holiday_electricity_gbp float,
+      last_year_holiday_gas_gbpcurrent float,
+      last_year_holiday_electricity_gbpcurrent float,
+      last_year_holiday_gas_kwh_per_floor_area float,
+      last_year_holiday_electricity_kwh_per_floor_area float,
+      name_of_last_year_holiday text
```
tbhi, line 13: probably need to change the alert here to have holiday type, start and end date like https://github.com/Energy-Sparks/energy-sparks/blob/b7522294e257fdccd83950e13824d1d79e08251f/db/views/electricity_consumption_during_holidays_v02.sql#L9 ?

urbanwide: Ahh ok, didn't realise this was a thing, thanks

urbanwide: Analytics doesn't have holiday_type available for this alert (it looks like it could have with a change though). Where did the requirement come from to add this extra info?

urbanwide: I have just worked out that you had to add it to analytics.

Gemfile.lock
```diff
@@ -829,4 +829,4 @@ DEPENDENCIES
   wisper-rspec
 
 BUNDLED WITH
-   2.5.6
+   2.4.22
```
ldodds, line 832: I think we're all on different versions of bundler? I've just done `bundle update --bundler` and I'm on 2.5.4.

ldodds (commented):
Looks OK apart from needing a new version of analytics?

Gemfile.lock
urbanwide: Have just updated to 2.5.6...

ldodds (approved):

PR #3521, opened 2024-03-20
Annual change in total energy use

ldodds:
Adds the "Annual change in total energy use" (`change_in_energy_since_last_year`) comparison report.

This diverges from the current implementation in a few ways:

* we now exclude any school that doesn't have enough data to do a comparison. The previous version included these in the table, but there's not much value in having them here if we can't show an actual change between periods
* the table columns have been reordered to add the metering columns at the start, with icons and translated text in place of the current fixed strings
* the "No recent data" column has been changed to a "Recent data" column
* if there's no recent data then the column is highlighted in red as we do elsewhere
* the recent data check is done using the school configuration, rather than a variable produced by the alert. There's a new school `.with_school_configuration` which efficiently includes the configuration if we're using it in the views
* I've added a footnote if the school tariff has changed in last 12 months


urbanwide:
Still in the process of reviewing this but noticed the popover for the footnotes isn't working for some reason. Also, the footnote link is red for me in this line, perhaps it should be black?

image

app/views/comparisons/change_in_energy_since_last_year/_table.html.erb
```diff
@@ -0,0 +1,46 @@
...
+        <% end %>
+
+        <%= r.with_var do
+              gas_or_electricity_data_stale?(result) ? t('common.labels.no_label') : t('common.labels.yes_label')
+            end %>
+
+        <% %i[kwh co2 £].each do |unit| %>
+          <%= r.with_var val: result.total_previous_period(unit: unit),
+                         unit: unit %>
+          <%= r.with_var val: result.total_current_period(unit: unit), unit: unit %>
```
urbanwide, line 39: Don't think the extra `unit: unit` should be in this line

urbanwide (approved):
Looks good but not sure why the footnote popovers are not working just for this report. Also see other couple of minor things.

urbanwide:
Also - wonder if these icons should be using .text-nowrap so they stay on one line?
image

ldodds:
> Looks good but not sure why the footnote popovers are not working just for this report. Also see other couple of minor things.

Took me a while to track this down @urbanwide. The footnotes also aren't working on the "Benefits of Solar" benchmark. After stepping through it I finally realised that the footnotes are only working on comparisons with charts. It turns out our JS code only activated Bootstrap popovers on pages that successfully load a chart. Tooltips are enabled on any page, but the code is within the chart js too. I'm going to do a separate Trello card and PR for that as its not specific to this change,

PR #3522, opened 2024-03-21
HTML percent issues

urbanwide:
The format_unit helper requests :html formatted values from analytics - so now all html formatted percent values include `&percnt;`, where as before they were inconsistently `%` or `&percnt;` depending on the type of percentage format requested.

i.e. the format_unit helper calls:
`
    FormatEnergyUnit.format(units, value, :html, false, in_table, user_numeric_comprehension_level).html_safe
`

The main issue was that up_downify by default calls sanitize on values, so now that it is being passed values containing `&percnt;` rather than `%` in some cases, it was outputting: `&amp;percnt;` which is rendered as `&percnt;`.

I've fixed these in the app by not sanitizing in this situation, assuming this is ok to do anyway.

The other issue we were having in a couple of places was that when passing a value containing html percentage symbol as a parameter to the translate helper, that it wasn't expecting html, (i.e. the translation key didn't end in _html or html_safe wasn't called on the output).

There are also a bunch of changes here for erblint...

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -307,7 +307,7 @@ module ApplicationHelper
              ''
            end
     text = sanitize(text) if sanitize
-    text + ' ' + icon.html_safe
+    (text + ' ' + icon).html_safe
   end
 
   def safely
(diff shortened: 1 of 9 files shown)


ldodds (approved):
Thanks for working through this!

PR #3531, opened 2024-03-25
Integrate footnote model into tables

urbanwide:
Completed but not had chance to give it a last check over. Would have liked to have added more tests for the references in the comparison table component but ran out of time.

Tried to keep the testing of footnotes as similar as possible to that already in place to make it easier to read. Such a large amount of changes! 

I would be tempted to put the link back in to the reference at some point, so it's easy to jump to the list of footnotes at the bottom (like wikipedia), however with popovers, the click is needed to bring up the popover up. We could use hover instead if we wanted to do this. Didn't quite have time to look at this properly though.


app/components/comparison_table_component.rb
```diff
@@ -21,13 +23,31 @@ def initialize(report:, advice_page:, table_name:, index_params:, headers: [], c
...
+  renders_many :notes, 'ComparisonTableComponent::NoteComponent'
+
+  def before_render
+    collect_references
+  end
+
+  def collect_references
+    seen = {}
+    rows.each do |row|
+      row.to_s # force early render to collect references, haven't found a better way as yet
```
ldodds, line 42: Presumably this causes a "double render" of the row, e.g. its rendered as a string here then later to HTML. I guess this could have performance issues for larger tables?

Was wondering if it'd be possible for the RowComponent or ReferenceComponent instances to call a method on the table component instance (or a shared object) so that the seen values are collected as the table is rendered?

Not suggesting a change now, but wondered if you'd tried that. We can merge this and then see if there are problems.

ldodds (approved):

app/components/comparison_table_component.rb
urbanwide: Fairly sure it's only rendered once - presuming this is the relevant slot code anyway: https://github.com/ViewComponent/view_component/blob/0f953d95b43abdb48dccdac7f6afb157bb8773f6/lib/view_component/slot.rb#L45

But I'll have a proper look when I'm back in (or before then if I get chance). I couldn't work out how to access the table from the Row/Reference, but may have missed something.

PR #3538, opened 2024-04-05
Add campaign landing pages, retire old "for-X" pages

ldodds:
Adds a new landing page to support a social media marketing campaign.

The campaign page will offer a choice of booking a demo or requesting more information. Both options initially take the user to a contact form to complete.

After completing the form we update CapsuleCRM with the details and email an admin. 

If the user requested more information then we send them an email too and show a final confirmation with an embedded video.

If the user requested a demo then they are show a Calendly event booking form. There are different event types for schools (these are group events that many people can join) and MATs (these are 1:1 meetings).

So the main changes are:

- Adds new pages for the landing page, forms and confirmation
- Add a new controller to serve those pages and to support some redirect urls that can be used as more stable links in emails
- A background job and service to handle the Capsule integration and do the email sending
- Embedding of Calendly form

There are specs around the mailer, service and system specs for the UI workflow.

The PR also removes the older marketing and enrol pages in favour of these new pages. So the PR also removes some templates and unused translation text, replacing the routes with redirects.


spec/mailers/campaign_mailer_spec.rb
```diff
@@ -10,20 +10,22 @@
     end
   end
 
+  let(:contact) do
```
tbhi, line 13: makes sense to group with other lets?

app/controllers/landing_pages_controller.rb
```diff
@@ -0,0 +1,148 @@
...
+  def mat_pack
+    redirect_to url_for(controller: :resource_files, action: :download, serve: :inline, id: 30)
+  end
+
+  def school_pack
+    redirect_to url_for(controller: :resource_files, action: :download, serve: :inline, id: 29)
+  end
+
+  def demo_video
+    redirect_to 'https://www.youtube.com/watch?v=x2EeYWwdEpE'
```
tbhi, line 22: do you want to use the no cookies domain?

tbhi (approved):

tbhi:
Next button looks a bit funky while disabled but think has been raised before and is a generic issue

app/controllers/landing_pages_controller.rb
ldodds: thats for embeds. video doesn't display properly (no controls, description, etc) if you try and watch via no-cookies version.

PR #3545, opened 2024-04-10
Updated privacy policy

ldodds:
Updates to our privacy policy to:

- clarify additional data we now collect as part of supporting users and our new contact forms
- list all of the relevant third-party services we now use that may include personal data
- add a note about marketing ads and agencies
- add a note about use provided content, this is already in the terms but is now highlighted here
- update the cookies section with link to new page


ldodds:
Going to merge as Claudia will review text changes

PR #3563, opened 2024-04-15
Add redirect for incorrect urls

ldodds:
We're getting errors from a crawler hitting the interventions page with the id for an activity (for the same school).

Either there's a problem with our internal linking or we previously had issues and the crawler is still fetching old/incorrect urls. I can't find any obviously incorrect urls in templates or redirects in our routes.rb.

As an interim fix, this PR checks to see if the observation that is loaded is actually an activity and, if so, redirects to the right location with a 301 Moved Permanently response.

### DIFF
--- a/app/controllers/schools/interventions_controller.rb
+++ b/app/controllers/schools/interventions_controller.rb
@@ -46,6 +46,11 @@ module Schools
     end
 
     def show
+      if @observation.observation_type == 'activity'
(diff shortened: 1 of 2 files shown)


urbanwide (approved):
Looks good. I've looked quite hard and can't find a place where these links are generated and it seems unlikely that a crawler would go to the trouble of fabricating them in this exact way 🤔 I'll carry on looking for now though!

PR #3567, opened 2024-04-17
New report for Energy Sparks competition

ldodds:
Implements a new report for the Winter Heat Saver competition.

The report uses some new period comparison alerts so the PR bumps the analytics and includes the database changes to register those and the new report.

The report consists of 4 tables (a total and one each for electricity, gas and storage heaters) and a chart.

The views can eventually be repurposed to become the basis for the new configurable period reports so I've spent a bit of time ensuring they have a sensible table structure and footnotes, although there's more that could be tidied.

I've also included a new concern `MultipleFuelComparisonView` which can be mixed into views that have columns for the three fuel types across multiple periods. The concern adds some helper methods for calculating totals and percentage changes. Some of the final set of Total Energy Use reports we have to implement will also use this.

I've also extracted some of the simple calculation functions into a separate `Calculator` class. These functions are mostly used by the existing benchmarking code. We're using some of them via helpers but as I wanted to use them in a model I've extracted into a separate class.

### DIFF
--- a/app/models/concerns/multiple_fuel_comparison_view.rb
+++ b/app/models/concerns/multiple_fuel_comparison_view.rb
@@ -43,20 +43,29 @@ module MultipleFuelComparisonView
     percent_change(total_previous_period(unit: unit), total_current_period(unit: unit))
   end
 
-  # Returns an array of field names for the specified unit, for the previous period
-  # E.g. [:electricity_previous_period_kwh, :gas_previous_period_kwh]
+  # Returns an array of values for the specified unit, for the previous period
+  # E.g. by calling :electricity_previous_period_kwh, :gas_previous_period_kwh, etc
   def all_previous_period(unit: :kwh)
     field_names(period: :previous_period, unit: unit).map do |field|
       send(field)
     end
   end
 
-  # Returns an array of field names for the specified unit, for the current period
-  # E.g. [:electricity_current_period_kwh, :gas_current_period_kwh]
+  # Returns an array of values for the specified unit, for the current period
+  # E.g. by calling :electricity_current_period_kwh, :gas_current_period_kwh, etc
   def all_current_period(unit: :kwh)
     field_names(period: :current_period, unit: unit).map do |field|
       send(field)
     end
+  end
+
+  def field_names(period: :previous_period, unit: :kwh)
+    unit = :gbp if unit == :£
+    field_names = []
+    fuel_types.each do |fuel_type|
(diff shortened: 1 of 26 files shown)


app/views/comparisons/heat_saver_march2024/_total.html.erb
```diff
@@ -0,0 +1,50 @@
...
+    <% c.with_row do |r| %>
+      <% r.with_school school: result.school %>
+      <% r.with_reference key: 'tariff_changed_last_year',
+                          if: result.any_tariff_changed? %>
+      <% r.with_reference key: 'electricity_change_rows',
+                          period_type_string: @period_type_string,
+                          if: result.pupils_changed %>
+      <% r.with_reference key: 'gas_change_rows',
+                          period_type_string: @period_type_string,
+                          schools_to_sentence: 'some schools',
```
ldodds, line 16: Added to make the Footnote work, but don't think this is required, so will update the footnote text and the code separately.

app/models/concerns/multiple_fuel_comparison_view.rb
```diff
@@ -0,0 +1,96 @@
...
+  def field_names(period: :previous_year, unit: :kwh)
+    unit = :gbp if unit == :£
+    field_names = []
+    fuel_types.each do |fuel_type|
+      field_names << field_name(fuel_type, period, unit)
+    end
+    field_names
+  end
+
+  def percent_change(base, new_val, to_nil_if_sum_zero = false)
```
urbanwide, line 85: Can the class methods from the Calculator class be included here rather than redefining them?

urbanwide (approved):
Looks good so far - see inline comment 👍

app/models/concerns/multiple_fuel_comparison_view.rb
ldodds: I could just call them directly rather than having local aliases, is that what you mean? That would probably be better. 

I'm still reviewing what the old framework is doing for some of the cases where there is limited data. I might end up removing the "sum_if_complete" anyway, in favour of more table footnotes/explanation.

urbanwide: Just meant rather than redefining to either include (if Calculator is a module, which I've just realised is not) or maybe extend Calculator to bring the methods in. Or maybe delegate? But calling them directly would also be better than aliasing them like this I think.

PR #3569, opened 2024-04-19
Add additional school features to edit school details & school estates page + issues search

urbanwide:
* Note search: wasn't sure if it was intended to just run an ad-hoc query to return a list of notes or to add a feature - it was as easy to add a search feature to issues/notes, so I did that. Hope that's ok. It looks in the title or description for search terms. You can do an or by typing in GSHP|ground source heat pump. 
* Added new fields - just the checkboxes to the edit school details page, but also the percent and notes fields to the school estates page (as per other similar fields)
* Removes your_school_estates feature flag, which can be removed from test/production env too once this has been deployed

Might benefit if we all had a new database dump as getting big conflicts for schema.rb ?

### DIFF
--- a/app/views/schools/your_school_estates/edit.html.erb
+++ b/app/views/schools/your_school_estates/edit.html.erb
@@ -21,28 +21,64 @@
       <p><%= t('schools.your_school_estates.edit.alternative_heating_sources.description_1') %>.</p>
       <p><%= t('schools.your_school_estates.edit.alternative_heating_sources.description_2') %>.</p>
 
-      <%= f.input :alternative_heating_oil, label: t('schools.your_school_estates.edit.alternative_heating_sources.oil.label'), input_html: {data: {reveals: '.school_alternative_heating_oil_fields' }} %>
+      <%= f.input :alternative_heating_oil,
+                  label: t('.alternative_heating_sources.oil.label'),
+                  input_html: { data: { reveals: '.school_alternative_heating_oil_fields' } } %>
       <div class='school_alternative_heating_oil_fields' data-revealed-by=".school_alternative_heating_oil">
-        <%= f.input :alternative_heating_oil_percent, label: t('schools.your_school_estates.edit.alternative_heating_sources.oil.percentage') %>
-        <%= f.input :alternative_heating_oil_notes, label: t('schools.your_school_estates.edit.alternative_heating_sources.notes') %>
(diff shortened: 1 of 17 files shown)


app/views/schools/your_school_estates/edit.html.erb
```diff
@@ -21,28 +21,64 @@
...
-        <%= f.input :alternative_heating_district_heating_percent, label: t('schools.your_school_estates.edit.alternative_heating_sources.district_heating.percentage') %>
-        <%= f.input :alternative_heating_district_heating_notes, label: t('schools.your_school_estates.edit.alternative_heating_sources.notes') %>
+        <%= f.input :alternative_heating_district_heating_percent,
+                    label: t('.alternative_heating_sources.district_heating.percentage') %>
+        <%= f.input :alternative_heating_district_heating_notes,
+                    label: t('.alternative_heating_sources.notes') %>
+      </div>
+
+      <%= f.input :alternative_heating_ground_source_heat_pump,
+                  label: t('.alternative_heating_sources.ground_source_heat_pump.label'),
```
urbanwide, line 65: Had to change all the i18n keys in this section to be relative rather than absolute in order to please erblint line length

config/locales/views/schools/schools.yml
```diff
@@ -689,6 +689,9 @@ en:
     your_school_estates:
       edit:
         alternative_heating_sources:
+          air_source_heat_pump:
+            label: Our school has an air source heat pump
+            percentage: What percentage of the school using the air source heat pump for heating?
```
urbanwide, line 694: I am not sure this actually makes sense, should it be 'is using'? I decided to use this sentence structure to be the same as the other alternative heating source fields. Which perhaps means these don't read correctly either? 🤔

Also the other ones capitalise the heating source, which also doesn't always seem right.

ldodds (approved):
Looks good. Couldn't think of a better way of handling the relative translation keys other than doing string concatenation in the template. Left a note about the labelling.

re: the search, I think adding a basic search like this is useful. We might want to improve it later to use pg_search. 

Can you follow up with team to indicate that they can use the search to then go back through and update schools that have heat pumps so we can get a better idea of level of use?

urbanwide:
Have run the migrations from a fresh db import and checked in the schema. So hopefully this is the right way now 👍

PR #3571, opened 2024-04-19
Remove n3rgy v1

ldodds:
Removes the code that works with v1 of the n3rgy API now that we've completed the upgrade.

- removes all the conditional code that checks for feature flag
- remove tests for both v1 and v2 of the api
- removes methods from MeterManagement class which are replaced by N3rgyMeteringService
- fixes up other specs that refer to the old client or the above code
- remove the 'sandbox' attribute from Meter as this is no longer needed and the sandbox meters have all been deleted

Code is running against this analytics branch (https://github.com/Energy-Sparks/energy-sparks_analytics/pull/688) to confirm that specs pass without the v1 code in the analytics. Will update to tagged version before merging.

I've also manually run the data import and confirmed that readings are being loaded correctly.


spec/services/amr/n3rgy_readings_download_and_upsert_spec.rb
```diff
@@ -20,55 +20,77 @@ module Amr
...
+    describe '#perform' do
+      context 'with an error downloading data' do
         it 'handles and log exceptions' do
           expect(AmrDataFeedImportLog.count).to be 0
-          expect(n3rgy_api).to receive(:readings).and_raise(StandardError)
-          upserter = Amr::N3rgyReadingsDownloadAndUpsert.new(n3rgy_api_factory: n3rgy_api_factory, config: config, meter: meter, start_date: start_date, end_date: end_date)
+          allow(Amr::N3rgyDownloader).to receive(:new).and_return(downloader)
+          expect(downloader).to receive(:readings).and_raise(StandardError)
+
+          upserter = Amr::N3rgyReadingsDownloadAndUpsert.new(config: config, meter: meter, start_date: start_date, end_date: end_date)
```
tbhi, line 32: good time to change to described_class?

tbhi (approved):

urbanwide (approved):
👍

PR #3578, opened 2024-04-22
Optimise loading of Footnote in comparison tables

ldodds:
When we add references to a table, the ReferenceComponent loads its Footnotes. Internally its cached, but as we create a new instance of each ReferenceComponent in a table, we end up loading the footnotes multiple times. This leads to lots of additional model instances being created.

This is a draft change that changes the reference to allow the footnote to be passed in, rather than loaded on demand.

This speeds up the report views and reduces memory overhead.

A further optimisation would be to only add the ReferenceComponent if the `if` check is true, rather than evaluating within the component. The same ReferenceComponent instance could also be shared across table rows (and potentially across tables), which would further reduce memory overhead.

### DIFF
--- a/app/components/comparison_table_component.rb
+++ b/app/components/comparison_table_component.rb
@@ -21,32 +21,29 @@ class ComparisonTableComponent < ViewComponent::Base
     @headers = headers
     @colgroups = colgroups
     @advice_page_tab = advice_page_tab
+    @comparison_footnotes = {}
+    @seen = {}
   end
 
   renders_many :rows, ->(**kwargs) do
     kwargs[:advice_page] = @advice_page
     kwargs[:advice_page_tab] = @advice_page_tab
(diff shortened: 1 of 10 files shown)


urbanwide:
Made a few changes:
* main one being passing a reference to the main component instance down to slots, so they can call methods on it. This means we can cache a list of footnote objects there as they are used rather than passing them in at the top level.
* Changing the references slot definition to use a lambda has meant we can collect references for footnotes on the way in rather than having to do it afterwards with the old collect_references method. This is much better!
* Also changed it so reference components are only created if the "if" evaluates to true.
* The footnote components are now sorted when rendered rather than when created
* I also spotted that multiple footnotes weren't separated with newlines, so sorted that out.
* Also added some more thorough tests for the footnotes.

urbanwide (approved):

PR #3582, opened 2024-04-25
Always reload last week of data from n3rgy

ldodds:
Further revision to the logic around loading data from n3rgy.

Their import processes regularly reloads the last 7 days of data to incorporate any missing readings. So ensure that we always reload the last 7 days every day to ensure that we have the latest data.

This means repeated daily requests, but ensures we still as current as possible.

### DIFF
--- a/app/services/amr/n3rgy_readings_download_and_upsert.rb
+++ b/app/services/amr/n3rgy_readings_download_and_upsert.rb
@@ -26,11 +26,6 @@ module Amr
       # incomplete readings.
       return if start_date > end_date
 
-      # don't load data if we are up to date, indicated by the
-      # start and ends being same date, after winding back the end date
-      # from midnight
-      return if existing_n3rgy_readings.any? && start_date.to_date == (end_date - 1).to_date
-
       import_log = create_import_log(start_date, end_date)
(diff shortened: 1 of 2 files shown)


app/services/amr/n3rgy_readings_download_and_upsert.rb
```diff
@@ -57,10 +52,11 @@ def create_import_log(start_date, end_date)
 
     def current_date_range_of_readings
       if existing_n3rgy_readings.any?
-        first = existing_n3rgy_readings.minimum(:reading_date)
-        last = existing_n3rgy_readings.maximum(:reading_date)
-        (n3rgy_first_reading_of_day(DateTime.parse(first + 'T00:30'))..
-          n3rgy_last_reading_of_day(DateTime.parse(last + 'T00:00')))
+        first = DateTime.parse(existing_n3rgy_readings.minimum(:reading_date))
```
urbanwide, line 57: Assume it's ok that you've lost the `+ 'T00:30'` etc from the params to DateTimeParse here?

urbanwide (approved):
Looks good - just wanted to check one thing - see inline!

PR #3583, opened 2024-04-25
Produce public facing reports for the custom period comparisons

tbhi:


app/models/concerns/arbitrary_period_comparison_view.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+module ArbitraryPeriodComparisonView
```
ldodds, line 3: I've [moved some of these methods](https://github.com/Energy-Sparks/energy-sparks/pull/3521/files#diff-2f77ae14cf68c21201b5b2d7044091025c1b9ab86293237ae4a7d3f6be359a94) to `View` base class in my report PR with some tidying and comments. I can rework after this has merged.

ldodds (approved):
Looks good. We'll need to get these listed in the "Date limited comparisons" section of the comparison report page. Might probably best to do that in a separate PR

PR #3588, opened 2024-04-29
Set maximum frequency for scoring on activities and interventions

ldodds:
Adds a `maximum_frequency` attribute to ActivityType and InterventionType

Adds a `Reportable` concern to try and centralise the logic around when to award a point score. This is implemented by the above. 

For the specs I've added some shared examples. These are parameterised to pass in the factories used to record the Recordable objects and the instances of the things that are recorded using them. Possibly slightly unusual, but seemed better than just duplicating the tests.

Reminds me that we need to iron out more of the inconsistencies around ActivityType and InterventionType and the related recordings.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -3067,6 +3067,110 @@ ActiveRecord::Schema.define(version: 2024_04_29_144652) do
             ORDER BY alert_generation_runs.school_id, alert_generation_runs.created_at DESC) latest_runs
     WHERE (data.alert_generation_run_id = latest_runs.id);
   SQL
+  create_view "heat_saver_march_2024s", sql_definition: <<-SQL
+      WITH electricity AS (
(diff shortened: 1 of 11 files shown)


db/schema.rb
```diff
@@ -3065,110 +3067,6 @@
             ORDER BY alert_generation_runs.school_id, alert_generation_runs.created_at DESC) latest_runs
     WHERE (data.alert_generation_run_id = latest_runs.id);
   SQL
-  create_view "heat_saver_march_2024s", sql_definition: <<-SQL
```
urbanwide, line 3068: I think this has slipped in from your other branch...

ldodds: Ugh. Will fix.

urbanwide (approved):
Looks good - I like how you have shared the functionality between activities and interventions

PR #3589, opened 2024-04-29
Display unlisted schools in comparison reports

urbanwide:
Currently looks like this:

image

image

It derives the count of unlisted schools by subtracting the count of the result set from the original list of schools. 

Clicking on the list of schools runs the original queries again, subtracts the schools returned in the query from the filtered schools. It possibly could be made quicker to load by passing a list of unlisted schools extracted from the original query, but this would likely add processing time to the original page.

The modal links to the analysis page for each school - as well as the school. Not sure about the analysis link as it quite often is a link to a page where it shows no data... See what you think.

This PR also adds a Link component to the FootnoteModal component, so have added this in other parts of the codebase where for links where FootnoteModal is in use.

### DIFF
--- a/app/components/comparison_table_component.rb
+++ b/app/components/comparison_table_component.rb
@@ -227,8 +227,12 @@ class ComparisonTableComponent < ViewComponent::Base
       @note = note
     end
 
+    def note
+      @note || content || ''
+    end
+
     def call
-      (@note || content).html_safe
+      note.html_safe
     end
   end
 end
--- a/app/components/comparison_table_component.rb
+++ b/app/components/comparison_table_component.rb
@@ -75,12 +75,7 @@ class ComparisonTableComponent < ViewComponent::Base
 
     # First column, showing school name and a link
     renders_one :school, ->(school:) do
-      path = if @advice_page.present?
-               helpers.advice_page_path(school, @advice_page, @advice_page_tab)
-             else
-               school_advice_path(school)
-             end
-      link_to school.name, path
+      link_to school.name, helpers.advice_page_path(school, @advice_page, @advice_page_tab)
     end
 
     # Footnote references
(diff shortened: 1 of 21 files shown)


ldodds (approved):
Looks good to me. Couple of comments, which might be addressed in follow-up:

- currently the best place to confirm how out of date the data is for a school is its dashboard as we have summary of available data there. So I'm wondering if we should link there, rather than the advice page index?
- I think the screenshots are for an older version? Was going to comment about plurals but looks like the YML handles that
- loading the schools only when needed and trying optimise the main page load seems reasonable to me
- we have a separate card to exclude schools with swimming pools from the hot water analysis, ideally we'd customised the unlisted schools and text for that so its clear which ones are excluded because of that vs limited data.

urbanwide:
> Looks good to me. Couple of comments, which might be addressed in follow-up:
> 
> * currently the best place to confirm how out of date the data is for a school is its dashboard as we have summary of available data there. So I'm wondering if we should link there, rather than the advice page index?
OK. There are currently two links - the name of the school links to the dashboard and the "View analysis" links to the advice page configured for the report. Shall I just remove the "View analysis" link all together?

> * I think the screenshots are for an older version? Was going to comment about plurals but looks like the YML handles that
Yes sorry they appear to be, didn't realise.

> * loading the schools only when needed and trying optimise the main page load seems reasonable to me
Great, thanks

> * we have a separate card to exclude schools with swimming pools from the hot water analysis, ideally we'd customise the unlisted schools and text for that so its clear which ones are excluded because swimming pools vs those with limited data.
Great - I'll have a look for the ticket and put a note on it.

ldodds:
For the unlisted schools I'd just link to the dashboard and remove the analysis link completely.

PR #3592, opened 2024-04-30
Allow admins to edit max frequency, show frequency to users

ldodds:
* [x] Updates admins forms so that the max frequency can be updated
* [x] Updated forms for those recording activities and actions to add some clearer messaging (and tweak form layouts slightly)
* [x] Updated forms to indicate if school has exceeded threshold and won't score any points


urbanwide (approved):
Looks good - if there's not a ticket elsewhere for this - we also need to update the points scored on the completed action / activity pages as it's displayed there too.

PR #3596, opened 2024-04-30
Add data feed config for EfT Centrica Historic Solar

ldodds:


tbhi (approved):
could use ' everywhere?

PR #3597, opened 2024-05-01
Transifex update 2024-05-01

ldodds:
Updates the transifex config for synchronisation. Latest versions of YML were then pushing using that config.

Also pulls in a couple of minor changes.

### DIFF
--- a/config/locales/cy/views/advice_pages/gas_costs.yml
+++ b/config/locales/cy/views/advice_pages/gas_costs.yml
@@ -5,7 +5,7 @@ cy:
         comparison:
           title: Cymhariaeth cost ar gyfer y 2 flynedd ddiwethaf
         cost_breakdown_by_charge:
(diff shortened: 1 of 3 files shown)


config/locales/cy/views/advice_pages/gas_costs.yml
```diff
@@ -5,7 +5,7 @@ cy:
         comparison:
           title: Cymhariaeth cost ar gyfer y 2 flynedd ddiwethaf
         cost_breakdown_by_charge:
-          title: Dadansoddiad cost yn ôl tâl am y %{ cyfnod} diwethaf
+          title: Dadansoddiad cost yn ôl tâl am y %{period}diwethaf
```
tbhi, line 8: should it have space after %{period}?

ldodds: Fixed here and 1 other place. Also updated in Transifex

PR #3598, opened 2024-05-01
Initialise tooltips and popovers on all pages

ldodds:
Currently Bootstrap tooltips and popovers are initialised in `analysis.js`

- all tooltips are initialised when page is ready, aftering setting up charts
- tooltips in chart annotations are initialised after the annotations have been loaded and added to the chart
- popovers are only initialised when a chart has been successfully loaded

This means that if a page doesn't have a chart then popovers don't work. Tooltips will be working everywhere though.

We also use popovers in the calendar views but these are explicitly initialised already.

I've changed things so that:

- there's a separate `init_bootstrap.js` file which initialises all tooltips and popovers. Seemed better to have that more clearly defined
- removed the call to initialise popovers in `analysis.js` as it doesn't seem to be needed there. The charts only use tooltips, not popovers, and the tooltips are already handled when chart is refreshed
- clarified some comments in the code

I've manually confirmed that:

- annotations still work on dashboards
- chart controls still work
- tooltips in advice pages work
- popovers work on comparison reports with and without charts
- calendar views work OK

### DIFF
--- a/app/assets/javascripts/analysis.js
+++ b/app/assets/javascripts/analysis.js
@@ -145,9 +145,6 @@ function processAnalysisCharts(){
       }
     });
   }
-
-  //activate tooltips
-  $('[data-toggle="tooltip"]').tooltip();
 }
 
 function processAnalysisChartAjax(chartId, chartConfig, highchartsChart) {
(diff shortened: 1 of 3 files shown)


urbanwide (approved):
This is a good find @ldodds. I had assumed it was switched on for all pages as it "just worked". Cheers 👍

PR #3599, opened 2024-05-01
Change in energy use since joining report

ldodds:
Implements report to compare energy use since school joined Energy Sparks and the current year.

This will not completely match the current report as we're going to use the same set of views as for the period comparisons, to ensure consistency but some details will be different.

We will also be removing the data relating to change in solar consumption as this is not needed and doesn't make sense to report here (we have no impact over changes in solar generation, any differences will likely be due to installing more panels or  weather).

* [x] Implement basic framework and view
* [x] Ensure where clauses and filters are working correctly, cross-referencing with other reports
* [x] Rationalise queries across the different concerns and the View model, after confirming differences across reports
* [x] Implement spec for new report


app/controllers/comparisons/change_in_energy_use_since_joined_energy_sparks_controller.rb
```diff
@@ -0,0 +1,28 @@
+module Comparisons
+  class ChangeInEnergyUseSinceJoinedEnergySparksController < Shared::ArbitraryPeriodController
+    def key
+      :change_in_energy_use_since_joined_energy_sparks
+    end
+
+    def set_headers
+      super
+      @include_previous_period_unadjusted = false
```
tbhi, line 9: don't need the rest if call super after?

tbhi (approved):

app/controllers/comparisons/change_in_energy_use_since_joined_energy_sparks_controller.rb
ldodds: The base class will reset the variable to true, so the headers will end up wrong. I've removed the call to super, which was there to ensure that the `BaseController` version was called and just overridden the method entirely

PR #3600, opened 2024-05-02
Add comparison report groups

urbanwide:
**Adds flipper**

- [x] Add a Comparison Report Group model. It will have a title, description (as currently shown on the list) and an order (numeric). It may have many Comparison Reports
- [x] The Report Group title and description should be synched with Transifex
- [x] We migrate the existing report group title, description from the YAML and order from the existing configuration
- [x] We generate the comparison report index from the report groups and reports. The groups are ordered by their order.
- [x] Within a group the reports are ordered by name.
- [x] We update Comparison::Report so that is belong to a report group, updating the forms to allow the report group to be picked by an admin. A Report must have a group

Doing the next tasks as a new branch
- Create admin interface for report groups and add inline edit buttons. Don't allow those attached to a report to be deleted. 
- Add Comparison admin menu to all comparison admin pages

### DIFF
--- a/app/views/admin/index.html.erb
+++ b/app/views/admin/index.html.erb
@@ -69,6 +69,7 @@
       <li><%= link_to 'Transport Types', admin_transport_types_path %></li>
       <li><%= link_to 'Videos', admin_videos_path %></li>
       <li><%= link_to 'Background Jobs', admin_good_job_path %></li>
+      <li><%= link_to 'Feature Flags', admin_flipper_path %></li>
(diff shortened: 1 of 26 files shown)


app/views/admin/index.html.erb
ldodds, line 1: Maybe also add links to the Report Group (and Report and Footnote) forms?

urbanwide: Thanks Leigh - have done this in the next PR

PR #3601, opened 2024-05-02
Error in Storage Heater Consumption during Holiday report

tbhi:


ldodds (approved):

app/models/chart_data_values.rb
```diff
@@ -71,7 +71,7 @@ def translate_categories_for(categories)
   end
 
   def translate_data_keys_for(data)
-    return unless data.present?
+    return if data.nil?
```
ldodds, line 74: Wondering if it should even reach here though?

Should `ComparisonChartComponent` implement `.render?` (https://viewcomponent.org/api.html#render--boolean) to check if there's any data available to add to a chart, or was the issue upstream from this?

PR #3608, opened 2024-05-08
Report group admin interface

urbanwide:
Branched off https://github.com/Energy-Sparks/energy-sparks/pull/3600 which will need reviewing first

- [x] Create admin interface for report groups
- [x] Add Comparison admin menu to all comparison admin pages
- [x] Spec admin interface
- [x] Add inline edit buttons for report groups.
- [x] Spec inline edit buttons
- [x] Don't allow report groups attached to a report to be deleted.
- [x] Add comparisons to main admin page / menu


app/helpers/application_helper.rb
```diff
@@ -515,4 +515,14 @@ def marketing_mat_count
   def round_down_to_nearest_ten(val)
     (val / 10) * 10
   end
+
+  def admin_link(path, to: 'Edit', tag: nil, classes: nil)
+    if current_user&.admin?
+      link = link_to to, path,
+                  class: "badge badge-light font-weight-normal border align-text-top #{classes}",
```
ldodds, line 522: I'm not sure about using badges rather than proper buttons, I'm not sure its immediately obvious that a badge style link means you can edit something. Most other places we're just using buttons I think. Can you change?

ldodds (approved):
One comment on button styling.

Also, now that we have Flipper we should take stock of any use of the old FeatureFlags code so we can either migrate or remove them.

PR #3611, opened 2024-05-09
Add table of contents and table titles to comparison reports with multiple tables

ldodds:
Some comparisons reports have multiple tables. To make it easier to navigate within potentially longer pages, this PR adds support for: 

- adding consistently formatted titles to each table
- automatically adding a table of contents linking to each title
- in-page links taking user back to the top of the page

Controllers that produce reports with multiple tables should include a new concern `MultipleTableComparison` and implement the `table_configuration` method rather than `table_names`. That method should return a hash of table name -> translated table title.

The PR updates the shared templates to include the additional markup where needed. The period comparison and heating coming on too early reports have been updated to use this new feature.


app/views/comparisons/base/_tables.html.erb
```diff
@@ -1,4 +1,19 @@
+<% if @table_names.size > 1 %>
+  <div class="mt-4" id="table-list">
+    <ul>
+      <% @table_names.each_with_index do |table_name, index| %>
```
tbhi, line 4: each.with_index(1) would be better?

tbhi (approved):

PR #3622, opened 2024-05-14
Rationalise site colours

urbanwide:
Large reorg of the colours used in the site.

- Created some new colour names / variables for rogue colours found on the site. Or used an existing variable if available or one that is a very similar colour. In particular there are many shades of grey that we will want to make sure are from a similar palette mix, they are a bit all over the place currently. Otherwise might choose to switch them out for bootstrap greys (gray-100 to gray-900). Also tried to make sure we aren't using the same colour names as the bootstrap variables, for now anyway.
- Introduced initializers/colours.rb as a central place to store all colours used on the site. These can be used in the css and ruby / javascripts. Refactored colours.scss to have only colour variables in.
- Renamed colours.scss to colours.scss.erb to so the colours from the initializer can be referenced.
- Renamed some javascript files in the same way, to also make use of the colours.
- Moved transport survey javascript from packs, so this could be done here too. Also makes things easier when we move away from Webpacker.
- Transport survey pie chart colours are now less pastel but other than that, as far as I can tell, everything looks the same.
- Quite a lot of css refactoring in affected areas. Could do this a lot more :-)

Think it would be a good idea to put this on test for a few days before going live...


app/helpers/schools_helper.rb
```diff
@@ -5,10 +5,6 @@ def kid_date(date)
     date.strftime('%A, %d %B %Y')
   end
 
-  def colours_for_supply(supply)
```
urbanwide, line 8: Unused, so removed.

app/assets/stylesheets/nav.scss
```diff
@@ -133,10 +133,10 @@
 
 .nav-lozenge {
   white-space: nowrap;
-  background: #232B49;
-  background-image: linear-gradient(180deg, #252E50 0%, #374268 99%);
+  background: $dark-blue;
```
urbanwide, line 136: Have changed this so we use $dark-blue as the base and taking shades of it, rather than adding more colours to the palette

ldodds:
Will take a look at this later today, as a lot to review. Agree it'd be best to release to test. So perhaps we do that tomorrow so it can be reviewed ahead of Thursday release.

config/initializers/colours.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  # Other non-standard colour definitions can be found in:
+
+  # assets/javascripts/common_chart_options.js
+  # colors: ["#9c3367", "#67347f", "#935fb8", "#e676a3", "#e4558b", "#7a9fb1", "#5297c6", "#97c086", "#3f7d69", "#6dc691", "#8e8d6b", "#e5c07c", "#e9d889", "#e59757", "#f4966c", "#e5644e", "#cd4851", "#bd4d65", "#515749"],
+
+  # Admin areas / mailers e.g.:
+  # controllers/admin/reports/amr_validated_readings_controller.rb
+
+  # test_warning.scss which seems unused
```
ldodds, line 78: Yes, this can be removed

ldodds:
I've checked this out locally, rebuilt all the assets and looked over the affected parts of the site including home/schools/dashboards/advice pages, as well as trying out the transport survey app. Nothing obviously broken or visually jarring so if you've tweaked a few colours they didn't jump out at me.

Initially I was wondering how the `.erb.js` and `.erb.scss` files would work with the asset pipeline, but looks like they just get evaluated first? I didn't realise that was an option. If that works with Rails 7 then seems like a big improvement.

Was slightly confused initially as Chrome is showing me links to, e.g. common chart options as an ERB file with unevaluated Ruby in it, but I'm not sure its actually using that version of the file to display the pages.

urbanwide:
Have now removed the unused test_warning.scss, it's import line in application.scss and  _test_site_warning.html.erb.

A quick look at Rails 7 indicates the asset pipeline (Sprockets) allows .css.erb and .js.erb to be preprocessed. Not sure about Propshaft without trying it, which looks like it is heading towards being standard in Rails 8 but maybe cross that bridge when we come to it.

Also just been a bit confused by Chrome dev tools showing common_chart_options.js.erb listed as an asset, but been reading about it and I think this is down to source map generation that Webpacker does, I think only in development.

ldodds (approved):
Looked over the new changes look OK to me.

ldodds:
> A quick look at Rails 7 indicates the asset pipeline (Sprockets) allows .css.erb and .js.erb to be preprocessed. Not sure about Propshaft without trying it, which looks like it is heading towards being standard in Rails 8 but maybe cross that bridge when we come to it.

OK that's good to know.

> Also just been a bit confused by Chrome dev tools showing common_chart_options.js.erb listed as an asset, but been reading about it and I think this is down to source map generation that Webpacker does, I think only in development.

That's what was confusing me too. But I couldn't find any evidence of breakage and using, e.g. the inspector confirmed that colours where coming from the compiled files. I tried rebuilding assets a few times to confirm.

Doing a test release will check out how the asset pipeline works during and after a release.

PR #3636, opened 2024-05-20
Remove schools with swimming pools from hot water report

ldodds:
Extracts the message for unlisted schools into a helper method that can be overridden on a per-controller basis.

Adds filter to where clause

Overrides the unlisted school modal to indicate whether a school has a swimming pool in the list.

### DIFF
--- a/spec/system/comparisons/hot_water_efficiency_spec.rb
+++ b/spec/system/comparisons/hot_water_efficiency_spec.rb
@@ -36,6 +36,7 @@ describe 'hot_water_efficiency' do
     it_behaves_like 'a school comparison report with a table' do
       let(:expected_report) { report }
       let(:expected_school) { school }
+
       let(:advice_page_path) { polymorphic_path([:insights, expected_school, :advice, advice_page_key]) }
 
       let(:headers) do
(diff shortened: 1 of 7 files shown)


spec/system/comparisons/hot_water_efficiency_spec.rb
```diff
@@ -59,4 +60,24 @@
...
+    end
+
+    it 'still has a single school' do
+      within("##{report.key}-table") do
+        expect(page).to have_link(school.name)
+        expect(page).not_to have_link(other_school.name)
+      end
+    end
+
+    it 'has custom unlisted message' do
```
urbanwide, line 79: Is it worth testing the different output in the modal too i.e listing if the school has a swimming pool?

urbanwide (approved):
Looks good - small comment about a possible extra test

spec/system/comparisons/hot_water_efficiency_spec.rb
ldodds: Yes. Although struggling to get the test to actually open the modal at the moment. Following same pattern as in compare_spec.rb but any assertions fail to find `.modal-body`

urbanwide:
Have you got js: true switched on?

spec/system/comparisons/hot_water_efficiency_spec.rb
urbanwide: Have you got js: true switched on?

PR #3638, opened 2024-05-20
Auto generate comparison report key

urbanwide:


tbhi:
Should the key change if title changes?

ldodds (approved):

urbanwide:
> Should the key change if the title changes? Trying it locally and doesn't seem to but not sure of the desired behaviour.

@tbhi I have assumed not - because these keys are referenced within the comparison files and would require those to be changed too. I may add a test to assert that though, thanks.

ldodds:
Yep, as they're used in the comparison report and are part of the public link, it shouldn't change.

If we change the keys for a report we'll need to add redirects to preserve urls.

ldodds (approved):

PR #3639, opened 2024-05-21
Annual energy use report

ldodds:
A new report, intended to replace the "annual change in cost" report.

Shows the kwh, £ and co2 for each fuel type for the last 12 months. Includes metadata to indicate if school data is up to date, school type, number of pupils and floor area.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -86,6 +86,7 @@ Rails.application.routes.draw do
   end
 
   namespace :comparisons do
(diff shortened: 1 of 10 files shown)


config/routes.rb
```diff
@@ -86,6 +86,7 @@
   end
 
   namespace :comparisons do
+    resources :annual_energy_use, only: [:index], concerns: :unlisted
```
tbhi, line 89: sorting?

tbhi (approved):

PR #3646, opened 2024-05-21
Fix transport js

urbanwide:
The changes discussed earlier did indeed allow the code to be deployed, however, the transport app loaded, but still didn't work properly - erroring with the following files missing:
Screenshot 2024-05-21 at 23 02 23

Which I think would make sense, as they are not listed in the precompile config line in assets.rb
(worked locally though, that's why I didn't add them)

I have now added them to this branch - which bought up a couple of similar uglifier issues using ||= operator in storage.js. Have also fixed those in this branch.

Ended up having to list all the individual assets, as transport_surveys/* glob doesn't appear to work

`rake assets:clobber assets:precompile` shows all assets compiling now.

### DIFF
--- a/config/initializers/assets.rb
+++ b/config/initializers/assets.rb
@@ -14,5 +14,6 @@ Rails.application.config.assets.precompile += %w( application_mailer.scss )
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
 # Rails.application.config.assets.precompile += %w( admin.js admin.css )
-# A note about sprockets v4 and using manifest.js:
+
+# A note about sprockets v4 and using manifest.js for assets instead:
 # https://github.com/rails/sprockets/blob/main/UPGRADING.md#manifestjs
--- a/config/initializers/assets.rb
+++ b/config/initializers/assets.rb
@@ -13,5 +13,4 @@ Rails.application.config.assets.precompile += %w( application_mailer.scss )
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
-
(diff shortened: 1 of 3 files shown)


config/initializers/assets.rb
```diff
@@ -13,5 +13,4 @@
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
-
-Rails.application.config.assets.precompile += %w( transport_surveys transport_surveys/charts.js )
+Rails.application.config.assets.precompile += %w( transport_surveys.js transport_surveys/carbon.js transport_surveys/charts.js transport_surveys/handlebars_helpers.js transport_surveys/helpers.js transport_surveys/notifier.js transport_surveys/storage.js )
```
ldodds, line 16: This seems fine but wondering with the glob if you'd tried something like: 

```
Dir[Rails.root.join('app/assets/javascripts/transport_surveys/*')].map { |f| "transport_surveys/#{File.basename(f)}" }
```

To build the array? Note this only adds the js files within the transport_survey directory, not the main transport_surveys.js file. So will look even uglier to add that, but might be more stable?

urbanwide: Thanks Leigh, agree it was messy. Have been reading around and think procs / globs no longer work here with later versions of sprockets, in favour of using manifest.js : https://github.com/rails/sprockets/blob/main/UPGRADING.md#manifestjs

urbanwide: See what you think of the latest change anyway... I'm hopeful :-)

ldodds (approved):

PR #3651, opened 2024-05-22
Add new school comparison for % of gas used for heating vs hot water

tbhi:


app/controllers/comparisons/heating_vs_hot_water_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    def key
+      :heating_vs_hot_water
+    end
+
+    def advice_page_key
+      :hot_water
+    end
+
+    def load_data
+      Comparison::HeatingVsHotWater.for_schools(@schools).where.not(last_year_gas_kwh: nil).order(last_year_gas_kwh: :desc)
```
ldodds, line 29: The where clause here excludes schools with no gas usage, which is correct but it means that the report will include schools that have no hot water usage either. I think it needs to check whether we have both. This ensures that we:

- rule out schools where we don't have gas data (no either analysis is present)
- rule out schools where we aren't successfully able to run the hot water analysis, e.g. due to limited data. The AlertHotWaterEfficiency analysis only runs if we have data covering a summer holiday with a few weeks either side

As noted on Trello the report should also exclude schools with swimming pools, like the Hot Water Efficiency report. This model defines an extra scope which is used in the controller:

```
scope :without_swimming_pool, -> { joins(:school).where({ school: { has_swimming_pool: false } }) }
```

That report also:

- uses a different message for reporting which schools are unlisted to add a note about swimming pools, see `unlisted_message` in controller
- updates the list of unlisted schools to indicate which ones have swimming pools. See `hot_water_efficiency/_unlisted.html.erb`

The Trello card also notes that the report should be sorted by % use of gas for heating, in descending order. Should be possible to calculate that in the SQL query as we do with other % comparisons.

ldodds (changes requested):

Gemfile
```diff
@@ -26,7 +26,7 @@ gem 'pg'
 gem 'scenic'
 
 # Dashboard analytics
-gem 'energy-sparks_analytics', github: 'Energy-Sparks/energy-sparks_analytics', tag: '5.2.7'
+gem 'energy-sparks_analytics', github: 'Energy-Sparks/energy-sparks_analytics', branch: '3826-add-new-school-comparison-for-of-gas-used-for-heating-vs-hot-water'
```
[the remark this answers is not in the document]
tbhi: done and merged

app/controllers/comparisons/heating_vs_hot_water_controller.rb
tbhi: made changes for this

ldodds (approved):

PR #3661, opened 2024-05-29
Fix query for heating meters

ldodds:
The HeatingControlService returns a list of gas meters used for heating.

It ignores those meters which have been configured to indicate they are monitoring the kitchen or the hot water system. It should return any gas meter without attributes as well as those explicitly configured as heating only.

The current query has bugs:

- not checking the attribute type
- not checking for active meter attributes

The latter is causing a bug where a meter that was incorrectly configured as hotwater_only, so has an inactive meter attribute, is not being included in the meter breakdown charts on the heating control advice page.

Fixes the query, improve the specs and factories.

### DIFF
--- a/app/services/schools/advice/heating_control_service.rb
+++ b/app/services/schools/advice/heating_control_service.rb
@@ -93,12 +93,12 @@ module Schools
         @school.meters
                .active
                .gas
-               .where.not(id: MeterAttribute.where(meter_id: @school.meters.pluck(:id))
+               .where.not(id: MeterAttribute.where(meter_id: @school.meters.pluck(:id), attribute_type: :function_switch)
                                             .where(
                                               <<-SQL.squish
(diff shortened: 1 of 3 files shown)


app/services/schools/advice/heating_control_service.rb
```diff
@@ -93,12 +93,12 @@ def find_meters
         @school.meters
                .active
                .gas
-               .where.not(id: MeterAttribute.where(meter_id: @school.meters.pluck(:id))
+               .where.not(id: MeterAttribute.where(meter_id: @school.meters.pluck(:id), attribute_type: :function_switch)
                                             .where(
```
tbhi, line 97: perhaps a separate PR but seems like this would be clearer without the heredoc and indentation of the last bracket seems odd (rubocop advising me to put it on the line above)

tbhi (approved):

PR #3663, opened 2024-05-29
Change how chart failures are displayed 

ldodds:
Currently if we get an error trying to display a chart we replace the entire `chart-wrapper` div with an error message. That means the title and controls are removed.

This isn't ideal for the meter breakdown charts where the chart wrapper also includes a select box allowing users to switch between meters. This means an error with one meter removes the ability for the user to look at other meters.

This can happen on the heating control pages where we should a breakdown of gas usage by meter. If we have limited data for a meter, or issues creating heating model the user can't switch to another meter.

This PR changes the standard `chartFailure` function in our javascript to:

- no longer display the chart id to users. This is internal and not necessary
- replace the div containing the Highchart chart, not the entire wrapper. This leaves the title, subtitle and other controls available
- disables the controls for switching the y-axis and navigating forwards and background, but leaves the meter selection box (if available) still active

This means a failed chart, if inserted into the page without a chart component (only used on some admin pages):

Screenshot from 2024-05-29 17-07-10

If inserted via chart component (which is majority of charts now) then failed chart looks like:

Screenshot from 2024-05-29 17-11-58

And finally for a meter breakdown chart:

Screenshot from 2024-05-29 17-06-43

The axis and explore controls are disabled, but the meter selection box is still available. 

Tested in development with `forest-school` gas meters `3262342507` and `44543802` which both have issues.

### DIFF
--- a/app/assets/javascripts/analysis.js
+++ b/app/assets/javascripts/analysis.js
@@ -25,18 +25,19 @@ function updateDatesInSubtitles(subTitleElement, chartData) {
   }
 }
 
-function chartFailure(chart, title) {
-  var $standardErrorMessage = document.getElementById('chart-error').textContent
+function chartFailure(chart) {
+  // the chart
   var $chartDiv = $(chart.renderTo);
+  // the chart wrapper containing the titles, header, various controls
+
   var $chartWrapper = $chartDiv.parents('.chart-wrapper');
+  // disable specific controls if chart fails
+  $chartWrapper.find('.chart-controls .axis-controls :input').attr("disabled", true);
+  $chartWrapper.find('.chart-controls .analysis-controls :input').attr("disabled", true);
 
-  $chartWrapper.addClass('alert alert-warning');
-
-  if (title) {
-    $chartWrapper.html(`<h3>${title}</h3>`)
-  } else {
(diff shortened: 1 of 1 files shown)


urbanwide:
Hi @ldodds,

These changes look good. I noticed a few things when looking at this though (that are not really issues with the PR to be fair):

- when looking at this page: `/schools/forest-school/advice/heating_control/analysis`, on the very top chart and second chart (with multiple meters), there are no units selected when the page initially loads (they are on the last chart though):
image
- Clicking on a unit on the second chart, changes the page focus back to the first chart. Think the js for these charts are interfering with each other somehow?
- the "We do not have enough data to display this chart" message isn't centred vertically, it might look a bit better if it was?

Will let you decide if these need to be done as part of this PR or not :-)

urbanwide (approved):

urbanwide:
Just noticed you've already addressed the main point of my last comment here: https://github.com/Energy-Sparks/energy-sparks/pull/3666, so please ignore that bit.

ldodds:
The alignment was off as the default h3 style has some extra padding. I've removed that on the element itself as didn't want to change across the site.

PR #3672, opened 2024-06-03
Rationalise site fonts

urbanwide:


ldodds (commented):
The changes look OK, but I can see now how you've had to adjust some of the fonts on buttons, etc to match the declarations in the SCSS. Will try locally.

ldodds (approved):

PR #3673, opened 2024-06-04
Rails 7.0 with jsbundling

tbhi:


tbhi:
replace with https://github.com/Energy-Sparks/energy-sparks/pull/3676

PR #3677, opened 2024-06-06
Use new fonts

urbanwide:
* Updated fonts to the latest version of the Storm design.
* Changes the headers to dark blue. Does not change the main text colour as per the new design - firstly because I didn't see this had been changed until the latest design and also because it might be best changed when we switch the site colours for the new ones.
* Has required quite a bit of tidying to allow the changes to be made consistently across the site. This has mostly been to the headers to make them all use h tags where possible.
* Found some places (i.e. some notices, but not all, on the dashboard) where h tags were being used for normal text, where we wouldn't have noticed the inconsistencies before as the fonts were the same (note to Deb to check I have all these!).
* Because the headers are now bigger sizes than previous, some text does look a bit oversized, especially where h tags were used inline rather than as headers. I've mostly tried to rectify this with exception of the scoreboard pages - wanted to see what you thought first @ldodds. The front page and the find out more pages still don't look ideal to me but compared to the original, they do look similar :-)
* I've kept the super big font on the pages that had it (now with a class of f-xl) but it would be good to go to just plain h1 if we could consistently for all headers.

### DIFF
--- a/app/assets/javascripts/common_chart_options.js.erb
+++ b/app/assets/javascripts/common_chart_options.js.erb
@@ -5,7 +5,7 @@ Highcharts.setOptions({
   },
   chart: {
     style: {
-      fontFamily: 'Quicksand'
+      fontFamily: 'Inter'
     }
   }
 });
--- a/app/assets/stylesheets/activities.scss
+++ b/app/assets/stylesheets/activities.scss
@@ -53,7 +53,7 @@ trix-toolbar .button_groups {
 }
 
 .select2-container {
-    width: 100% !important;
+  width: 100% !important;
 }
 
 .select2-selection--multiple:after{
--- a/app/assets/stylesheets/activities.scss
+++ b/app/assets/stylesheets/activities.scss
@@ -49,7 +49,7 @@ trix-toolbar .button_groups {
 
 .select2-container--bootstrap .select2-results__group {
   color: $text;
-  font-size: $f3;
+  font-size: $f4;
 }
 
 .select2-container {
--- a/app/assets/stylesheets/advice_page.scss
+++ b/app/assets/stylesheets/advice_page.scss
@@ -44,23 +44,6 @@
   }
 }
 
-.advice-priority-table {
-  margin-top: 10px !important;
-
-  td h4 {
-    padding-top: 0px;
-    padding-bottom: 0px;
-  }
-
-  td h4 div.trix-content div {
-    font-size: inherit;
-  }
-
-  td h4 a div.trix-content div {
-    font-size: inherit;
-  }
-}
-
 .text-positive {
   color: $green;
 }
--- a/app/assets/stylesheets/analysis.scss
+++ b/app/assets/stylesheets/analysis.scss
@@ -39,7 +39,7 @@ div.analysis-chart, div.usage-chart {
   text-align: center;
   color: $dark-grey;
   line-height: 700px;
-  font-size: $f6;
+  font-size: $f7;
   margin: 0 auto;
   padding-bottom: 25px;
   padding-top: 10px;
--- a/app/assets/stylesheets/analysis.scss
+++ b/app/assets/stylesheets/analysis.scss
@@ -168,5 +168,5 @@ div.popover {
 }
 
 .highcharts-subtitle {
-  font-size: $f5;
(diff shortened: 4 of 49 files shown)


urbanwide:
Nearly there but just noticed transport surveys don't look right 😬

urbanwide:
Just noticed the responsive layout on the masthead isn't right, just looking now

ldodds:
Some things I've noted during initial testing

- the title on the /find-out-more page is now smaller. Think it should be a h1 now?
- on the school group admin page, the text in the green pills for the school status (visible, etc) seems misaligned in the buttons. They display differently to the same set of buttons shown on school dashboard. I think that's because the versions in the page are now Inter?
- on the pupil dashboard the equivalence text is now larger so that the text overlaps with the left arrow on the carousel. Think it should be an h2 instead?
- the font sizes in the table in the priority actions page look off now. The table is using h4 for the text and number styles, which is probably wrong. But the action name now looks smaller than the numbers, which are also now a different colour. Think that needs adjusting. There's some specific styling for that table (`.advice-priority-table`) to try and get the text to display right, so maybe not working now for some reason?

Otherwise I'm generally finding the tweaked font sizes and coloured headings much easier to read. The Inter font might take a bit of getting used to, but it's more readable I think.

I've not tried responsive views yet.

urbanwide:
> * the title on the /find-out-more page is now smaller. Think it should be a h1 now?

Have sorted this, along with the responsive layout on smaller devices to work the same as the existing front page, which I think behaves better (still not ideal but this can be looked at in more detail when these pages / elements are restyled).

> * on the pupil dashboard the equivalence text is now larger so that the text overlaps with the left arrow on the carousel. Think it should be an h2 instead?

The header tags are set in the equivalence admin interface (unless using default equivalences) e.g.:
[https://test.energysparks.uk/admin/equivalence_types/30/edit](https://test.energysparks.uk/admin/equivalence_types/30/edit)
So have styled h1 to look like h2 in the carousel for now. Alternative would be to go through and change the headers in the equivalence types with an afterparty task?

> * the font sizes in the table in the priority actions page look off now. The table is using h4 for the text and number styles, which is probably wrong. But the action name now looks smaller than the numbers, which are also now a different colour. Think that needs adjusting. There's some specific styling for that table (`.advice-priority-table`) to try and get the text to display right, so maybe not working now for some reason?

I removed the h4 tags and initially tried using the f4 sized font (h4 sized) but the text just looked too oversized. So I have scaled down to f5, which is actually the standard text size for the site. Which has since led me on to removing all extra formatting for these tables, which is where it is now. See what you think. It's definitely smaller than before but looks standard I would say.

> Otherwise I'm generally finding the tweaked font sizes and coloured headings much easier to read. The Inter font might take a bit of getting used to, but it's more readable I think.

Yes it's a bit different... pondering whether the general font size could be a bit smaller but I am not a designer :-) I am sure I'll get used to it

> * on the school group admin page, the text in the green pills for the school status (visible, etc) seems misaligned in the buttons. They display differently to the same set of buttons shown on school dashboard. I think that's because the versions in the page are now Inter?

I am about to look at this again - couldn't find a quick fix earlier. Wondering if it's line height issue or just change of font as you say. Will let you know how I get on!

urbanwide:
@ldodds this is how the new fonts look on the School Group admin page for me:
image

This is what it looks like currently on test:
image

I'm now thinking the alignment is better in the new version of these buttons and looks off on the original? What does it look like for you?

urbanwide:
Have added another font size in to bridge the gap between f4 and f5 (this font size is listed in Components -> Text Styles in Sketch H5 - but not in the "New Typography" page) and so we can have a H5 that is bigger than general paragraph text still... (we do use H5). Meant F6 is now the default text size so everything f5 and below needed shifting down one e.g. f5 became f6, f6 became f7 etc. Hope this is ok, just thought it wise to do it now rather than later.

ldodds (approved):
Changes look OK I think. Lets get it on test later today for further feedback

urbanwide:
Just a note to say I built a quick page with just bootstrap css (and non of our extra css) + replacing font with Inter / 18px / 30px line height and there is the same issue:
Screenshot 2024-06-19 at 09 59 48

PR #3683, opened 2024-06-13
Add feature flag for new navigation

urbanwide:
Adds feature flag named 'navigation' with an after party task - this is not switched on for anyone by default for now. 

This also does some other tidying, including removing the no longer used 'header-fix' and also making the top margins of containers (required for a fixed nav bar) switchable (needed for doing this)
Currently if the feature is switched on, no nav is presented. This will be implemented in the next PRs :-)

The easiest way to check this is working is to go to the flipper admin interface:
`http://localhost:3000/admin/flipper/features/navigation`
and enable the navigation feature for the admin group

To enable feature for a specific user on the console:

```
 user = User.find_by(email: 'deb.bassett@energysparks.uk')
 Flipper.enable_actor(:navigation, user)
```

To enable feature for a school group:
```
group = SchoolGroup.find_by(slug: 'name')
Flipper.enable_actor(:navigation, group)
```

To add actors via the admin interface, you must input the flipper id
This can be found by calling flipper_id on the actor. e.g.
`user.flipper_id` 
or
`group.flipper_id`


app/views/layouts/home.html.erb
```diff
@@ -4,16 +4,22 @@
...
+      </div>
+    <% end %>
     <% if flash.any? %>
       <section class="home alerts">
         <%= render 'shared/alerts' %>
       </section>
     <% end %>
     <!-- don't render application container as you do for the other pages -->
-    <%= yield %>
+    <div class="<%= nav_margin_classes %>">
```
urbanwide, line 20: Put this here rather than amend every file that uses the home layout (there are quite a few!)

ldodds (approved):
Looks good. Tested locally (with feature switched off) to check nav/pages looked correct

PR #3688, opened 2024-06-17
Rails 7.0 depreciations

tbhi:


app/views/layouts/action_text/contents/_content.html.erb
ldodds, line 1: Is this not the default behaviour? Or is this to handle something that is changing in Trix?

Although this is useful to add as @urbanwide and I were discussing whether this could be changed, as often we're having to add CSS styling to work around the addition of the extra div.

ldodds (approved):
Looks good. As far as I can see it was just the to_fs change, plus the ActionText customisation?

app/views/layouts/action_text/contents/_content.html.erb
tbhi: this was autocreated by bin/rails action_text:install
accidentally committed here due to mass edit
probably makes sense to do it as a separate commit

PR #3689, opened 2024-06-17
Consolidate redirects

ldodds:
Primary goal here was to remove some old routes for find out more pages as these are no longer used but we are getting noise from crawlers still accessing them.

I've removed the old controller and added a redirect directly in routes.rb.

There were a few other legacy urls that we are maintaining, some of these were done via controller redirects rather than in the routes, so I've consolidated these into the routes file and removed the old controllers.

Added a separate system spec for these so we can more easily decide to retire them in future.


config/routes.rb
```diff
@@ -718,22 +710,31 @@
...
       get 'analysis/:energy/:presentation(/:secondary_presentation)', to: 'analysis#show', as: :analysis_tab
     end
   end
+
+  # LEGACY PATHS
+  # Old paths that (may) have been indexed by crawlers or used in comms/emails that
+  # we want to maintain and redirect.
+
+  # Old 'find out more' pages
+  get '/schools/:name/find_out_more', to: redirect('/schools/%{name}/advice', status: 301)
```
tbhi, line 726: think 301 is the default?

tbhi (approved):

config/routes.rb
ldodds: Thanks. I was getting mixed up with `redirect_to` which defaults to 302 instead so where these redirects where done in controllers previously, we had to set status.

PR #3693, opened 2024-06-18
Fix ordering issues in comparison report specs

ldodds:
We have been getting some flaky tests in the comparison reports. Some of these, e.g. the heating coming on too early, involve sorting schools by name. And then asserting the order.

However the existing School factory creates names like: `test school #{n}`. 

When we get to around 100 schools then the specs can fail because we might have schools: "test school 98", "test school 99" and "test school 100". When sorted by name, they will be returned in the reverse order.

This PR updates the sequence of school name to be: `sequence(:name, 'School AAAAA1')` which produces names using [succ](https://apidock.com/ruby/String/succ). This should produce sensible sorts.

Also:

- fixes a flake in the Alert model spec. Spec asserts a fixed order in the array but the SQL doesn't impose an order
- updates the activity type and intervention type factories to also use the above style of sequence naming. This might sort out some of the other occasional test flakes we get, e.g. with search.

### DIFF
--- a/spec/factories/activity_types_factory.rb
+++ b/spec/factories/activity_types_factory.rb
@@ -1,7 +1,7 @@
 FactoryBot.define do
   factory :activity_type do
     activity_category
-    sequence(:name)                         {|n| "test activity_type name #{n}"}
+    sequence(:name, 'Test Activity Type AAAA1')
     score                                   { 25 }
     active                                  { true }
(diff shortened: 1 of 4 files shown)


urbanwide:
This is really useful, good find!

urbanwide (approved):

PR #3694, opened 2024-06-19
update rails defaults - enable sha256

tbhi:
- migrate pupil password field
  - rename field, copy contents
  - needs environment variables setting
- later PR needed to drop old field and remove EncryptedField
- [x] is the index still needed?
  - don't think it can because using the default non-deterministic encryption
- [x] migration actiontext signed urls
  - https://systemweakness.com/unravelling-signed-url-changes-in-rails-7-e9156da43022?gi=bc753743ca11
- [x] set environment variables on elastic beanstalk environments 
  - `bin/rails db:encryption:init`


app/models/encrypted_field.rb
```diff
@@ -1,6 +1,7 @@
 class EncryptedField
   KEY = ActiveSupport::KeyGenerator.new(
-    Rails.application.secret_key_base
+    Rails.application.secret_key_base,
+    hash_digest_class: OpenSSL::Digest::SHA1
```
ldodds, line 4: If we're explicitly saying to use SHA1 rather than SHA256, why do we need to do a database migration to `User`. I assumed the `User` updates were to re-encrypt the files as default.

lib/tasks/deployment/20240626145318_sha256_action_text_attachment_migrate.rake
```diff
@@ -0,0 +1,33 @@
...
+  desc 'Deployment task: sha256_action_text_attachment_migrate'
+  task sha256_action_text_attachment_migrate: :environment do
+    puts "Running deploy task 'sha256_action_text_attachment_migrate'"
+
+    # https://discuss.rubyonrails.org/t/rails-6-1-7-gives-404-for-actiontext-attachments/80803
+    def refresh_trix(trix)
+      return unless trix.embeds.size.positive?
+
+      trix.body.fragment.find_all('action-text-attachment').each do |node|
+        embed = trix.embeds.find do |attachment|
```
ldodds, line 11: I'm not familiar with the API, but: if there are multiple embeds, shouldn't be checking and processing all of them and not just the first?

ldodds (commented):
Left a couple of questions. Overall looks OK, but wondering how we handle rolling back if there's an issue. The Trix migration looks destructive so wondering how we add a safety net and some extra testing.

app/models/encrypted_field.rb
tbhi: user model change missing from the previous commits to specify new encrypted field

lib/tasks/deployment/20240626145318_sha256_action_text_attachment_migrate.rake
tbhi: updated code to clarify this - it's trying to find the embed matching the action-text-attachment node it is processing

tbhi:
> Left a couple of questions. Overall looks OK, but wondering how we handle rolling back if there's an issue. The Trix migration looks destructive so wondering how we add a safety net and some extra testing.
> 

would manual testing on test and that looks ok be sufficient confidence?

> Also wondering about time required to do the user and trix updates. E.g. should we expect any more downtime than usual during a release that includes this?

yeah, took a couple of minutes locally, would expect slower on production

tbhi:
deploy to test
```
2024/08/13 09:29:52.276041 [INFO] Running script: .platform/hooks/predeploy/99_after_party.sh
2024/08/13 09:32:05.152115 [INFO] Finished running scripts in /var/app/staging/.platform/hooks/predeploy
```

lib/tasks/deployment/20240618161601_migrate_pupil_password.rake
```diff
@@ -0,0 +1,19 @@
+namespace :after_party do
+  desc 'Deployment task: migrate_pupil_password'
+  task migrate_pupil_password: :environment do
+    puts "Running deploy task 'migrate_pupil_password'"
+
+    User.find_each do |user|
```
ldodds, line 6: Could scope this to just the pupil users?

app/models/encrypted_field.rb
ldodds: OK, so this is to ensure the existing EncryptedFields work as before, but User model has changed to use built in support.

ldodds (approved):

PR #3695, opened 2024-06-19
Comparison Report for Last / Current Holiday and Final Week of Term

ldodds:
Adds a report that compares usage in the last holiday for the school (or current holiday if we're in one) with usage in the final week of term.

The report is based on the arbitrary comparison report framework so reuses some templates, but overrides the partials for the individual fuel types to include the name of the last holiday in the results.


config/routes.rb
```diff
@@ -86,6 +86,7 @@
   end
 
   namespace :comparisons do
+    resources :holiday_and_term, only: [:index], concerns: :unlisted
```
tbhi, line 89: sorting?

tbhi (approved):

PR #3696, opened 2024-06-19
Fix application container margin / padding issue

urbanwide:
Adds margin and padding back in to where is was before.

Not a long term solution as these pages will be replaced, just putting back in like for like.


ldodds (approved):
Looks good to me. Tested locally.

urbanwide:
Thanks @ldodds, I'll get it merged

PR #3698, opened 2024-06-19
Update Engie configuration

ldodds:
Database update to update the header and mpan_mprn column for the Engie config. The CSV files have changed format, although are currently still loading as the column is in the same position.

### DIFF
--- a/lib/tasks/deployment/20240619123745_fix_engie_config.rake
+++ b/lib/tasks/deployment/20240619123745_fix_engie_config.rake
@@ -1,0 +1,19 @@
+namespace :after_party do
+  desc 'Deployment task: fix_engie_config'
+  task fix_engie_config: :environment do
+    puts "Running deploy task 'fix_engie_config'"
+
+    config = AmrDataFeedConfig.find_by(identifier: :engie)
+    if config.present?
+      config.update!(
+        mpan_mprn_field: 'Meter Reference Number',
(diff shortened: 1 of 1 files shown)


lib/tasks/deployment/20240619123745_fix_engie_config.rake
```diff
@@ -0,0 +1,19 @@
+namespace :after_party do
+  desc 'Deployment task: fix_engie_config'
+  task fix_engie_config: :environment do
+    puts "Running deploy task 'fix_engie_config'"
+
+    config = AmrDataFeedConfig.find_by(identifier: :engie)
+    if config.present?
+      config.update!(
+        mpan_mprn_field: 'Meter Reference Number',
+        header_example: 'Meter Reference Number,Date,00:00,00:30,01:00,01:30,02:00,02:30,03:00,03:30,04:00,04:30,05:00,05:30,06:00,06:30,07:00,07:30,08:00,08:30,09:00,09:30,10:00,10:30,11:00,11:30,12:00,12:30,13:00,13:30,14:00,14:30,15:00,15:30,16:00,16:30,17:00,17:30,18:00,18:30,19:00,19:30,20:00,20:30,21:00,21:30,22:00,22:30,23:00,23:30'
```
urbanwide, line 10: Should 'Meter Reference' and 'Number' be on the same line here?

ldodds: They are on the same line?

urbanwide (approved):
Looks good but have left one small comment to check before merging...

lib/tasks/deployment/20240619123745_fix_engie_config.rake
ldodds: The field is called "Meter Reference Number"

ldodds: Oh, I see it looks different in the review page, but its just wrapping.

urbanwide: Ahh sorry! Should have realised. All good 👍

PR #3702, opened 2024-06-20
Create new electricity and gas meter breakdown advice pages

ldodds:
This PR adds 2 new advice pages providing a breakdown of electricity and gas usage by meter.

The pages are only accessible by admins currently, as they are controlled by a feature flag.

The initial version of the pages just have slightly updated versions of the meter breakdown table and chart that we currently have in a section of the long term usage pages. Eventually these pages will replace that section. Specifically the new pages have:

- an insights tab with a simplified version of the meter breakdown table which shows % usage by meter
- an analysis tab has the current chart and table, with slightly amended text

Further PRs will build out these pages some more to add some additional analysis and links to sections of other advice pages.

The rest of the changes in this PR are removing the old versions of the meter breakdown pages that are currently only available to admins via the "Old analysis" link on the Manage School menu. These are no longer required. I should have broken up the PR to remove that but go carried away. Those changes involve: 

- removing an older controller and partials
- removing the link to the older pages
- tidying up a few references to the older pages in specs, helpers, redirects, etc.


app/views/schools/advice/_nav.html.erb
ldodds, line 1: erblint changes

app/views/schools/content_reports/show.html.erb
ldodds, line 1: Removing old analysis + erblint

app/views/schools/school_targets/progress/_charts.html.erb
```diff
@@ -13,14 +13,14 @@
 </ul>
 <div class="tab-content" id="targets-overview-tabs">
   <div class="tab-pane fade show active" id="cumulative-progress-<%= fuel_type.to_s %>" role="tabpanel" aria-labelledby="cumulative-progress-tab">
-    <%= render 'schools/analysis/chart_name', school: @school, content: "targeting_and_tracking_weekly_#{fuel_type.to_s}_to_date_cumulative_line".to_sym, mpan_mprn: nil %>
```
ldodds, line 16: We were still referring to an older partial rather than using the chart component so have switched to that as part of removing the partial

config/routes.rb
```diff
@@ -310,8 +312,6 @@
         end
       end
 
-      resources :analysis, controller: :analysis, only: [:index, :show]
```
ldodds, line 313: Remove old analysis controller

config/routes.rb
```diff
@@ -729,6 +729,10 @@
   # Maintain old scoreboard URL
   get '/schools/:name/scoreboard', to: redirect('/scoreboards')
 
+  # Old analysis pages
+  get '/schools/:name/analysis', to: redirect('/schools/%{name}/advice')
```
ldodds, line 733: Redirect the old analysis to new advice pages

tbhi (approved):

app/helpers/advice_page_helper.rb
```diff
@@ -143,6 +143,14 @@ def warm_weather_on_days_status(rating)
     end
   end
 
+  def advice_pages_for_school_and_fuel(advice_pages, school, fuel_type)
+    if school.multiple_meters?(fuel_type) && Flipper.enabled?(:meter_breakdowns, current_user)
```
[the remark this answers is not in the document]
ldodds: No, that only returns pages that are for multiple meters. The check is returning all pages (so where multiple_meters is true or false on the advice page model. Otherwise its returning only those that are single meters.

app/controllers/schools/advice/base_meter_breakdown_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    class BaseMeterBreakdownController < AdviceBaseController
+      before_action :redirect_if_single_meter
+
+      def insights
+        @annual_usage_meter_breakdown = usage_service.annual_usage_meter_breakdown
+        @meters_for_breakdown = sorted_meters_for_breakdown(@annual_usage_meter_breakdown)
+      end
+
+      def analysis
+        @annual_usage_meter_breakdown = usage_service.annual_usage_meter_breakdown
```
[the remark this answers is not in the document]
ldodds: They diverge in the next PR

PR #3710, opened 2024-06-24
Programme progress fix

urbanwide:
New functionality has meant we encourage users to start programmes where a user has already completed activity types that are part of it. Something we'd not thought of is that a user might start a programme having already completed all activity types within a programme, meaning the activity is effectively complete immediately after starting it.

On production, someone had done this - all activities in a programme were marked as complete but the programme wasn't flagged as completed. This is because the programme creator didn't take this into account when starting a programme. This code fixes this. 

When I was working on this I also spotted an issue with the join programmes prompt. It was saying (on my local machine):

"You've recently completed 14 activities that are part of the Beat your baseload programme. Do you want to enroll in the programme?"

When there are only 7 activity types in the programme.

This is because it was counting completed activities and not activity types.

This fixes that now too.

---

Additional work on prompts based on feedback:

- Changes join programme prompt to recognise programmes where all activities are already complete and changes button
- Changes prompt on top of programme page to also recognise this, in addition to programmes that are partially complete
- Also restructure programme view page slightly as the HTML was out. It's slightly better now but not perfect. This can be addressed in the redesign.

### DIFF
--- a/app/views/schools/_prompt_join_programme.html.erb
+++ b/app/views/schools/_prompt_join_programme.html.erb
@@ -1,11 +1,12 @@
 <% if school %>
   <% if programme_type = school.suggested_programme_types.first %>
     <%= component 'info_bar',
-        status: :neutral,
-        style: local_assigns[:style],
-        title: t("schools.prompts.join_programme.message_html", count: programme_type.activity_count, title: programme_type.title),
-        icon: fa_icon('tasks fa-3x'),
-        buttons: { t("common.labels.start") => programme_type_path(programme_type) }
-    %>
+                  status: :neutral,
+                  style: local_assigns[:style],
+                  title: t('schools.prompts.join_programme.message_html',
+                           count: programme_type.activity_types_for_school(school).length,
+                           title: programme_type.title),
+                  icon: fa_icon('tasks fa-3x'),
+                  buttons: { t('common.labels.start') => programme_type_path(programme_type) } %>
   <% end %>
 <% end %>
--- a/app/views/schools/_prompt_join_programme.html.erb
+++ b/app/views/schools/_prompt_join_programme.html.erb
@@ -1,12 +1,23 @@
 <% if school %>
   <% if programme_type = school.suggested_programme_types.first %>
-    <%= component 'info_bar',
-                  status: :neutral,
-                  style: local_assigns[:style],
-                  title: t('schools.prompts.join_programme.message_html',
-                           count: programme_type.activity_types_for_school(school).length,
-                           title: programme_type.title),
-                  icon: fa_icon('tasks fa-3x'),
-                  buttons: { t('common.labels.start') => programme_type_path(programme_type) } %>
+    <% if programme_type.activity_types.length == programme_type.activity_type_count %>
+      <%= component 'info_bar',
+                    status: :neutral,
+                    style: local_assigns[:style],
+                    title: t('schools.prompts.join_programme.completed_message_html',
+                             count: programme_type.bonus_score,
+                             title: programme_type.title),
+                    icon: fa_icon('tasks fa-3x'),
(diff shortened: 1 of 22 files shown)


app/views/schools/_prompt_join_programme.html.erb
```diff
@@ -1,11 +1,12 @@
...
-        status: :neutral,
-        style: local_assigns[:style],
-        title: t("schools.prompts.join_programme.message_html", count: programme_type.activity_count, title: programme_type.title),
-        icon: fa_icon('tasks fa-3x'),
-        buttons: { t("common.labels.start") => programme_type_path(programme_type) }
-    %>
+                  status: :neutral,
+                  style: local_assigns[:style],
+                  title: t('schools.prompts.join_programme.message_html',
+                           count: programme_type.activity_types_for_school(school).length,
```
urbanwide, line 7: Changed this count as it was counting completed activities rather than unique activity_types, which can lead to text such as:

You've recently completed 14 activities that are part of the Beat your baseload programme. Do you want to enroll in the programme?

When there are only 7 activity types in the programme.

lib/tasks/deployment/20240624130949_complete_started_already_completed_programmes.rake
```diff
@@ -0,0 +1,21 @@
+namespace :after_party do
+  desc 'Deployment task: complete_started_already_completed_programmes'
+  task complete_started_already_completed_programmes: :environment do
+    puts "Running deploy task 'complete_started_already_completed_programmes'"
+
+    Programme.started.each do |programme|
+      if programme.all_activities_complete?
+        puts "Setting programme '#{programme.programme_type.title}' to complete for school: #{programme.school.name}"
+        programme.ended_on = programme.started_on
```
ldodds, line 9: Would it not be better to end them on the date this task is run?

ldodds (approved):
This does fix the issue of the odd looking because where there are zero activities to complete and points to score. So I thinks its OK to merge.

But if a user has completed all the activity types for a programme we'll still be suggesting they enrol in that programme, before immediately marking it as complete. I guess that means there's a specific user interaction to join and then immediately complete the programme, but we could also be doing something better there. E.g. if all the activity types are completed, the message to start a programme could say something like you've completed all the activities in this programme. Mark it done to score X bonus points? And then have "Complete" as the button label rather than "Start".

lib/tasks/deployment/20240624130949_complete_started_already_completed_programmes.rake
urbanwide: Set this intentionally as the start date, given the date they started the programme will have been the date they finished it. Happy to change though?

urbanwide:
> This does fix the issue of the odd looking because where there are zero activities to complete and points to score. So I thinks its OK to merge.
> 
> But if a user has completed all the activity types for a programme we'll still be suggesting they enrol in that programme, before immediately marking it as complete. I guess that means there's a specific user interaction to join and then immediately complete the programme, but we could also be doing something better there. E.g. if all the activity types are completed, the message to start a programme could say something like you've completed all the activities in this programme. Mark it done to score X bonus points? And then have "Complete" as the button label rather than "Start".

Yes agreed - sorry was just fixing the issue at hand but you're right, that is a much better solution.

lib/tasks/deployment/20240624130949_complete_started_already_completed_programmes.rake
ldodds: OK, I guess as you're plugging a loop-hole it makes sense. So leave it as it is

urbanwide:
This has had a few changes since you last looked @ldodds, so worth a re-review if you don't mind. See top description for further changes.

ldodds (approved):

PR #3712, opened 2024-06-24
change funder to only be on a school and not a group

tbhi:


ldodds (approved):

ldodds:
Looks good. After this is merged, can you do a PR for the second part (removing old fields, etc) and we'll merge/release that after this has gone to production

PR #3713, opened 2024-06-25
migrate to TimeHelpers

tbhi:
- use TimeHelpers
- after blocks not needed as done automatically - https://www.rubydoc.info/gems/rubocop-rails/2.13.1/RuboCop/Cop/Rails/RedundantTravelBack


urbanwide (approved):
Looks good 👍

PR #3715, opened 2024-06-25
Revise data sharing settings to use enum

ldodds:
Currently we use a `public` flag in the school model to control whether a school's dashboard and analysis is public.

If the flag is public then the school is visible by everyone. If the flag is private then the school dashboard and analysis is not visible to anyone not logged in from that school or from a school in the same group. 

We have some groups that will contain schools that need a further restriction: their analysis will only be visible to their users and not others in the group.

This PR changes the current access control rules from a boolean to an enum. That enum can have values of `:public`, `:within_group`, and `:private`. Its implemented using the new Rails 7 support for postgres enums.

All existing schools that have `private: false` will be updated to have `data_sharing: :within_group`. This preserves the current definition of that behaviour.

The abilities have been redefined to use the new `data_sharing` enum rather than the `public` flag. Other behaviour (some redirects, admin functionality to change the setting) have also been updated to use the enum.

The existing system and model tests for the abilities covered most of the core functionality already but these have been refactored to make them easier to manage. Some additional tests have been added for other user types (e.g. volunteers).

This should be backwards compatible. It does not yet do anything to support schools with `:private` data_sharing, or fix other bugs in the school group dashboard.


app/controllers/schools/public_controller.rb
```diff
@@ -4,13 +4,13 @@ class PublicController < ApplicationController
 
     def create
       authorize! :change_public, @school
-      @school.update!(public: true)
+      @school.update!(public: true, data_sharing: :public)
```
ldodds, line 7: For this PR I've keep the admin buttons for toggling public/private. These will be removed in the next PR. I'm still setting the `public` flag to leave it consistent with previous code until its removed.

app/views/activities/index.html.erb
```diff
@@ -45,7 +45,7 @@
       </tbody>
     </table>
 
-  <%= link_to t('activities.choose_next_activity'), activity_categories_path, class: 'btn btn-success' if can? :suggest_activity, @school %>
```
ldodds, line 48: This is an unrelated change, but this allowed me to remove an otherwise unused ability.

app/views/shared/_school_status_buttons.html.erb
```diff
@@ -1,35 +1,43 @@
...
-    <%= link_to 'Visible', school_visibility_path(school), class: 'badge badge-pill badge-danger', method: :post, data: { confirm: 'Are you sure?' } %>
+    <%= link_to 'Visible', school_visibility_path(school), class: 'badge badge-pill badge-danger', method: :post,
+                                                           data: { confirm: 'Are you sure?' } %>
   <% end %>
 <% end %>
 
 <% if can?(:change_public, school) %>
-  <% if school.public? %>
-    <%= link_to 'Public', school_public_path(school), class: 'badge badge-pill badge-success', method: :delete, data: { confirm: 'Are you sure?' } %>
+  <% if school.data_sharing_public? %>
```
ldodds, line 12: This is the functional change in this file. The rest are ERBlint corrections

app/models/school.rb
```diff
@@ -196,6 +197,13 @@ class ProcessDataError < StandardError; end
   enum funding_status: [:state_school, :private_school]
   enum region: [:north_east, :north_west, :yorkshire_and_the_humber, :east_midlands, :west_midlands, :east_of_england, :london, :south_east, :south_west]
 
+  # Defined as Postgres Enum. Mapping is from ruby values to database values
+  enum :data_sharing, {
+    public: 'public',
+    within_group: 'within_group',
```
urbanwide, line 203: This might just be personal preference, but I wonder if 'group' rather than 'within_group' would be better as all the others are single words. Appreciate that would be a pain to change though

urbanwide (approved):

PR #3717, opened 2024-06-26
Allow admins to configure more limited data sharing options for schools

ldodds:
Next phase of changes to support better access control.

This PR adds the ability for admins to configure the full range of data sharing options. This includes:

- Adding a field to the school configuration form to allow the `data_sharing` option to be configured.
- Changing the public/private button on the school navbar to show the actual data sharing status with appropriate colour coding for the three states, not just a public/private label
- Change the button to link to the school configuration form, rather than acting as a toggle
- Remove the controller that was implementing the toggle behaviour
- Adding the `data_sharing` enum to `SchoolOnboarding` so that the option can be configured for a school prior to them onboarding, as we do with the older public/private flag. The enum has been populated from existing data and the value is copied to the school when created

I've also further reworked the abilities spec to improve coverage a bit more and attempt to organise things a bit better.


spec/models/school_spec.rb
ldodds, line 1: The specs here we partly duplicating what was in `abilities_spec.rb` so I've improved the abilities spec and removed these

app/models/ability.rb
```diff
@@ -229,7 +227,7 @@ def initialize(user)
 
       # Extend default permission to see visible schools with public data to also add permission to
       # view visible schools in the same group that have data sharing set to be 'within_group'
-      can [:show, :show_pupils_dash, :show_management_dash], School, { data_sharing: :within_group, school_group_id: user.school.school_group_id, visible: true }
```
urbanwide, line 232: Assume :show_management_dash was intentionally removed here?

app/models/ability.rb
```diff
@@ -112,9 +112,7 @@ def initialize(user)
     can :live_data, Cad, visible: true, public: true
     can :read, Scoreboard, public: true
 
-    can :read, [FindOutMore, Observation, TransportSurvey, TransportSurvey::Response, ProgrammeType]
-
-    can [:show, :index], SchoolTarget
```
urbanwide, line 117: Just checking it is right to replace a :show, :index for SchoolTarget with just a :read ?

urbanwide (approved):
Looks good, just a couple of questions about ability.rb

app/models/ability.rb
ldodds: Yes `:read` means `:show, :index`. One thing we can decide in tidying this further is to allow short hands like `:read` and `:manage` or just list all permissions.

app/models/ability.rb
ldodds: Yes, this setup apparently gave staff and volunteers access to some features on other schools that school admins didn't have which seemed odd. I could get it to trigger though, and have tests around the expected behaviour so I think we're OK.

app/models/ability.rb
urbanwide: Ahh that's handy to know and much tidier 👍

PR #3718, opened 2024-06-27
removed rb-readline gem

tbhi:


ldodds (approved):
This was possibly used in analytics code at some point in the past.

PR #3719, opened 2024-06-27
Add new Energy Assets/Corona format and time format

ldodds:
Data format has changed for gas data for one of the largest groups.

The PR adds a new data feed config for the format. It uses a row per reading style with separate date and time columns.

The time format is currently rejected by our loader, so have had to tweak that to ensure its accepted

### DIFF
--- a/app/controllers/admin/amr_uploaded_readings_controller.rb
+++ b/app/controllers/admin/amr_uploaded_readings_controller.rb
@@ -31,8 +31,6 @@ module Admin
         render :new
       end
     rescue => e
-      puts e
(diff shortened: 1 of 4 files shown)


app/controllers/admin/amr_uploaded_readings_controller.rb
```diff
@@ -31,6 +31,8 @@ def create
         render :new
       end
     rescue => e
+      puts e
```
tbhi, line 34: probably don't want this?

tbhi (approved):

PR #3729, opened 2024-07-01
Fix current target progress

ldodds:
This PR fixes three issues with the current and expired target reports:

- ensures that we're using the right time series for reporting overall progress
- adds an extra check to whether we display a link to the previous years expired target, as we may have only had limited data last year and so there's nothing to actually display in that report
- revises how we decide which months to report on to ensure we only show up to 12 months of data

As most of the tests already cover the general rendering behaviour in most cases I've just tweaked the data setup to ensure only the right time series is available.

### DIFF
--- a/app/controllers/schools/school_targets/progress_controller.rb
+++ b/app/controllers/schools/school_targets/progress_controller.rb
@@ -78,7 +78,6 @@ module Schools
       def render_school_target_expired
         @progress = @school_target.saved_progress_report_for(@fuel_type)
         @latest_progress = latest_progress
-        final_month_index = @progress.months.find_index(final_month) || 11
         @reporting_months = reporting_months(@progress)
         render :expired
       end
--- a/app/controllers/schools/school_targets/progress_controller.rb
+++ b/app/controllers/schools/school_targets/progress_controller.rb
@@ -93,7 +93,7 @@ module Schools
           final_month = @school_target.target_date.prev_month.beginning_of_month
           @progress.cumulative_performance[final_month]
         else
-          @progress.current_cumulative_performance_versus_synthetic_last_year
+          @progress.current_cumulative_performance
         end
       end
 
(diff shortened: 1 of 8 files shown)


spec/factories/school_targets.rb
```diff
@@ -10,5 +10,30 @@
...
     storage_heaters_progress { {} }
+
+    trait :with_progress_report do
+      transient do
+        fuel_type { :electricity }
+      end
+
+      after(:build) do |target, evaluator|
+        report = {
+          'fuel_type': 'electricity',
```
tbhi, line 21: evaluator.fuel_type?

app/controllers/schools/school_targets/progress_controller.rb
```diff
@@ -80,23 +78,38 @@ def render_school_target_current
...
-          final_month = @school_target.target_date.prev_month.beginning_of_month
           @progress.cumulative_performance[final_month]
         else
-          @progress.current_cumulative_performance_versus_synthetic_last_year
+          @progress.current_cumulative_performance
         end
       end
 
+      # The analytics can return a report with >12 months of data in it.
+      # e.g. if we've been continuining to generate a report for an expired target
```
tbhi, line 96: continuing?

spec/factories/school_targets.rb
```diff
@@ -10,5 +10,30 @@
...
+          'months': ['2024-01-01'],
```
tbhi, line 22: don't think it makes a difference these are quoted

tbhi (approved):

spec/factories/school_targets.rb
ldodds: I was leaving them as string as that's how it'll come out of the database. There's code in the school target model to symbolize them and reformat some values when its converted into an object. So was preserving the database version.

tbhi: then it should be `'months' => ['2024-01-01']`?

PR #3737, opened 2024-07-03
unused?

tbhi:


ldodds:
Have a PR here which covers removing all of the old code. https://github.com/Energy-Sparks/energy-sparks/pull/3749

PR #3739, opened 2024-07-05
Remove legacy code and content management from old analysis pages

ldodds:
Following from https://github.com/Energy-Sparks/energy-sparks/pull/3738 and https://github.com/Energy-Sparks/energy-sparks/pull/3702 we now no longer need to use any of the code in the analytics responsible for producing the old versions of the advice pages.

This means we can:

- remove a step in the regeneration process and the model which it populates (`GenerateAnalysisPages`,`AnalysisPages`)
- remove some unused database and form fields that was managing content associated with those old pages (the `analysis_active` field in `AlertTypeRating` and various fields for title and description in `AlertTypeRatingContentVersions`)

This PR has:

- migrations to remove the old fields and model. This content was never sent for translation so is completely legacy and nothing to remove from Transifex
- removes form fields for managing that content
- remove some additional helpers and simplifies code


app/controllers/analysis_page_finder_controller.rb
```diff
@@ -12,9 +11,15 @@ def show
 
     alert_type = find_advice_page_of_class(analysis_class)
     if alert_type && alert_type.advice_page.present?
-      redirect_to advice_page_path(school, alert_type.advice_page)
+      redirect_to advice_page_path(school, alert_type.advice_page), status: :moved_permanently
```
ldodds, line 14: Added status code to force crawlers to update indexes. These links were only present in the old version of the comparison pags.

PR #3756, opened 2024-07-12
configurable period comparison reports timing out

tbhi:


ldodds (approved):
Looks good. If this boosts performance, then seems like its worth updating the `latest_runs` query on the other views as well? Could you do a follow-up for that?

PR #3758, opened 2024-07-12
Add meter breakdown component

ldodds:
On two of the advice pages (baseload, heating control) we use a partial to customise a chart so that it has an additional meter selection control. The control lists the meters with readings and allows the usage to switch the chart to showing different meters.

As part of working to make more of the charts have meter switching behaviour, I've reworked the existing code into a component. 

Felt cleaner to write a new component that wraps the old with additional behaviour rather than adding more parameters to ChartComponent.


app/components/meter_selection_chart_component.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def chart_descriptions
+    @date_ranges_by_meter.each_with_object({}) do |(mpan_mprn, dates), date_ranges|
+      date_ranges[mpan_mprn] = I18n.t(@chart_subtitle_key,
+        start_date: dates[:start_date].to_fs(:es_short),
+        end_date: dates[:end_date].to_fs(:es_short),
+        meter: dates[:meter].present? ? dates[:meter].name_or_mpan_mprn : mpan_mprn)
+    end
+  end
```
tbhi, line 47: perhaps
```  
  def chart_descriptions
    @date_ranges_by_meter.to_h do |mpan_mprn, dates|
      [mpan_mprn, I18n.t(@chart_subtitle_key,
                         start_date: dates[:start_date].to_fs(:es_short),
                         end_date: dates[:end_date].to_fs(:es_short),
                         meter: dates[:meter]&.name_or_mpan_mprn || mpan_mprn)]
    end
  end
```

tbhi (approved):

PR #3763, opened 2024-07-15
Implement new persistent second level navbar

urbanwide:
For a first look. I've tried to implement as close to the designs as possible, using the new colours. The font sizes are not the same again.

*In order for any of the new navigation to show, the 'navigation' flag must be switched on*

I've created a new menu called 'your school menu' which is a prototype of the menu of the same name shown in the design. It has an asterisk(*)  against the items that are not shown in the design but which I wasn't sure where they'd go otherwise. 'Your school settings' doesn't go anywhere useful right now as it's not implemented yet.
This new menu can be switched on with the 'your_school_menu' feature flag, for users with a school attached to their account. If the flag is not switched on, the old behaviour will show -  it will show the old 'my_school_menu' if a user is not an admin and has a school, otherwise will show the 'our_schools_menu'.

I've tried to keep the menus as they currently are now, but my feeling is that there are too many of them.
I think we're also missing a 'sign up' link or similar which is not in the designs (think it might say 'login' page should also have a signup link within it)

Other caveats:

* Needs tests
* Needs some translations adding where there are new menu items / text.


ldodds:
Initial feedback:

- Was confused initially about not seeing the new nav then realised I needed to enable both the navigation and your_school_menu feature together. From a testing point of view it'd be slightly simpler to be able to enable the entire new navbar for, e.g. admins or specific users. At the moment I have to fully enable both which will make doing some limited review tricky as we'll need to separate enable users on both features
- With the new font sizes, the second level nav text and menu item sizes feels much too small to me. I think it would look better if they were the same size. 
- "Our schools" should be in the top-nav and so persistent for everyone. It's part of the overall site navigation, rather than a user/schools specific navbar
- An issue with the design is that it's still not clear (other than in the breadcrumb) which dashboard you currently are on. The colour coding helps but only once you notice it. We might want to rethink this a bit
- The "Back to dashboard" link when you're elsewhere on the site feels a bit lost/unncessary especially if we keep the breadcrumbs.
- My plan was to try and combine the Manage School and Your school menu into a single menu. But that will need some discussion. This is the change that might impact most users
- For a school group user then "Dashboards" could be "My schools" instead
- I've not tried a user yet with multiple "cluster schools", would like to do away with the "Switch schools" menu entirely.
- For admins, we'll need to add in the toggle buttons. Looks like they could fit into the design for now?

urbanwide:
> * Was confused initially about not seeing the new nav then realised I needed to enable both the navigation and your_school_menu feature together. From a testing point of view it'd be slightly simpler to be able to enable the entire new navbar for, e.g. admins or specific users. At the moment I have to fully enable both which will make doing some limited review tricky as we'll need to separate enable users on both features

See your point - however, this extra feature flag switches on non-final behaviour - it will force the menu to be shown to non-admins too, so that admins can see the menu. However will see what I can do to simplify this.

> * With the new font sizes, the second level nav text and menu item sizes feels much too small to me. I think it would look better if they were the same size.

OK, this smaller font size was as per the design but happy to change it to the same as the top.

> * "Our schools" should be in the top-nav and so persistent for everyone. It's part of the overall site navigation, rather than a user/schools specific navbar

Can change this

> * An issue with the design is that it's still not clear (other than in the breadcrumb) which dashboard you currently are on. 
The colour coding helps but only once you notice it. We might want to rethink this a bit

OK 👍 

> * The "Back to dashboard" link when you're elsewhere on the site feels a bit lost/unncessary especially if we keep the breadcrumbs.

This was in the design but happy to remove 👍 

> * My plan was to try and combine the Manage School and Your school menu into a single menu. But that will need some discussion. This is the change that might impact most users

The only issue with this is we have the concept of 'your school' and 'current school'. The manage school menu is scoped to the current school we are looking at, where as the 'your school' menu appears for the user's own school. Maybe this isn't right.

> * For a school group user then "Dashboards" could be "My schools" instead

Agreed

> * I've not tried a user yet with multiple "cluster schools", would like to do away with the "Switch schools" menu entirely.

Need to look at this too

> * For admins, we'll need to add in the toggle buttons. Looks like they could fit into the design for now?

I'll put them in

Thanks for looking 👍

urbanwide:
I think this is more or less as we discussed @ldodds (plus a large refactor).

I wrote down to "Check pupils have 'Your school' menu". Currently staff, volunteer and pupils currently have permission to view the 'Your/My school menu'. Is this right or did I write the wrong thing down?

The other thing I wanted to check is the way we are determining the current user's school group. The current nav looks to the current user -> school group first, then if not found there, will use the current user -> school -> school group.
This is different to the comparison pages, which do current user -> school -> school_group then if not found, current user -> school_group

I've written two helpers for these at the moment, I imagine we would want to stick to one way of doing it, unless I've missed something anyway.

Going to press on and write some spec next. 

Thanks

ldodds:
> I wrote down to "Check pupils have 'Your school' menu". Currently staff, volunteer and pupils currently have permission to view the 'Your/My school menu'. Is this right or did I write the wrong thing down?

I think that's right yes. I don't think pupils have the Manage School menu though. 

> The other thing I wanted to check is the way we are determining the current user's school group. The current (and new) nav looks to the current user -> school group first, then if not found there, will use the current user -> school -> school group. This is different to the comparison pages, which do current user -> school -> school_group then if not found, current user -> school_group
> 
> I've written two helpers for these at the moment, I imagine we would want to stick to one way of doing it, unless I've missed something anyway.

The ordering has no real meaning here and in practice I don't think it would ever be different.

We're only expecting group admins to be linked directly to a school group. So a normal school user (including cluster users) won't have a school group.

There may be a small number of older users that were originally school admins (possibly with 'cluster schools') that became group admins where there may be a link to a school as well as a group. But it will still end up resolving to a school group.

Normal school users would never be linked to a group.

I'd suggest standardising on checking school group first and then via the school. You could also check to see if they're a group admin. If they are, then use the group. If not, then try via the school.

urbanwide:
There is a lot to review here, sorry! Have done about as much as I can do now @ldodds. I've written quite a bit of spec but I'd like to bring in the spec which tests the contents of some of the more complex menus at some point, I think once we're more clear on what should go where.

Any questions, please ask! I'm away tomorrow to Monday but around next week Tuesday to Thursday if need be. I don't mind. If not, see you in a few weeks!

ldodds (approved):

PR #3766, opened 2024-07-17
Fix setting defaults on school onboarding forms

ldodds:
There's a long standing bug in how the admin school onboarding forms were written.

Currently the populating the school onboarding with the default values from the school group are done in the `edit` action on the configuration controller. So when an admin first visits that form the defaults are copied over and can then be accepted or customised.

The bug is that if they visit the form again later as making other changes, then the group default values are reapplied before displaying the form. So the form state looks like any original selections were never saved as its no longer reflecting the database state. 

This PR fixes that bug so that the school group defaults are applied when the school onboarding is initially created.

### DIFF
--- a/app/models/school_onboarding.rb
+++ b/app/models/school_onboarding.rb
@@ -74,7 +74,7 @@ class SchoolOnboarding < ApplicationRecord
   enum country: School.countries
 
   def populate_default_values(user)
-    self.assign_attributes({
+    assign_attributes({
       uuid: SecureRandom.uuid,
       created_by: user,
       template_calendar: school_group&.default_template_calendar,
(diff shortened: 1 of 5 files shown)


app/models/school_onboarding.rb
```diff
@@ -73,6 +73,20 @@ class SchoolOnboarding < ApplicationRecord
   enum default_chart_preference: [:default, :carbon, :usage, :cost]
   enum country: School.countries
 
+  def populate_default_values(user)
+    self.assign_attributes({
```
tbhi, line 77: self not needed?

tbhi (approved):

PR #3767, opened 2024-07-17
Rework how the SingleReadConverter drops rows

ldodds:
We currently support an option on `AmrDataFeedConfig` called `missing_readings_limit`. If set then any day that has more missing half hourly readings that this limit will automatically be rejected by the data loader. It allows us to silently reject incomplete data without generating lots of warnings for the admins.

While this was previously used on some older configs, it's not currently used. When set it's applied by the `DataFeedValidator`.

Separately the `SingleReadConverter` uses a hard coded `BLANK_THRESHOLD` when it is converting "row per reading" formats into the standard format. The value of that constant has always been 1.

For a new data supplier we need to be more tolerant of some missing half-hourly readings. Currently they are sending us data in a "row per reading" format that regularly has several half-hourly readings that are missing. This means all the data is getting rejected by `SingleReadConverter` which isn't ideal.

This PR changes `SingleReadConverter` so that the threshold is instead configurable using the `missing_readings_limit` parameter on the config.

For backwards compatibility, I've approached this by:

- ensuring the `DataFeedValidator` does not apply `missing_readings_limit` to "row per reading" formats. As the `SingleReadConverter` will handle rejecting any rows. It doesn't make sense for the validator to do this as we haven't get gathered up the individual rows into an array for these format
- changing `SingleReadingConverter` to have access to the `AmrDataFeedConfig` and then read the threshold from there, via a new `.blank_threshold` method which will default to 1 for "row_per_reading" formats if `missing_readings_limit` is not configured
- revised the rest of the behaviour and tests accordingly

I decided not to just set a default value of `missing_readings_limit` of 1 for all "row per reading" formats for the moment. Or set a validation/callback to ensure that any new similar formats have a default. This is because there's another issue to be resolved later around how we are interpreting timestamps in some of those row per reading files. Resolving that may mean that we don't need a default value, so wanted to handle that in the code rather than having to do several database changes.

There's a bit of reworking of some of the specs, but haven't done a full tidy up at the moment.

### DIFF
--- a/app/models/amr_data_feed_config.rb
+++ b/app/models/amr_data_feed_config.rb
@@ -110,6 +110,6 @@ class AmrDataFeedConfig < ApplicationRecord
   # formats can produce days with missing readings due to handling of 23:30-00:00 half-hour.
   def blank_threshold
     return nil unless row_per_reading?
-    return missing_readings_limit || BLANK_THRESHOLD
+    missing_readings_limit || BLANK_THRESHOLD
   end
 end
--- a/app/models/amr_data_feed_config.rb
+++ b/app/models/amr_data_feed_config.rb
@@ -57,6 +57,8 @@ class AmrDataFeedConfig < ApplicationRecord
   validates :identifier, :description, uniqueness: true
   validates_presence_of :identifier, :description
 
+  BLANK_THRESHOLD = 1
+
   def map_of_fields_to_indexes(header = nil)
     this_header = header || header_example
     header_array = this_header.split(',')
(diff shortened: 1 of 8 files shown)


app/models/amr_data_feed_config.rb
```diff
@@ -96,4 +98,18 @@ def local_bucket_path
...
+  # Only applicable to row_per_reading formats
+  #
+  # To preserve current loader behaviour this returns either the value of missing_readings_limit, to
+  # allow that to be configured, or a default of 1
+  #
+  # This can later be replaced with missing_reading_limit, but need to resolve how row_per_reading
+  # formats can produce days with missing readings due to handling of 23:30-00:00 half-hour.
+  def blank_threshold
+    return nil unless row_per_reading?
+    return missing_readings_limit || BLANK_THRESHOLD
```
tbhi, line 113: return not needed

app/services/amr/data_feed_validator.rb
```diff
@@ -82,5 +82,12 @@ def partial_row?(row)
       return true unless row.count > index_of_last_reading_field
       row[index_of_first_reading_field..index_of_last_reading_field].count(&:blank?) > @config.missing_readings_limit
     end
+
+    # Don't apply the missing_readings_limit to reject partial rows for row_per_reading format
+    # That is done in SingleReadConverter for those configs
+    def should_reject_rows?
+      return false if @config.row_per_reading?
+      return @config.missing_readings_limit.present?
```
tbhi, line 90: return not needed

app/services/amr/single_read_converter.rb
```diff
@@ -2,13 +2,12 @@ module Amr
   class SingleReadConverter
     class InvalidTimeStringError < StandardError; end
 
-    BLANK_THRESHOLD = 1
-
     # @param Array single_reading_array an array of readings
     # @param boolean indexed whether the array should be interpreted in HH order, rather than via timestamp.
```
tbhi, line 6: need to update the comment

tbhi (approved):

PR #3770, opened 2024-07-18
Differentiate between real SMETS2 meters and other meters available from n3rgy

tbhi:
- [x] migrate Meter.dcc_meter to enum - no, smets2, other
- [x] update uses of Meter.dcc_meter so smets2 or other works like true did
- [x] update DccChecker to set dcc_meter to smets2 if deviceId is formatted like 00-1D-24-02-01-1A-DF-41
- [x] update http://localhost:3000/schools/[school/]meters/[id] to show meter type
- [x] update emails sent to admins, where we want to indicate whether a meter is a SMETS-2 meter in an extra column (via DccCheckerJob)


app/controllers/schools/meters_controller.rb
```diff
@@ -77,17 +82,12 @@ def destroy
...
       return unless can?(:view_dcc_data, @school)
-      service = Meters::N3rgyMeteringService.new(@meter)
-      @known_to_n3rgy = service.available?
-      if @known_to_n3rgy && @meter.dcc_meter
-        @n3rgy_status = service.status
-        @n3rgy_consent_confirmed = service.consented?
-        @available_cache_range = service.available_data if @n3rgy_consent_confirmed
-      end
+
+      @n3rgy = Meters::N3rgyMeteringService.new(@meter)
```
ldodds, line 90: The original code was trying to reduce number of API calls. I think moving this into the view means we're doing more calls again?. Original: one for available, then consented? if known to n3rgy, and then we only get cache if consented. New view code calls available, consented? (which also calls available again), then consented? again (which calls available again) before getting the cache.

Suggest either adding the vars back here or some caching in the service.

spec/models/meter_spec.rb
```diff
@@ -100,10 +100,10 @@
 
     context 'when finding meters for review' do
       let(:meter_review) { create(:meter_review) }
-      let(:electricity_meter_reviewed) { create(:electricity_meter, dcc_meter: true, meter_review: meter_review, mpan_mprn: 1_234_567_890_111) }
-      let(:electricity_meter_not_reviewed) { create(:electricity_meter, dcc_meter: true, mpan_mprn: 1_234_567_890_222, meter_review_id: nil, consent_granted: false) }
-      let(:electricity_meter_not_dcc) { create(:electricity_meter, dcc_meter: false, mpan_mprn: 1_234_567_890_333) }
-      let(:electricity_meter_consent_granted_already) { create(:electricity_meter, dcc_meter: true, consent_granted: true, mpan_mprn: 1_234_567_890_444) }
+      let(:electricity_meter_reviewed) { create(:electricity_meter, dcc_meter: :smets2, meter_review: meter_review, mpan_mprn: 1_234_567_890_111) }
```
ldodds, line 103: Small thing but should there be a test that its finding meters marked as `:other` as we're still only checking the 2 original states. Same possibly true for some of the other specs too?

db/migrate/20240716160020_change_meter_dcc_meter_to_enum.rb
ldodds, line 1: Nice way to migrate the existing values. I'm wondering whether we should also recheck all the `:smets` meters to reconfirm if they are `:smets` or `:other`. The regular dcc checker rake task will recheck everything every 7 days, but might be good to force a recheck of the `:smets` meters when we release this? E.g. could add a variant of the rake task in `dcc_meter_checker.rake` that does a recheck of previously checked meters only?

ldodds (approved):
Approving, but see comment about number of API calls on the meter show page

app/controllers/schools/meters_controller.rb
tbhi: caching added

spec/models/meter_spec.rb
tbhi: added some other meters

db/migrate/20240716160020_change_meter_dcc_meter_to_enum.rb
tbhi: added after party task to run dcc_meter_checker

ldodds: Looks like you've added an after party task to run the existing dcc checker on deploy. I'd prefer not making remote API calls as part of the release process. 

Running the existing rake tasks isn't going to help recheck and possibly reclassify the meters as I suggested here. What I was thinking was a task that could be run manually post deploy and which would select only those meters that are currently marked as `:smets`, regardless of when they were last checked. The goal being to correctly re-classify those meters that are currently `:smets` but should be `:other`. There would be no need to send an email in that scenario as we're just updating their classification.

Just means we start with correct statuses everywhere, rather than waiting up to 7 days for a recheck.

tbhi: OK, have removed the after party task and amended meters:check_for_dcc so DCC_METER_CHECKER_ALL_SMETS2 can be set to just update just those meters - would that work?

ldodds: Yes that works. Could you run this after doing the release today?

PR #3773, opened 2024-07-18
Fix alert email previews

ldodds:
We have a built-in controller and template for previewing alert emails as they would have looked when sent to users.

This is out of date so wasn't providing a correct view, e.g. some parameters that drive conditional sections were not being set and the formatting was incorrect. 

I've updated the code to link to the standard Rails mailer preview, removed the custom code, and updated our preview class to load data for a saved `Email`.

I needed this to help with debugging a potential problem reported by a user (which turned out to be unrelated). But these changes should help with debugging any future problems


spec/mailers/previews/alert_mailer_preview.rb
```diff
@@ -1,11 +1,32 @@
...
   end
 
   private
 
   def locale
     @params['locale'].present? ? @params['locale'] : 'en'
   end
+
+  def include_target_prompt_in_email?(school)
+    return Targets::SchoolTargetService.targets_enabled?(school) && Targets::SchoolTargetService.new(school).enough_data?
```
tbhi, line 17: don't need return

tbhi (approved):

PR #3780, opened 2024-07-24
Intervention type admin report

tbhi:


app/views/admin/reports/intervention_types/show.html.erb
```diff
@@ -0,0 +1,127 @@
...
+          <th>School</th>
+          <th>Images?</th>
+        </tr>
+      </thead>
+      <tbody>
+        <% @observations.each do |observation| %>
+          <tr>
+            <td><%= nice_dates(observation.created_at) %></td>
+            <td><%= nice_dates(observation.at) %></td>
+            <td><%= link_to observation.school.name, school_timeline_path(observation.school) %></td>
```
ldodds, line 84: This should be a link to the recorded intervention rather than the timeline.

ldodds (approved):

app/views/admin/reports/intervention_types/show.html.erb
tbhi: updated

PR #3781, opened 2024-07-26
 Log who records an activity or intervention 

tbhi:


spec/services/activity_creator_spec.rb
```diff
@@ -1,44 +1,46 @@
...
+    activity = build(:activity, activity_type:)
+    described_class.new(activity, user).process
     expect(activity).to be_persisted
   end
 
   it 'creates an observation for the activity with the points' do
-    activity = build(:activity, activity_type: activity_type)
-    ActivityCreator.new(activity).process
+    activity = build(:activity, activity_type:)
+    described_class.new(activity, user).process
```
ldodds, line 24: Add some assertions that the user is populated, maybe also in the system specs?

ldodds (approved):

PR #3792, opened 2024-07-31
let nginx serve static files?

tbhi:
- using nginx which is configured by default on elastic beanstalk to serve /assets and /public from /var/app/current
- also needs cdn change to cache /assets/*


tbhi:
deployed to test

PR #3799, opened 2024-08-02
Update out of hours usage pages with meter breakdown

ldodds:
Updated the gas and electricity out of hours pages to allow two of the charts (the initial pie chart, and the day of the week breakdown) to be switched to show usage across different meters.

To summarise the code changes:

* Adds `Charts::MeterSelection` a new service to help with building the list of meters to be used to populate a select component for the `MeterSelectionChartComponent`. This is a separate service to allow it to be used in other contexts.
* Refactors the `MeterSelectionChartComponent` to depend on `MeterSelection` and extends it to support  `header` and `footer` slots like `ChartComponent`
* Add a fallback to  `MeterSelectionChartComponent`  so it just delegates to `ChartComponent` if there's a single meter
* Updates the analytics to include changes that allow the day type breakdown charts to work with all meters. See https://github.com/Energy-Sparks/energy-sparks_analytics/pull/708
* Some refactoring of other uses of `MeterSelectionChartComponent` so that they also use `MeterSelection` to try and consolidate some of the code to use the new service
* Updates to the translations to make some of the existing subtitles more dynamic (to include meter names) and remove some duplicates (e.g. 'Whole school').


tbhi (approved):
was trying it on /schools/chase-lane-primary-and-nursery-school/advice/electricity_out_of_hours/analysis and noticed the swimming pool chart is empty but might be a local data issue

PR #3818, opened 2024-08-15
Allow schools to complete programmes each academic year

tbhi:
- have assume not changing functionality via admin pages
- [x] bonus points independent of academic year
- [x] repeat
- [x] restart
- [x] keep prompt to start programme functionality
  - looks like programme_types.prompt.message_html - already looking at last academic year so seems like it should just work


ldodds (approved):
Looks good to me :+1:

PR #3820, opened 2024-08-16
 simplify rubocop config 

tbhi:


tbhi:
replace with https://github.com/Energy-Sparks/energy-sparks/pull/4456

PR #3821, opened 2024-08-16
Skip decommissioned meters in meter breakdown table

ldodds:
Fixes an issue in new meter breakdown table.

The table loops through all meters for a fuel type. But some schools have decommissioned/old meters which are active but are not included in the usage breakdown calculations because the meter data is >1 year old.

Added a fix to the template that is breaking to skip over any meters not included in the analysis.

### DIFF
--- a/app/controllers/schools/advice/base_long_term_controller.rb
+++ b/app/controllers/schools/advice/base_long_term_controller.rb
@@ -18,11 +18,6 @@ module Schools
         @estimated_savings_vs_exemplar = usage_service.estimated_savings(versus: :exemplar_school)
         @estimated_savings_vs_benchmark = usage_service.estimated_savings(versus: :benchmark_school)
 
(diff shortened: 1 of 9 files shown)


tbhi (approved):
makes sense but wonder if it should be filtered before the template

ldodds:
> makes sense but wonder if it should be filtered before the template

Yes, that's probably better. I've reworked this fix as follows:

- retained the original list of meters, as this is all meters that may feature on these pages
- created a reduced list of meters that matches those in the usage breakdown. This is now used to populate that table so templates are just iterating over the reduced list
- refactored the controller slightly to share some code between insights/analysis
- removed the meter breakdown chart/table from the long term usage pages as the new Meter Comparison pages are now live

Mind having another looks?

tbhi (approved):

PR #3823, opened 2024-08-19
Fix interpolation arguments in updated translations

ldodds:
The rework I did to support meter breakdowns involved renaming some interpolation arguments. I missed a couple in the Welsh translations which are causing errors.

Have reviewed all the pages and only these are affected.


tbhi (approved):
would be good to have something automatically check for that

PR #3845, opened 2024-08-28
 Reimplement generation of "average" school 

tbhi:
- added rake task run calculation and save to yaml (code copied from https://github.com/Energy-Sparks/energy-sparks_analytics/blob/master/script/modelling/generate_average_school.rb)

rollout plan
- run `rake school:calculate_average` to generate new data
- commit generated yaml file with https://github.com/Energy-Sparks/energy-sparks_analytics/pull/714 to use it


tbhi:
> Do you have a sense of the impact of adding the generation and upload of the YAML to s3 adds to the overall timing and memory overhead of the daily regeneration process? That process is already quite slow, so am wary about it running for longer even on a day when we're just dumping these files, unless we do it over a weekend perhaps? I'm wondering about whether it would be better to just have a separate job to do this.

It ran ok on test but did take longer than the production run that day.  I will look at AggregateSchoolService.

ldodds (approved):

PR #3846, opened 2024-08-29
New pupil analysis pages

ldodds:
Reworks the pupil analysis pages to refresh the design.

As part of reimplementing this I have:

- written a new `PupilExploreComponent` to handle adding the box-outs. These all have a coloured background, icon and then a link to either another page, or one of two different types of chart pages
- created new versions of the templates in `views/pupils/analysis/new` to make it easier to remove older pages and partials in future
- written specs for all fuel type and chart type variants of the page. It looks like we didn't have any system specs covering this previously
- updated the IconComponent to support using different FontAwesome icon sets, e.g. the brand (`fab`) icons as used on the pupil analysis pages to display an icon for storage heaters
- fixed the page titles on the new dashboard
- corrected the links to the pupil analysis in the DashboardEquivalenceComponent

It's a big PR but mostly templates. Felt it was easier to do this in a block rather than do pages individually.

The different pages for each fuel type don't have a consistent set of charts across all of them. E.g. the gas page doesn't have some charts that are on the electricity pages and the storage heater pages have very limited options.

I've not refactored the controller that drives the pages, or the underlying configuration in the analytics which defines which charts are shown. We should revisit this later as the inconsistent approach to fuel types and configuration keys makes things messy.

### DIFF
--- a/app/components/dashboard_equivalences_component.rb
+++ b/app/components/dashboard_equivalences_component.rb
@@ -51,6 +51,12 @@ class DashboardEquivalencesComponent < ApplicationComponent
     I18n.t("common.#{equivalence_content.equivalence_type.meter_type}").downcase
   end
 
+  def analysis_category(equivalence_content)
+    fuel_type = equivalence_content.equivalence_type.meter_type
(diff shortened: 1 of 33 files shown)


urbanwide (approved):
Looks good - nice to see all the new components in use!

PR #3859, opened 2024-09-03
Create new advice page index

ldodds:
Builds the new advice page index for the redesign.

Ended up being a larger PR than planned as I wasn't sure initially how to approach the changes, as this design varies from the wireframes in a few ways so needed to prototype it first.

Summary of changes:

- use feature flag to bring in new version of show page for the advice page index
- rework page so that it uses the dashboard layout and has coloured title bar similar to adult dash
- create `DashboardChartsComponent` to show the tabbed set of charts previously on adult dashboard
- incorporates the `AdvicePageListComponent` with some styling and colour updates
- create `AdviceRemindersComponent` for the prompts for what to do next
- create `ComparisonOverviewComponent` to provide links to comparison tool plus summary of gas and electricity comparisons
- adds link to school targets page from advice nav bar so that set of reports is better integrated
- some minors tweaks to styling (mostly margins/padding) to other components to improve display in the new design 

I've added tests for all the new components and a proper set of system specs for the advice page index which didn't have good coverage before.

Uses the pale/light colours from the redesign to avoid making the advice page list too garish. But these might need tweaking again.


spec/components/comparison_overview_component_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+                            reading_start_date: Time.zone.yesterday)
+      end
+
+      it { expect(component.can_benchmark_gas?).to be(false) }
+    end
+
+    context 'with electricity enough data' do
+      let(:school) do
+        create(:school,
+                            :with_basic_configuration_single_meter_and_tariffs,
```
tbhi, line 80: odd indentation?

tbhi (approved):

PR #3867, opened 2024-09-05
remove old pupil password field and associated encrypted field

tbhi:


tbhi:
rolled into https://github.com/Energy-Sparks/energy-sparks/pull/3921

PR #3889, opened 2024-09-10
Schools with limited gas data have failing out of hours analysis

ldodds:
This change checks to see if the total out of hours usage is zero before allowing a user to access the out of hours usage page.

This cropped up for several schools where there seems to be a data issue as we have around 8 months of all zero data. The meters should have been deactivated, but while they were active the out of hours page triggered an error which isn't ideal. 

In another case a school only had data over the summer period where gas data might legitimately have been zero. E.g. if they had electric heating. In this case we needed a message for the user.

This PR checks to see if the total kwh usage for the reported period is zero before allowing access to the insights/analysis pages. If there's zero usage then there's no much to show them. Works similarly to the special case for hot water, etc.

### DIFF
--- a/app/controllers/schools/advice/base_out_of_hours_controller.rb
+++ b/app/controllers/schools/advice/base_out_of_hours_controller.rb
@@ -24,7 +24,7 @@ module Schools
         render 'not_enough_data' and return unless @analysable.enough_data?
         @analysis_dates = analysis_dates
         @annual_usage_breakdown = annual_usage_breakdown_service.usage_breakdown
-        render 'schools/advice/out_of_hours/no_usage' and return if @annual_usage_breakdown&.out_of_hours.kwh.zero?
+        render 'schools/advice/out_of_hours/no_usage' and return if @annual_usage_breakdown&.out_of_hours&.kwh&.zero?
       end
 
       def aggregate_meter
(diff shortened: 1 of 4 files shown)


app/controllers/schools/advice/base_out_of_hours_controller.rb
```diff
@@ -15,12 +14,19 @@ def insights
...
         @annual_usage_breakdown = annual_usage_breakdown_service.usage_breakdown
         @holiday_usage = holiday_usage_calculation_service.school_holiday_calendar_comparison
-        @analysis_dates = analysis_dates
         @meter_selection = Charts::MeterSelection.new(@school, aggregate_school, advice_page_fuel_type, date_window: 363)
       end
 
       private
 
+      def check_can_run_analysis
+        @analysable = create_analysable
```
tbhi, line 23: would it be the same/clearer to call super for the first 2 lines?

tbhi (approved):

app/controllers/schools/advice/base_out_of_hours_controller.rb
ldodds: The `render ... and return` doesn't seem to work correctly when calling super as it returns to the overriding method.

PR #3895, opened 2024-09-11
Revise school target comparison reports

tbhi:


ldodds:
Sorry, card could have been clearer. The change wasn't to the column order, needs some further changes.

E.g. looking at: https://energysparks.uk/comparisons/electricity_targets?country=wales&funder=&school_types%5B%5D=primary&school_types%5B%5D=secondary&school_types%5B%5D=special&school_types%5B%5D=infant&school_types%5B%5D=junior&school_types%5B%5D=middle&school_types%5B%5D=mixed_primary_and_secondary&search=country

What we should be showing for Maindee school is their actual target, which is 5% (see https://energysparks.uk/schools/maindee-primary-school/school_targets/389). And then their current progress against that which is +10.2%

We don't need the % above/below last year as that's the change in their general annual production. Can also remove the last year kWh column as well I think. 

So the report basically just becomes a summary of what targets schools have set, how they measure against that, those % expressed in kWh terms and the dates.

tbhi:
Have revised 
image

ldodds (approved):

PR #3901, opened 2024-09-12
Add public display version of pupil analysis

ldodds:
Some MATs have public signage (e.g. large LCD screens) connected to computers or a digital signage product.

We've been asked to provide some views of our analysis for display on this type of signage.

This PR provides an initial prototype of this new feature. It provides:

- a stable URL pattern for requesting a random equivalence for a fuel type (`electricity` or `gas`)
- a stable URL pattern for requesting either an out of hours usage chart, or a comparison of use over last 2 weeks
- an index page that provides links to the above to help school admins setup the links

The equivalences and charts uses a layout that strips out our branding and some CSS tweaks to bump up font sizes for the equivalences. This is to allow the pages to be embedded in a digital signage system.

The idea is to test this implementation with a MAT then decide how to roll-out further.

### DIFF
--- a/app/components/equivalence_component/equivalence_component.scss
+++ b/app/components/equivalence_component/equivalence_component.scss
@@ -41,11 +41,10 @@
 // so more readable from a distance
 .equivalence-component.public-displays.horizontal {
   .trix-content div, div {
+    @include font(f1);
     font-family: $header-font-family;
     font-weight: $header-font-weight;
-    font-size: $f1;
     color: $dark-blue;
-    line-height: 58px;
     padding-bottom: 10px;
   }
 }
--- a/app/components/equivalence_component/equivalence_component.scss
+++ b/app/components/equivalence_component/equivalence_component.scss
@@ -35,3 +35,17 @@
     }
   }
 }
+
+// Used for public display version of equivalence
+// Force the text to match the header height and colours
+// so more readable from a distance
+.equivalence-component.public-displays.horizontal {
+  .trix-content div, div {
+    font-family: $header-font-family;
+    font-weight: $header-font-weight;
+    font-size: $f1;
+    color: $dark-blue;
+    line-height: 58px;
+    padding-bottom: 10px;
+  }
+}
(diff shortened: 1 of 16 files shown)


app/models/chart_data_values.rb
```diff
@@ -133,6 +133,8 @@ def colour_lookup
   end
 
   def work_out_best_colour(data_type)
+    puts 'HERE'
```
urbanwide, line 136: Think this isn't intended?

app/assets/stylesheets/screens.scss
urbanwide, line 1: This is empty, should it be here?

app/models/chart_data_values.rb
ldodds: oops!

app/assets/stylesheets/screens.scss
ldodds: Yes, I never ended up using it, so will remove. Did have some styles in here temporarily.

app/components/equivalence_component/equivalence_component.scss
```diff
@@ -35,3 +35,17 @@
...
 }
+
+// Used for public display version of equivalence
+// Force the text to match the header height and colours
+// so more readable from a distance
+.equivalence-component.public-displays.horizontal {
+  .trix-content div, div {
+    font-family: $header-font-family;
+    font-weight: $header-font-weight;
+    font-size: $f1;
```
urbanwide, line 46: Just for future reference, you could have done the following, instead of specifying font-size and line-height:
`@include font(f1);
`

app/controllers/pupils/public_displays_controller.rb
```diff
@@ -0,0 +1,146 @@
...
+      @chart = find_chart(@fuel_type, @chart_type)
+    end
+
+    private
+
+    def set_fuel_type
+      @fuel_type = params.require(:fuel_type).to_sym
+    end
+
+    def check_fuel_type
```
urbanwide, line 53: I think this is something that could be a method in the school model, at some point anyway.

spec/system/pupils/public_displays_spec.rb
```diff
@@ -0,0 +1,202 @@
+require 'rails_helper'
+
+describe 'Pupil dashboard' do
```
urbanwide, line 3: Should this be "pupil dashboard public displays" or similar rather than "Pupil dashboard"?

urbanwide (approved):
Looks good 👍

PR #3909, opened 2024-09-13
add a time period to the table on the solar PV analysis page

tbhi:
- add table introduction text similar to other advice pages on the solar PV analysis page
- make usage of how_have_we_analysed_your_data modal more explicit so see failures when used without dates set
- fix advice pages where it is used without setting


app/controllers/schools/advice/storage_heaters_controller.rb
```diff
@@ -6,12 +6,12 @@ class StorageHeatersController < AdviceBaseController
       before_action :set_annual_usage_breakdown, only: [:insights, :analysis]
       before_action :set_usage_categories, only: [:insights, :analysis]
       before_action :set_heating_thermostatic_analysis, only: [:insights, :analysis]
+      before_action :set_no_how_we_analysed_modal, only: [:insights]
+      before_action :set_analysis_dates, only: [:analysis]
```
ldodds, line 10: I think its a bug that the modal links aren't shown on the storage heater pages.

ldodds (changes requested):
Could also add a spec to assert the dates shown on the page?

re: changes to add `set_no_how_we_analysed_modal` I'm not sure these are all correct. And the general rule should be: if the page has a table on it, then that table should have a modal that explains the date ranges and other factors used in the calculation. The current conditional disables the modal markup but not the link to it that might be present in 1+ tables on the page.

I think there's probably a better approach where by we add the modal if a link to it has been added to the page, rather than disabling it from the top-down.

app/controllers/schools/advice/storage_heaters_controller.rb
tbhi: added links to storage heater pages

ldodds (approved):
Looks good now, thanks.

ldodds:
> Any preference?

Mild preference for the version on 2 lines. But happy either way. We should standardise this, but that can be done separately.

I've held off on adding a generic table component as often the HTML markup is clearer, but maybe we need a "table with footer" type component that just accepts the thead/tbody slots and a footer to ensure tables are styled consistently.
 
> Have modified to make the view partial set an instance variable which doesn't feel entirely clean but prevents needing to set the variable from the controller.

Think that's a bit nicer.

I did wonder whether we could use some sharing of variables between components to have the FootnoteModal component that adds the actual modaly markup only render if a FootnoteModal::Link has been added to the page already. Would need an id to indicate the type of link and modal, ensure that we could have multiple modals on a page. Again, don't have to do here but might tidy things up further.

PR #3918, opened 2024-09-17
Add number of staff and admin users to the Funders allocation report

tbhi:


app/services/schools/funder_allocation_report_service.rb
```diff
@@ -73,6 +74,7 @@ def csv
             country(school),
             school.number_of_pupils,
             school.percentage_free_school_meals,
+            school.staff.count + school.users.school_admin.count,
```
urbanwide, line 77: I just had a look in the school model to check my understanding of this and found these methods.

Wonder if you need to be looking at using all_adult_school_users here? I don't know the answer but wondered if cluster users need including? It's not mentioned in the original ticket but might be assumed

```
  def school_admin
    users.school_admin
  end

  def staff
    users.staff
  end

  def all_school_admins
    school_admin + cluster_users
  end

  def all_adult_school_users
    (all_school_admins + staff).uniq
  end
```

tbhi: would seem like the right method

urbanwide: It does sound right, is it worth checking with Claudia just in case though?

urbanwide (approved):

PR #3920, opened 2024-09-18
add Find out more link for AlertEnergyAnnualVersusBenchmark

tbhi:


ldodds (approved):
Think this is fine as a special case for the moment. Could extract the check for the class name into a class method on AlertType, e.g. AlertType.link_to_index?

PR #3922, opened 2024-09-20
 Downloadable engaged schools report to cover previous school year 

tbhi:
- since only have user last_sign_in_at also including users that have logged on in the current year
  - also for last visit column


app/controllers/admin/reports/engaged_schools_controller.rb
```diff
@@ -2,23 +2,26 @@ module Admin
...
 
         respond_to do |format|
           format.html do
             @visible_schools = School.visible.count
             @percentage = percentage_engaged
           end
           format.csv do
-            send_data csv_report(@engaged_schools), filename: "engaged-schools-report-#{Time.zone.now.iso8601}".parameterize + '.csv'
+            send_data csv_report(@engaged_schools),
+                      filename: "engaged-schools-report-#{Time.zone.now.iso8601}".parameterize + '.csv'
```
ldodds, line 16: might be helpful to adjust the name for last years report?

ldodds (approved):

PR #3928, opened 2024-09-25
Add lock/unlock buttons to school users page

tbhi:
- also tried to clarify some of the user


ldodds (approved):
Looks good. It would make sense for school admins to be able to resend confirmations

PR #3931, opened 2024-09-26
Handle HEAD requests to all advice pages

ldodds:
Introduces some extra routing to intercept HEAD requests to the insights, analysis and learn_more actions for all advice pages. The requests are handled by a generic action that returns OK and no content. The majority of the before_actions in the controller will be skipped as well, so its should be a quick response.

Rails default behaviour is to respond to a HEAD request by treating it as a GET request, running the same action and then throwing away the rendered content to just return the default headers. While it does allow actions to set additional headers, its wasteful in most cases. And particularly on our advice pages.

Seems like the email scanners that are doing these requests during our weekly alert email sending, don't always wait for the response anyway, as nginx is logging a 499 request for many of them.

We could apply a more general approach across the site, but these are the most performance sensitive routes so have just prioritised these for now.

The fix is simple but has some downsides if we were to use it elsewhere. The main one is that the headers aren't always identical. E.g. there's no ETag header. So browsers/caches/CDNs inspecting that those via a HEAD request may expire their content. However its more normal when using ETags to do a Conditional GET not a HEAD request, so browsers should be fine. In any case, Rails handling of ETags is also not ideal, so think this approach is OK here.

### DIFF
--- a/app/controllers/schools/advice/advice_base_controller.rb
+++ b/app/controllers/schools/advice/advice_base_controller.rb
@@ -39,7 +39,7 @@ module Schools
       # Generic action used to respond to HEAD requests
       # See routes.rb for routing
       def handle_head
-        return head(:ok)
+        head(:ok)
       end
 
       def show
--- a/app/controllers/schools/advice/advice_base_controller.rb
+++ b/app/controllers/schools/advice/advice_base_controller.rb
@@ -36,6 +36,12 @@ module Schools
         end
       end
 
+      # Generic action used to respond to HEAD requests
+      # See routes.rb for routing
+      def handle_head
+        return head(:ok)
+      end
+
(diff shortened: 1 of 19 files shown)


app/controllers/schools/advice/advice_base_controller.rb
```diff
@@ -36,6 +36,12 @@ class AdviceBaseController < ApplicationController
         end
       end
 
+      # Generic action used to respond to HEAD requests
+      # See routes.rb for routing
+      def handle_head
+        return head(:ok)
```
urbanwide, line 42: Don't think you need the `return` here?

urbanwide (approved):
Good solution - see tiny suggestion 👍🏻

PR #3934, opened 2024-09-27
 increase priority of puma 

tbhi:


tbhi:
avoided the extra restart with renice

ldodds (approved):

PR #3944, opened 2024-10-02
Style buttons

urbanwide:
On the whole I have tried to:
* Where there are view, edit and delete buttons together (in pubic facing interfaces), view has become primary, edit secondary and delete danger. In other (admin) cases, there has only been edit and delete buttons, in which case they are edit primary and delete danger - not sure if this is right as we need to get the balance right between consistency and using primary for buttons that are most important. Some admin interfaces have been left as is for now, so this is not always the case. I've always tried to add btn-danger to the delete buttons where I've found them though.
* I have left in some btn-warning (yellow) in for now in admin interfaces.
* I have styled btn-success as per the teal button in the design. I have moved away from using btn-success in the codebase for now though as teal buttons really in the design are for on dark blue background. Also have moved away from btn-info too as it didn't seem to be used in the right places. See what you think
* The outline buttons currently have a white background. I think this looks better than transparent. Some buttons on the original design have transparent backgrounds though so not sure which option to go with.
* I have removed the used of btn-outline-dark and associated friends (btn-outline-dark rounded-pill font-weight-bold) previously used to style our outline button. I have replaced it (and friends) with btn-default for now, as didn't want to loose the placeholder. btn-default has has the same style as the base btn.
 
image

### Button groups
I have changed the default (accidental) styling for button groups as we had soo many button groups and I didn't want to remove the grouping all together. So btn-group still groups buttons together but if you want to make it look like they are in a panel (i.e. not rounded in between buttons), you now need to add the panel class too. I thought it would be better to choose where we wanted to do this rather than accidentally have this all over the site.

image

### Radio buttons, checkboxes, range and progress

image

* also moved away from the old greys in this PR
* also fixes the admin/data_sources toggler


ldodds:
Comments:

- there are some buttons, e.g. on /campaigns/find-out-more, which are still use a gradient background. Should these not be a plain teal or similar colour? This might be something that will be fixed in the redesign of those pages, but not sure whether these gradients are elsewhere (couldn't see them in button designs in your images)

Screenshot from 2024-10-14 14-54-08

- /pricing now has two different styles of button, should be the same

- the meter management page needs to have some colours on the buttons, but could be done in separate PR. probably needs some discussion but "Edit" might be the primary, "Delete" as danger, "Activate" and "Deactivate" as warning, report as "Info", rest as secondary. There's some extra spacing in the button groups below?

Screenshot from 2024-10-14 15-09-50

Otherwise looks good to me. There's some of the admin forms/pages where we could add some colours to the buttons but that's not urgent. I've reviewed all the user facing pages and forms and think they're ok

ldodds (approved):
Looks good apart from notes in comments. Again suggest merging tomorrow ready for Thursday release

urbanwide:
This is the button colouring as outlined above. My thoughts are it might be better for the report button to be secondary for now, but see what you think.

The extra spacing around the buttons I think should be done in a separate PR
image

ldodds:
I think that looks fine tbh, but can make the report button secondary if you want. I was going with the bootstrap meanings. makes sense to do the spacing in another PR. I also think the buttons could be better organised but might want some feedback

urbanwide:
@ldodds do you want a last check of this before merge? Assuming we are ok with this to be merged to master without going on test first anyway?

PR #3945, opened 2024-10-03
set open_close_breakdown on all meters to fix daily regeneration error

tbhi:


ldodds:
Ignore this, looks like the problem is still occuring. Will look at the school again and review

ldodds (approved):
Good to merge after updating tags

PR #3946, opened 2024-10-03
Use to not cc

ldodds:
Fixes an issue with revised way of sending alerts (one email per school).

The code was written to `cc` everyone. The rails default mail delivery used in development accepted that without any issues.

However the mailgun ruby library [expects a `to` address without checking if it exists](https://github.com/jorgemanrubia/mailgun_rails/blob/master/lib/mailgun_rails/deliverer.rb#L53C1-L53C41). So we got an error.

The [Mailgun API requires a `to` address](https://documentation.mailgun.com/docs/mailgun/api-reference/openapi-final/tag/Messages/#tag/Messages/operation/httpapi.(*apiHandler).handler-fm-19) as well, so even if the client was more robust there would still be an issue.

So this changes code to use To.

I thought SMTP allowed at least one of TO, CC, BCC but we're limited by Mailgun here.

### DIFF
--- a/app/mailers/alert_mailer.rb
+++ b/app/mailers/alert_mailer.rb
@@ -19,11 +19,7 @@ class AlertMailer < LocaleMailer
     @alert_content = self.class.create_content(@events)
     @title = @school.name
 
-    email = if @batch_email
-              make_bootstrap_mail(to: @email_addresses)
-            else
-              make_bootstrap_mail(to: @email_addresses)
-            end
+    email = make_bootstrap_mail(to: @email_addresses)
     add_mg_email_tag(email, 'alerts')
   end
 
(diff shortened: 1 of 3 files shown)


app/mailers/alert_mailer.rb
```diff
@@ -20,7 +20,7 @@ def alert_email
     @title = @school.name
 
     email = if @batch_email
```
urbanwide, line 22: I don't think you need this if statement anymore as the outcome is the same for both?

spec/mailers/alert_mailer_spec.rb
```diff
@@ -66,8 +66,7 @@
 
       it 'send to right cc addresses' do
```
urbanwide, line 67: Spec title needs to not say cc now

urbanwide (approved):
See comments...

app/mailers/alert_mailer.rb
ldodds: yes, d'oh

PR #3947, opened 2024-10-03
Optimise SQL on schools page

ldodds:
We're going to be rebuilding this page, but noticed that this page is still slow on live site which I think is due to the SQL queries. 

Looks like we're loading full school models even though we only need names and links.

So have tweaked the controller and templates to just pull in what is required for now

### DIFF
--- a/app/views/schools/index.html.erb
+++ b/app/views/schools/index.html.erb
@@ -11,7 +11,7 @@
 <div class="col-md-12">
 
   <p><%= t('schools.index.introduction_1') %>.</p>
-  <p><%= t('schools.index.introduction_2', count: School.visible.count) %>.</p>
+  <p><%= t('schools.index.introduction_2', count: @schools.length) %>.</p>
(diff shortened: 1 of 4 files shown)


app/views/schools/index.html.erb
```diff
@@ -11,7 +11,7 @@
 <div class="col-md-12">
 
   <p><%= t('schools.index.introduction_1') %>.</p>
-  <p><%= t('schools.index.introduction_2', count: @schools.count) %>.</p>
+  <p><%= t('schools.index.introduction_2', count: School.visible.count) %>.</p>
```
urbanwide, line 14: Would @schools.length be more efficient here as you already have @schools rather than doing another query? I might be wrong... I suppose a count would be quite minimal

urbanwide (approved):
See comment...

app/views/schools/index.html.erb
ldodds: Yes, that would be better.

PR #3948, opened 2024-10-03
Bump regeneration queue size

ldodds:
Old machine had 2 vCPUs. Regenerations with >2 jobs running didn't complete any quicker as the server was CPU bound.

The new machine has 4 vCPUs. Should be possible to increase queue size to 3 without maxing out all CPUs, whilst still reducing time in queue. 

Suggest we try this first to max most use of available resources before adjusting niceness levels so we can see the impacts of both. May not need to adjust nice level if we have spare CPU capacity during regenerations?


tbhi:
I think you could still get contention with heavy web traffic as puma looks to scale workers by number of CPUs but will be interesting to see if this makes a difference to regeneration time with normal web traffic.

tbhi (approved):

PR #3949, opened 2024-10-03
Tidy up adult dashboard code

ldodds:
Removes the old version of the adult dashboard.

This ended up being quite large as:

- there were a number of components to update, as well as older partials and specs to remove. The existing system specs have largely been replaced by the specs for the components that now make up the pages
- in simplifying the alerts component I needed to update a number of advice page partials

There's also a couple of fixes:

- the links should no longer be showing on the alerts included on advice page
- the school timeline page has been reverted to a version based on the old design to allow activities and actions to be easily edited and deleted

As well as simplifying the code it reduces the amount of data being loaded by the schools controller. Much of that is already being done in the new components.


app/components/alerts_component/alerts_component.html.erb
```diff
@@ -1,63 +1,36 @@
...
+          <% prompt.with_link do %>
+            <%= link_to t('schools.show.find_out_more'),
+                        advice_page_path(school,
+                                         alert_content.alert.advice_page,
+                                         alert_content.alert.alert_type.advice_page_tab_for_link_to,
+                                         anchor: alert_content.alert.alert_type.link_to_section) %>
+          <% end if show_links %>
+        <% end %>
+        <%= alert_content.send(content_field) %>
     <% end %>
```
urbanwide, line 34: There's something odd going on with the indentation here (alert_content is too far out)?

urbanwide (approved):

urbanwide:
Looks like a big clear out! See small comment ^

app/components/alerts_component/alerts_component.html.erb
ldodds: That's where erblint wants it to be so parameters of all the functions align

PR #3955, opened 2024-10-07
ignore alerts with low savings

tbhi:


ldodds:
Looks good. Before merging down can we release this branch to test. I'll refresh the test db so we can compare results.

ldodds (approved):

PR #3962, opened 2024-10-08
add workaround for n3rgy issue with other dcc meters

tbhi:


ldodds (approved):
Would be nice to have a spec for this. There's already some specs for some of the other date manipulation in this class

PR #3964, opened 2024-10-08
show all schools in the my school menu

tbhi:


ldodds:
I think you will need some additional changes, due to how the wider logic works.

If an existing school admin is added to a new school, then we call `add_cluster_school` via `ClusterAdminsController` which adds that new school. But the users existing school won't be in the list initially.

If a user switches school then we call `add_cluster_school` via `SchoolSwitcherController` which adds their existing school to the list, then updates their direct association.

So the list of schools returned via `.cluster_schools` won't necessarily be complete. Also need to make sure that the users original school is in there.

It works like this because the menu was originally "Switch schools" and showed the list of schools you could switch to, rather than all those that you are an admin of.

(Longer term the whole auth mechanism here needs updating)

Also need some revisions to existing specs around this.

ldodds (approved):

PR #3980, opened 2024-10-15
Add additional crawler exclusions

ldodds:
Plus a couple of gaps in the robots.txt. In particular disallows crawling of the comparison page results which can be slow.

These were previously all blocked, but the rework of the urls meant that this is currently permitted.


tbhi (approved):
I believe trailing wildcards are ignored - https://developers.google.com/search/docs/crawling-indexing/robots/robots_txt

PR #3992, opened 2024-10-18
Don't create amr import job for folder name, tidy ups

tbhi:


ldodds:
Have you tested this against S3 as well as updating the specs? There's some files in the dev amr-data bucket already.

ldodds (approved):
OK to merge assuming you've done a quick test against S3.

PR #3995, opened 2024-10-18
Fix chart usage controls

urbanwide:
This is working as expected locally but not in production, so a bit odd. In production, `$fuel-name` was remaining as `electricity` with each loop iteration, where as locally it is the different fuel name (apart from electric, which is getting changed to electricity, as expected)

Also realised that site fuel colours were being used rather than the old chart ones for this, so have changed that too.

### DIFF
--- a/app/assets/stylesheets/colours.scss.erb
+++ b/app/assets/stylesheets/colours.scss.erb
@@ -28,10 +28,3 @@
   }
 }
 
-@function rename($key, $old-name, $new-name) {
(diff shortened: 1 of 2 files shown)


ldodds (approved):
As the test server runs as a "production" environment according to Rails you could release this PR to test prior to merging to further test the fix.

urbanwide:
Thanks, will do

urbanwide:
Ok that worked 👍 Will merge, thanks

PR #4002, opened 2024-10-21
Implement new footer design

urbanwide:
The top image is the implemented version, the bottom is the actual design:

image

Differences I see as follows, aside from some text looking slightly bolder than the design and padding variations (looked better when it came to implementation):
'Contact us' is called 'Contact' (as per current top nav)
'Careers' is called 'Jobs' (as per current top nav)
'Workshops' is called 'Educational workshops' (as per current top nav)
'Privacy & Cookies Policy' is split into two separate links (as per current top nav)
Do we want to change these links to the design, and if so, do we want to change those in the top nav too?

Have added social media links based off what I think we have.

Perhaps this can go on test for an easier look when we're ready

### DIFF
--- a/app/assets/stylesheets/bootstrap-overrides.scss
+++ b/app/assets/stylesheets/bootstrap-overrides.scss
@@ -10,6 +10,7 @@ $link-decoration: underline;
 $tooltip-bg: $grey-very-dark;
 
 $btn-font-family: $header-font-family;
+$btn-font-weight: $font-weight-semibold;
 $btn-border-radius: 10rem;
 $btn-border-radius-sm: $btn-border-radius;
 $btn-border-radius-lg: $btn-border-radius;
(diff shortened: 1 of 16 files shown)


ldodds:
Looks good to me. I think I prefer our current labelling rather than the design, so suggest we keep that throughout. 

Navigation is always tricky especially as we're not sure what users might be looking for at the top-level. However, I think we can rejig the original proposal to reflect the things we want to prioritise.

## Quick links:

- use roughly similar ordering as the top nav, so activities first then actions.
- remove "Compare schools" which isn't as commonly used as the "Schools" page, Not sure we need both as these are quick links to common pages 
- add link to Schools, before Scoreboards

## Services

I think I'd change this to add Training below workshops, and replace the "For...X" links with a "Find out more" link that links to /find-out-more and a "Book Demo" which jumps to the book demo link.

## Other links

This is a bit of random mix. I'd replace it with Jobs, Case Studies, Blog and School statistics.

ldodds (approved):

PR #4021, opened 2024-10-29
Set minimum solar self-consumption of zero

tbhi:


ldodds:
Would be good to try this on test for a few days to see if there's any impact on the time to aggregate schools with a number of solar panels.

tbhi:
released with https://github.com/Energy-Sparks/energy-sparks/pull/4039

PR #4027, opened 2024-10-31
Tune autovacuum for alerts table

ldodds:
Reconfigure the autovacuum options for the alerts table.

- reduces cost to zero to ensure autovacuum runs as quickly as possible when triggered
- change the configuration for the `vacuum` and `analyze` operations so rather than being triggered based on a % of table changed + number of records, they now use a fixed threshold

Similar configuration for the readings table removed some of the database waits and ensuring vacuuming was running more frequently. Currently the table is only getting vacuumed after the monthly (and now weekly) cleanup.

Thresholds are estimated, so may need tweaking in future.

### DIFF
--- a/app/models/alert.rb
+++ b/app/models/alert.rb
@@ -39,7 +39,14 @@
 #  fk_rails_...  (comparison_report_id => comparison_reports.id)
 #  fk_rails_...  (school_id => schools.id) ON DELETE => cascade
 #
-
+# Postgres autovacuum specific settings:
+# See: https://www.postgresql.org/docs/current/runtime-config-autovacuum.html
+# Applied using: ALTER TABLE amr_data_feed_readings SET (X = n)
+# autovacuum_vacuum_cost_delay = 0
+# autovacuum_analyze_scale_factor = 0
+# autovacuum_analyze_threshold = 50000
+# autovacuum_vacuum_scale_factor = 0
+# autovacuum_vacuum_threshold = 50000
 class Alert < ApplicationRecord
   include EnumReportingPeriod
   include AlertTypeWithComparisonReport
--- a/lib/tasks/deployment/20241031141954_tune_autovacuum_on_alerts.rake
+++ b/lib/tasks/deployment/20241031141954_tune_autovacuum_on_alerts.rake
@@ -1,0 +1,24 @@
+namespace :after_party do
+  desc 'Deployment task: tune_autovacuum_on_alerts'
+  task tune_autovacuum_on_alerts: :environment do
+    puts "Running deploy task 'tune_autovacuum_on_alerts'"
+
+    # Reduce cost delay to complete autovacuum processes (vacuum, analyse) as quickly as possible
+    ActiveRecord::Base.connection.execute("ALTER TABLE alerts SET (autovacuum_vacuum_cost_delay = 0)")
+
+    #Set scale factor to zero, rather than 0.2 so that we trigger the autovacuum process to run ANALYZE based on number of
+    #records changed, not a % of the table.
+    ActiveRecord::Base.connection.execute("ALTER TABLE alerts SET (autovacuum_analyze_scale_factor = 0)")
+    ActiveRecord::Base.connection.execute("ALTER TABLE alerts SET (autovacuum_analyze_threshold = 50000)")
+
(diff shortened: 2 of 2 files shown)


tbhi (commented):
so before it was 50 + 20%? I've 1,700,071 rows locally so that would seemed to be about 34k

seems like there are separate thresholds on insert - will those be triggered on daily regeneration? (looks like 1000 + 20%)

ldodds:
> so before it was 50 + 20%? I've 1,700,071 rows locally so that would seemed to be about 34k
> 
> seems like there are separate thresholds on insert - will those be triggered on daily regeneration? (looks like 1000 + 20%)

There's some calculation in the doc I shared. There are 697372 rows in alerts after the clean-up currently.

Based on that and the current defaults are:

for vacuum after inserts:
```
(autovacuum_vacuum_insert_scale_factor  * rows) + autovacuum_vacuum_insert_threshold = (0.2 * 697512)+1000 = 140,502
```

This is currently far higher than what we insert each day.

for vacuum after updates and deletes 
```
(autovacuum_vacuum_scale_factor  * rows) + autovacuum_vacuum_threshold = (0.1 * 697512) + 50 = 69,801.2
 ```

for analyse after inserts, updates, deletes:
```
(autovacuum_analyze_scale_factor * rows) + autovacuum_analyze_threshold = (0.05 * 697512) + 50 = 34,925.6
```

We're currently inserting around `num_data_enabled_schools * number_of_alert_types * number of regenerations` alert records each day.

- number of alert types relevant to school will depend on fuel types, holidays, and how many custom period reports we have in the system, etc. But between 30-50
- number of regeneration is normally 1 but might be several if admins are doing work

So roughly 50,000 records a day. I used that as the basis for the thresholds.

tbhi:
so we'd expect it to run every few days from the daily regeneration inserts and then also after the weekly delete?  would we want to avoid triggering it during the day?

tbhi:
is the weekly delete the only time we'd expect vacuuming to process dead tuples?

also seems like reducing autovacuum_vacuum_cost_delay will make vacuuming have a greater impact on server performance since it pauses less so not sure we'd want that?

ldodds:
>so we'd expect it to run every few days from the daily regeneration inserts and then also after the weekly delete? 

Yes. Perhaps daily if we increase number of schools and/or custom reports.

I want it to run more predictably that currently. Due to the current thresholds and the previous monthly clean-up it may have run at the start of the month, but not when the table grew large enough that the % threshold wasn't met.
 
> would we want to avoid triggering it during the day?

We have no control over that. Other than disabling autovacuum on the table and running it manually. But I don't think it's that much of an issue. Vacuums should be running fairly regularly.

> is the weekly delete the only time we'd expect vacuuming to process dead tuples?

Yes. Workers are triggered by the changes, so should have at the time of changes or very shortly after if there's multiple changes happening.

> also seems like reducing autovacuum_vacuum_cost_delay will make vacuuming have a greater impact on server performance since it pauses less so not sure we'd want that?

It will but it will also guarantee that once the vacuum starts it will complete. We've already got `autovacuum_vacuum_cost_delay` set to zero on the two largest tables (by row count, `amr_data_feed_readings`, `amr_validated_readings`). 

The default settings and cost threshold mean that a vacuum can start but then end up being killed if it reaches cost threshold. So it may never successfully complete. This was happening with the other tables leading to recurring "vacuum wait" times but no successful runs.

The alternative to setting this to zero is increasing the cost threshold. Not sure how to judge that, so I could revert this change and see if we get vacuum waits?

Let me know, so we can make any final tweaks and get this released.

tbhi:
So what's the advantage of having the vacuum run more often?

Wonder if we'd get the most benefit running the vacuum manually after the delete - which sounds like it would be more often than previously, at a low usage time and when it's in a position to reallocate space.

ldodds:
> So what's the advantage of having the vacuum run more often?

Vacuum does a variety of things, identifying free pages is one of them.

https://www.cybertec-postgresql.com/en/tuning-autovacuum-postgresql/

We also want the analyse running regularly (which it should do with a smaller table size) to ensure query plans are up to date.

I just feel it would be better for it to be running more frequently.

> Wonder if we'd get the most benefit running the vacuum manually after the delete - which sounds like it would be more often than previously, at a low usage time and when it's in a position to reallocate space.

I guess we could do that. 

btw, vacuum doesn't reallocate space, it updates the free space map (amongst other things).

PR #4028, opened 2024-11-01
Manual vacuum after deletion

ldodds:
Alternative to https://github.com/Energy-Sparks/energy-sparks/pull/4027

Run a manual vacuum of the alerts table after we have deleted content.

- may trigger second vacuum if autovacuum has already kicked in
- as a manual vacuum this will run equivalent to `autovacuum_vacuum_cost_delay` so will run until completion
- relies on postgres default vacuum settings to run vacuum and/or analyse at other times

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,8 +27,8 @@ gem 'pg'
 gem 'scenic'
 
 # Dashboard analytics
-gem 'energy-sparks_analytics', github: 'Energy-Sparks/energy-sparks_analytics', tag: '6.0.16'
-# gem 'energy-sparks_analytics', path: '../energy-sparks_analytics'
+#gem 'energy-sparks_analytics', github: 'Energy-Sparks/energy-sparks_analytics', tag: '6.0.16'
+gem 'energy-sparks_analytics', path: '../energy-sparks_analytics'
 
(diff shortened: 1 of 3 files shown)


tbhi:
should we do the toast table too?

ldodds:
> should we do the toast table too?

Default behaviour is to vacuum the toast table too. You can see that if you run a vacuum with the verbose setting.

ldodds:
There's a large number of other tables that are touched by the delete (see associations in `ContentGenerationRun` and `AlertGenerationRun`). We could extend this to all those tables. 

Longer term I think we need to tune postgres for our "append-mostly" behaviour for a lot of these tables.

tbhi (approved):
would make sense doing affected tables

seems like postgres would have difficultly knowing when best to run the vacuum so wonder if delete and vacuum is the best pattern

ldodds:
> would make sense doing affected tables

Added
 
> seems like postgres would have difficultly knowing when best to run the vacuum so wonder if delete and vacuum is the best pattern

For these bulk deletions probably yes. For the reading tables we were doing manual vacuums too for a while, but ultimately tuning the autovacuum proved better as it caught all activity (e.g. manual uploads, manual regenerations, etc across the day) without having to add manual vacuums everywhere.

PR #4029, opened 2024-11-01
Build new schools page

ldodds:
Screenshot from 2024-11-01 17-50-20

Implement new schools page. 

- Rework map and introductory text layout
- Adjust map so that it no longer recentres after a load. Avoids an unnecessary jump in the map position
- Tweak the clustering at different zoom levels
- Remove popup content from GeoJSON served to build map layer. Speeds up the map loading as we don't have to load and process all school configurations. Use school id for feature id
- Add new route, controller and template for serving popup content
- Adjust `onEachFeature` to dynamically load popup content
- System specs to check all behaviour. Provides additional tests for the search component behaviour as well as its integration with the controller.

Some of the spec files are just changes to reduce dependency on navigating via the schools page to test school specific functionality. Seems unnecessary to keep testing that navigation in other specs.

I've not ended up changing the JS much to reduce use of JQuery. The map javascript is using JS fetch and promises rather than JQuery functionality.


app/controllers/schools_controller.rb
```diff
@@ -116,10 +117,11 @@ def destroy
 
   def set_search_scope
     @tab = SchoolSearchComponent.sanitize_tab(search_params.fetch(:scope).to_sym)
+    @schools = current_user_admin? ? School.active : School.visible
```
urbanwide, line 120: Would it be beneficial to put a notice for admins saying that this is happening?
'Note for admins: admins can also see schools marked as invisible.'

app/controllers/schools_controller.rb
```diff
@@ -116,10 +117,11 @@ def destroy
...
     @scope = if @tab == :schools
-               current_user_admin? ? School.active : School.visible
+               @schools
              else
-               SchoolGroup.all
+               SchoolGroup.with_visible_schools
```
urbanwide, line 124: Would it make sense to do the same filtering here for admins?

app/controllers/map_controller.rb
```diff
@@ -2,8 +2,14 @@ class MapController < ApplicationController
   skip_before_action :authenticate_user!
 
   def index
-    schools = School.includes(:configuration).visible.order(:name)
-    schools = schools.where(school_group_id: params[:school_group_id]).order(:name) if params[:school_group_id]
+    schools = School.visible
```
urbanwide, line 5: If the schools controller is showing different schools for admin users (as in not visible), should the map too?

app/views/map/popup.html.erb
```diff
@@ -0,0 +1,21 @@
+<p class="small">
+  <%= link_to @school.name, school_path(@school) %>
+</p>
+<p class="small">
+  <%= I18n.t('maps.school_features.school_type') %>: <%= @school.school_type.humanize %>
+</p>
+<p class="small">
+  <%= I18n.t('maps.school_features.fuel_types') %>:
+  <% if @school.has_electricity? %>
+    &nbsp;<i class='fas fa-bolt'></i>
```
urbanwide, line 10: Why not use the icon component here?
`<%= render IconComponent.new(fuel_type: fuel_type) %>`

spec/system/map_spec.rb
```diff
@@ -9,23 +9,40 @@
...
   end
 
-  it 'provides JSON for one group' do
-    get map_path(school_group_id: school_group_2.id, format: :json)
-    json = JSON.parse(response.body)
+  context 'when requesting popup' do
+    let!(:school_1) { create(:school, :with_fuel_configuration, number_of_pupils: 100) }
+
+    it 'serves popup content' do
+      visit map_popup_path(id: school_1.id)
```
urbanwide, line 34: Should these `visit` and `get` statements go in a before statement as they are test setup?

urbanwide (commented):
See inline comments...

ldodds:
As discussed will ask for user feedback and then refactor if needed to adjust the totals.

PR #4037, opened 2024-11-04
adjust log level rather than changing $stdout

tbhi:


app/services/schools/school_metrics_generator_service.rb
```diff
@@ -59,14 +59,11 @@ def suppress_output(&block)
     end
 
     def _suppress_output
-      original_stdout = $stdout.clone
-      original_stderr = $stderr.clone
-      $stderr.reopen File.new('/dev/null', 'w')
-      $stdout.reopen File.new('/dev/null', 'w')
+      original_level = @logger.level
```
ldodds, line 62: If we're no longer throwing away $stdout, is there any reason to also reduce the log level from `INFO` to `WARN` in prod?

ldodds (approved):
Avoids the error but not sure whether the suppress output is doing much now. I assume the analytics logger is already running at INFO.

app/services/schools/school_metrics_generator_service.rb
tbhi: only for less output

PR #4047, opened 2024-11-07
error in log suppression - remove as isn't really needed

tbhi:


ldodds (approved):
Didn't seem to have any majort impact on test server, so suggest we merge this to resolve the original stdout error.

PR #4049, opened 2024-11-08
Update email header and footer to match new designs

tbhi:
mostly changed header
image
and footer
image

possibly more to do?


app/views/shared/footer/_second.html.erb
```diff
@@ -39,3 +39,24 @@
     </div>
   </div>
 </div>
+<% if Flipper.enabled?(:download_footer_icons) %>
```
ldodds, line 42: not sure we should have this tbh, its not saving much time and won't help with adding other icons to emails, we're going to need to add a set of fuel type and other icons.

app/views/layouts/mailer.html.erb
```diff
@@ -15,10 +15,9 @@
         <div class="col-lg-1"></div>
         <div class="col-lg-10">
           <div class="row mt-2">
-            <div class="col-12 bg-black">
+            <div class="col-12 dark">
               <%= link_to root_url do %>
-                <%= image_tag "nav-brand-transparent-#{I18n.locale}.png", class: 'ax-center', width: '300px',
-                                                                          height: '300px' %>
```
ldodds, line 21: From what I can see its more common to have header/footer in emails to be centered. The layout doesn't have to match the application, just the colours. Can you recentre the logo and make the footer content centered.

ldodds (changes requested):

ldodds:
@urbanwide reviewing this I noticed that the button styling is still different (using bootstrap defaults). See, e.g. the alert email preview. Can you add this to your list of things to sort out for the rest of the redesign?

ldodds:
Further improvements to the footer, e.g. to add charity number, can come later as I also need to look at the Mailchimp setup so will try and align across the different tools.

urbanwide:
Yes no probs, I’ll add it as a card

app/views/shared/footer/_second.html.erb
ldodds: Was thinking about this overnight. Agree it would be helpful to have the icons easily available. How about this as an alternative:

- update the icon component preview to include a section with all of the icons we need to use in emails, so those for the footer and those you'll need in the alert email changes. The section can have a title and notes about how to download the icons and how that works
- integrate the JS you wrote into that section, only

Gives us a useful developer tool and avoids the need to use a feature flag as its in the preview and not the main application templates. We can easily expand the set of icons displayed to cover all situations.

ldodds (approved):
Email header/footer look good. Have a think about moving the icon download to the preview. Could be done here or in separate PR.

PR #4060, opened 2024-11-12
Adjust meter data download format

ldodds:
A user was confused by the individual meter download format:

- the MPXN is in the CSV file name but this gets truncated when imported into Excel which has a limit on size of tab names
- the CSV file itself doesn't contain the MPXN

The user was therefore unsure whether we had given them the right data.

Adjust the downloads to shorten the file name and add the mpxn and meter type into the CSV.


app/models/amr_validated_reading.rb
```diff
@@ -67,13 +66,16 @@ def self.download_all_data
 
   def self.download_query_for_meter(meter)
     <<~QUERY
-      SELECT reading_date,
+      SELECT m.mpan_mprn,
+             CASE m.meter_type WHEN 0 THEN 'Electricity' WHEN 1 THEN 'Gas' WHEN 2 THEN 'Solar PV' WHEN 3 THEN 'Exported Solar PV' END,
```
urbanwide, line 70: Should the meter type be translatable if it's a public facing report?
Also pondering if the enum mapping could be used rather than hardcoding here? Not sure if you can with it being pure sql though.

urbanwide (approved):

urbanwide:
See comments 👍

app/models/amr_validated_reading.rb
ldodds: I'm going to add a card for translating the downloads. will require some rework of the query and also how the CSV is generated at the moments its streaming the results from the database query directly to the response using a `COPY`

PR #4061, opened 2024-11-12
Import latest region and local authority associations

ldodds:
Import latest associations between schools and local authorities.

### DIFF
--- a/lib/tasks/deployment/20241112151914_load_regions_and_local_authority_districts_2024_11.rake
+++ b/lib/tasks/deployment/20241112151914_load_regions_and_local_authority_districts_2024_11.rake
@@ -12,7 +12,7 @@ namespace :after_party do
         lad = LocalAuthorityArea.find_by_code(row[2])
         school.update!(region: region, local_authority_area: lad)
       else
-        puts "Can't find school with id #{row[0]}"
+        $stderr.puts "Can't find school with id #{row[0]}"
       end
     end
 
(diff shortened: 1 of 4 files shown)


lib/tasks/deployment/20241112151914_load_regions_and_local_authority_districts_2024_11.rake
```diff
@@ -0,0 +1,24 @@
...
+  task load_regions_and_local_authority_districts_2024_11: :environment do
+    puts "Running deploy task 'load_regions_and_local_authority_districts_2024_11'"
+
+    file_name = File.join( File.expand_path(File.dirname(__FILE__)) , "schools-to-lads-2024-11.csv" )
+    #id,RGN22NM,LAD22CD
+    CSV.foreach( file_name, headers: true ) do |row|
+      school = School.find_by_id(row[0])
+      if school.present?
+        region = row[1].present? ? row[1].parameterize.underscore.to_sym : nil
+        lad = LocalAuthorityArea.find_by_code(row[2])
```
urbanwide, line 12: Do we need to check if the LocalAuthorityArea is found here, or are we pretty confident it will be?

ldodds: If it's not found then setting the value to nil is the right behaviour, I think. I'm using the same list of codes to do the matching so unlikely to be an issue anyway. If the administrative boundaries change then we'll need to update both the database and the boundaries so I think its fine.

urbanwide: Ok, sounds good, thought I better check 👍

urbanwide (approved):

PR #4063, opened 2024-11-13
Add rake task to reload zero reading periods from Meteostat

ldodds:
We've had a period where we seem to have incorrect readings from Meteostat. Reloading data for the period correctly imports updated data. It's affecting 7 weather stations on east of england.

I've added this Rake task to allow days to be selectively reimported if there are any zeroes. I've not done this as an after party task as I prefer to avoid running tasks that depend on external APIs during a release. So I will run the update manually after the next release.

### DIFF
--- a/app/models/weather_observation.rb
+++ b/app/models/weather_observation.rb
@@ -23,6 +23,14 @@ class WeatherObservation < ApplicationRecord
   scope :by_date, -> { order(:reading_date) }
   scope :since, ->(date) { where('reading_date >= ?', date) }
 
+  scope :between, ->(start_date, end_date) do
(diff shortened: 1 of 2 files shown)


tbhi (approved):

app/models/weather_observation.rb
```diff
@@ -23,6 +23,14 @@ class WeatherObservation < ApplicationRecord
   scope :by_date, -> { order(:reading_date) }
   scope :since, ->(date) { where('reading_date >= ?', date) }
 
+  scope :between, ->(start_date, end_date) do
+    where('reading_date >= :start_date AND reading_date <= :end_date',
```
tbhi, line 27: perhaps `where(reading_date: start_date..end_date)`

PR #4065, opened 2024-11-13
Add stark portal gas config

ldodds:
Adds new data feed config for Stark Portal gas data.

I've also added a rails generator to try and make it easier to add these in future. I often just copy from an existing deploy task.

Generators provides a default config for a row per day format with notes about configuring other variants. Future improvements could be to add more options but this is a quick improvement.


urbanwide (approved):
Good idea on the generator 👍

PR #4066, opened 2024-11-13
Allow admins to see equivalences before school is data enabled

ldodds:
Makes the dashboard equivalences component work similar to the components on the adult dashboards.

Normally we allow admins to see the data enabled view even if the flag has not yet been set so they can preview the school's data before making the data live.

### DIFF
--- a/app/components/dashboard_equivalences_component.rb
+++ b/app/components/dashboard_equivalences_component.rb
@@ -24,8 +24,7 @@ class DashboardEquivalencesComponent < ApplicationComponent
   end
 
   def data_enabled?
-    return true if user.present? && user.admin? && @school.process_data?
-    @school.data_enabled?
+    user&.admin? && @school.process_data? || @school.data_enabled?
   end
 
   def render?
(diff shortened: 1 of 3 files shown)


app/components/dashboard_equivalences_component.rb
```diff
@@ -24,6 +24,7 @@ def equivalences
   end
 
   def data_enabled?
+    return true if user.present? && user.admin? && @school.process_data?
```
tbhi, line 27: same as `user&.admin? && @school.process_data? || @school.data_enabled?`?

spec/components/dashboard_equivalences_component_spec.rb
```diff
@@ -45,6 +45,13 @@
       let(:school) { create(:school, data_enabled: false) }
 
       it { expect(component.data_enabled?).to be false }
+
+      context 'with admin and school is processing data' do
+        let(:school) { create(:school, process_data: true) }
+        let(:user) { create(:admin)}
```
tbhi, line 51: missing whitespace

tbhi (approved):

PR #4068, opened 2024-11-13
Improve Weather Station report

ldodds:
Related to https://github.com/Energy-Sparks/energy-sparks/pull/4063

We had a run of around ten days where data we had loaded from Meteostat had a large number of zero readings during the day. The other PR adds a task to allow this type of day to be backfilled.

This PR adjusts the admin data report for weather stations to summarise the average temperature, rather than whether there are missing readings. We rarely, if ever have missing days so reporting on other aspects of the data is more helpful here.

These controllers/views are quite old so I've tidied the code a bit to simplify it. But the primary changes are adjusting the query and views to show the average temperature for a day, to help highlight days where it is unusually cold (or possibly warm). This might also help highlight some other data issues that need following up with Meteostat.

The thresholds for the temperature ranges are a bit arbitrary but I've using 5.0 as lower threshold to highlight days with low temperatures, and `AnalyseHeatingAndHotWater::HeatingModelTemperatureSpace::BALANCE_POINT_TEMPERATURE` as the threshold for a warm day to align with the analytics. We're considering revising this threshold so the updated report helps to highlight those days.

Missing days are still shown on the calendar, but we're not showing any partial days any longer. But as noted this doesn't ever happen as we always request a full day from Meteostat.

Screenshot from 2024-11-13 17-12-31


urbanwide:
Looks good - wondering if there should be some rounding on the average temperature figure shown in the tooltip:
image

ldodds:
Added if you'd like to rereview

PR #4069, opened 2024-11-15
Rework weekly alert emails

tbhi:


app/views/alert_mailer/_alerts.html.erb
```diff
@@ -0,0 +1,66 @@
...
+        </div>
+        <div class="col-lg-11 py-4 text-lg lh-md">
+          <strong><%= sanitize alert_content.email_title %></strong>
+          <%= alert_content.email_content %>
+          <% path = find_out_more_path_from_alert_content(school, alert_content, params: weekly_alert_utm_parameters,
+                                                                                 mailer: true) %>
+          <% if path %>
+            <%= link_to(t('alert_mailer.alert_email.find_out_more'), path, class: 'my-4') %>
+          <% end %>
+          <% if alert_content.unsubscription_uuid %>
```
ldodds, line 23: Minor, but we're not generating these uuids any more so you could remove this conditional.

app/views/alert_mailer/_alerts.html.erb
```diff
@@ -0,0 +1,66 @@
...
+    .positive { background-color: <%= Colours.positive_light %>; }
+    .neutral { background-color: <%= Colours.neutral_light %>; }
+    .lh-md { line-height: 1.75; }
+  </style>
+  <% dashboard_alert_groups(alert_contents).each do |group, alerts| %>
+    <h4 class="py-4"><%= t("advice_pages.alerts.groups.#{group}") %></h4>
+    <% alerts.each do |alert_content| %>
+      <div class="row rounded-lg <%= status_for_alert_colour(alert_content.colour) %>">
+        <div class="col-lg-1 py-3">
+          <%= image_tag "email/fa-#{alert_type_icon(alert_content.alert.alert_type)}.png", class: 'px-4' %>
```
ldodds, line 13: There are some missing icons. This method can return 'bolt', 'fire', 'sun', 'fire-alt' and 'arrow-right'. Fire alt is used for storage heaters so emails don't generate: `The asset "email/fa-fire-alt.png" is not present in the asset pipeline.`

In practice the `fuel_type` for an `AlertType` is currently restricted to gas, electricity and storage heaters so while the two solar related icons aren't likely to be needed now, I think it would be better to include them.

Can you add the extra icons here and in the icon preview. I think we ideally need some variants of the generate email notification specs so that we test generating an alert for every fuel type.

ldodds (changes requested):

app/views/alert_mailer/_alerts.html.erb
tbhi: added icons and different fuel type alerts to the spec

PR #4074, opened 2024-11-18
Todos

urbanwide:
**Revised version of the model with updated naming**

**NB the feature flags have been renamed to todos and todos_parallel**

* Provides the new models for creating lists of activities and actions, now named todos.
* We have two new models, Todo and CompletedTodo. Todos can be used to assign tasks (i.e. currently ActivityTypes or InterventionTypes) to another model (in our case, Audits and Programme Types). CompletedTodo, is used to tick off the progress of the todo, i.e. if it is done or not. This also records the model it was completed against (i.e. Audit or Programme), along with the task recording of the event (Activity or Observation).
* This PR provides the new editing interface for actions and activities, largely based on the current activity and action editing interface for audits. However with the added ability to be able to sort the lists. For audits, this is inline with the audit editing form and for programme_types, this is in a separate form as it is currently.
* There are two feature flags in this PR, todos and todos_parallel. The idea of the todos flag is to switch on the full feature once the rest of the backend is in place (will be a separate PR) but this can be switched on in test initially. The todos_parallel feature is more of a temporary feature, allowing actions to be added to the new backend model in production, before the full feature is switched on. This leaves the old editing interface in place, but also switches on the new actions interface for programme types. 
* You'll also see a todos import task, which can be used to import audit activities and actions and programme activities and from the current system. The plan would be to run this (although will modify it to add existing progress for programme actions) when the new feature is fully switched on to migrate the existing data over.

In short. To view the full editing feature as intended, switch on the todos feature. To view just the additional programme type action interface, switch on todos_parallel


app/components/toggler_dropdown_component.rb
```diff
@@ -0,0 +1,17 @@
+# Put content in togglable box that is closed by default
+
+class TogglerDropdownComponent < ApplicationComponent
```
urbanwide, line 3: Made this to 'deprecate' things in the admin interface, but I've decided not to use it. Have left it here for now.

app/services/audits/progress.rb
```diff
@@ -33,15 +33,15 @@ def notification
     end
 
     def completed_activities_count
-      audit.completed_activity_types.count
+      audit.activity_types_completed.count
```
urbanwide, line 36: These have been renames so we could use the namespace in the new model.

urbanwide:
Thanks - nice spot on the adding of new items not using select2 - hadn't noticed that difference sorry, was too wrapped up in the other javascript :-)

> this might be on your list for the follow on PR but the admin report for actions needs a column for Programmes (/admin/reports/intervention_types)
Yes this is on a todo later list. But appreciate it's more admin based so probably should be here. Think this PR could be big enough for now

All other stuff is also helpful feedback, thank you

ldodds (approved):
Looks good. Thanks for doing the renaming, think it reads much clearer. I've not reviewed the data migration as you said you might revise that later. So will take a look at that before we're going to use it.

urbanwide:
No probs, thanks for working it through. This way sits much better for me too and sets things up well for the direction of any future refactoring too.

PR #4076, opened 2024-11-19
improve heating control advice for schools with limited data

tbhi:


config/locales/views/advice_pages/heating_control.yml
```diff
@@ -57,6 +57,16 @@ en:
         view_our_analysis_html: <a href="%{href}">View our analysis</a>
         warm_weather:
           title: Leaving the heating on during warm weather
+      not_enough_data:
+        p1: When we have additional data for your school we will be able to provide you with detailed analysis which will
```
ldodds, line 61: missing word "include"?

ldodds, line 61: Reviewing the text again, can we please include the `not_enough_data.available_from` message after these bullet points, so it described what will be in the analysis and then "Assuming we continue to regularly receive data we expect this analysis to be available after %{available_from}.", before going into the "in the meantime" part.

ldodds (commented):
Changes to text look OK, but can we add a test for this please?

app/views/schools/advice/heating_control/not_enough_data.html.erb
```diff
@@ -0,0 +1,32 @@
...
+  </h2>
+  <p>
+    <%= t('advice_pages.heating_control.not_enough_data.p1') %>
+    <ul>
+      <li><%= t('advice_pages.heating_control.not_enough_data.p1_li1') %></li>
+      <li><%= t('advice_pages.heating_control.not_enough_data.p1_li2') %></li>
+      <li><%= t('advice_pages.heating_control.not_enough_data.p1_li3') %></li>
+    </ul>
+  </p>
+  <% if @analysable.data_available_from.present? %>
```
tbhi, line 15: it seems like this is always nil so perhaps shouldn't have? https://github.com/Energy-Sparks/energy-sparks/blob/bae329427d699a77fc2521a87c494d4a2da0b1ad/app/services/schools/advice/heating_control_service.rb#L37

ldodds: Sorry, I'd hadn't realised we didn't have an available date here, so yes best remove.

ldodds: Actually, I think its fine to leave here for now. We can sort out estimating a suitable time separately.

ldodds (approved):

PR #4080, opened 2024-11-20
break up other section of admin index

tbhi:
feels like would be good to break up the other section, not exactly sure on naming


ldodds (approved):
I had been thinking similarly, but was considering adding an extra "Developer" menu item alongside Admin and Reports. But this is totally fine. I'd put the Site colour palette in the dev list.

PR #4084, opened 2024-11-21
comparison refresh improvements

tbhi:
- keep going if one refresh fails
- analytics changes


ldodds (commented):
Separate to how we handle the floor area issue, I think the changes to ensure each view is refreshed look good.

ldodds (approved):
See comment on analytics PR

PR #4091, opened 2024-11-27
Revise styling in emails

tbhi:
- change primary button colour and border radius to match web design
- change main font size to match web design (18px)
- other tweaks to alerts list to be more inline with web design


ldodds:
Having looked at them I think they would look better with either the success or the secondary style buttons rather than the dark blue. The teal success buttons are in the new Mailchimp template and are used as call for actions in the design. Could you try that?

The button text also seems too small. The default buttons are 16px font I think?

ldodds (approved):

PR #4093, opened 2024-11-29
initial SolisCloud integration

ldodds:
Just a quick spike to test out integration with SolisCloud, so I can see actual data for an installation.

Implements the API authentication and a couple of methods so I could inspect the JSON.

The user_station_list and station_day methods work but I've not had data from inverter_day which may be due to the metering at the site.


tbhi:
based initial implementation on the spike so have updated the same branch

app/services/solar/solis_cloud_upserter.rb
```diff
@@ -0,0 +1,15 @@
...
+module Solar
+  class SolisCloudUpserter < BaseUpserter
+    private
+
+    def meter_attributes(details)
+      { pseudo: true, solis_cloud_installation: @installation, meter_serial_number: details['id'] }
+    end
+
+    def synthetic_mpan(meter_type, details)
+      Dashboard::Meter.synthetic_combined_meter_mpan_mprn_from_urn(details['sno'].to_i(16), meter_type)
```
tbhi, line 12: assuming this is unique and in hex

app/models/solis_cloud_installation.rb
```diff
@@ -0,0 +1,39 @@
...
+# Foreign Keys
+#
+#  fk_rails_...  (amr_data_feed_config_id => amr_data_feed_configs.id) ON DELETE => cascade
+#  fk_rails_...  (school_id => schools.id) ON DELETE => cascade
+#
+class SolisCloudInstallation < ApplicationRecord
+  belongs_to :school, inverse_of: :solar_edge_installations
+  belongs_to :amr_data_feed_config
+
+  has_many :meters, dependent: :destroy
```
ldodds, line 28: This would delete any historical data, so not sure we want to use `:destroy` here. That's what we're doing with SolarEdge

app/models/solis_cloud_installation.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  # validates_presence_of :site_id, :mpan, :api_key
```
ldodds, line 30: Probably want some validation to ensure api id, secret, etc added?

app/services/solar/solis_cloud_upserter.rb
```diff
@@ -0,0 +1,15 @@
+require 'dashboard'
+
+module Solar
+  class SolisCloudUpserter < BaseUpserter
+    private
+
+    def meter_attributes(details)
```
ldodds, line 7: Minor, but initially thought this was referring to MeterAttributes. Wonder if could rename, e.g. `meter_defaults`?

lib/data_feeds/solis_cloud_api.rb
```diff
@@ -0,0 +1,64 @@
...
+      @api_id = api_id
+      @api_secret = api_secret
+    end
+
+    def user_station_list
+      get_data('/v1/api/userStationList', {})
+    end
+
+    def station_day(id, day)
+      sleep 2
```
ldodds, line 24: This might avoid throttling for this specific instance, but if we've got multiple reloads happening (e.g. triggered by admins) and daily loads across servers then we might still hit throttle. So we should have some kind of retry behaviour in the requests handling too. We're using Faraday Retry elsewhere.

tbhi: added retry and limiter

lib/data_feeds/solis_cloud_api.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    private
+
+    def connection
+      Faraday.new(BASE_URL) do |f|
+        f.response :json
+        f.response :raise_error
+        f.response :logger if Rails.env.development?
+        f.request(:retry, { retry_statuses: [429], interval: 2.0, backoff_factor: 2 })
```
tbhi, line 36: not sure how the api will throttle

ldodds:
PR looks good. I've reviewed but can't approve as its technically my PR. Fine to merge, or maybe @urbanwide can give the :+1:

urbanwide (approved):

PR #4104, opened 2024-12-06
Don't reject incomplete readings when merging

ldodds:
I recently revised the loader to allow row per reading formats to be configured to allow merging. This meant tolerating days with partial readings and then later merging in any new readings when they are provided during later data loads.

E.g. some suppliers miss the last reading of the day in one file, then its provided in the next.

While I'd revised the `SingleReaderConverter` to not reject readings and updated the `DataFeedUpserter` to support merging of arrays during the upsert call, I'd missed the fact that the AmrReadingData class was also validating readings and rejecting the partial days. I hadn't added an end-to-end test for the entire loader so it got missed.

I've fixed this problem by updating `AmrReadingData` to skip the missing reading validation if the config is set to `allow_merging`. I've also refactored it to be passed an `AmrDataFeedConfig` rather than individual settings.

That's the only functional change. The rest of this PR is a big rework of the specs for `AmrReadingData` and `CSVParserAndUpserter`. The latter specs were particularly messy and it was unclear what was actually being tested in some of the specs and I found a few bugs in the specs.

I switched the `CSVParserAndUpserter` spec from using FakeFS and embedding all the data in the spec to using separate fixture files. I think this makes the spec a more readable, at the cost of adding some small extra files.


app/models/amr_reading_data.rb
```diff
@@ -113,8 +114,18 @@ def inconsistent_reading_date_format?(reading_date)
...
+  #
+  # If we are merging partial data received from suppliers, then we never
+  # generate a warning
+  #
+  # Otherwise we check whether there are more than an allowed threshold.
+  # For "row per reading" formats this is configured in the config, but
+  # for "row per day" formats the threshold is always zero.
   def missing_readings?(readings)
-    readings.compact.count {|reading| reading.present? && reading != '-'} < (48 - @missing_reading_threshold)
+    return false if @amr_data_feed_config.allow_merging?
```
ldodds, line 126: This is the actual functional change that fixes the bug

spec/models/amr_reading_data_spec.rb
```diff
@@ -1,252 +1,268 @@
...
-    end
-
-    it 'knows when it is valid' do
-      expect(amr_reading.valid?).to be true
-      expect(amr_reading.valid_reading_count).to be 2
-      expect(amr_reading.warnings?).to be false
-      expect(amr_reading.warnings.count).to be 0
-    end
+  subject(:amr_reading_data) do
+    AmrReadingData.new(amr_data_feed_config: amr_data_feed_config, reading_data: reading_data)
```
tbhi, line 5: described_class

tbhi (approved):
looks like might be the only user of fakefs so perhaps that can be removed aswell?

PR #4113, opened 2024-12-09
Automating the Perse HH data downloads via their API

tbhi:


app/services/amr/perse_upsert.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Amr
+  module PerseUpsert
+    def self.perform(meter)
+      config = AmrDataFeedConfig.find_by!(identifier: 'perse-half-hourly-api')
+      date = AmrDataFeedReading.where(amr_data_feed_config: config, meter: meter)
```
ldodds, line 7: could use `MAX` here?

app/views/schools/meters/_form.html.erb
```diff
@@ -75,6 +75,12 @@
           <%= t('schools.meters.form.leave_this_blank_message') %>
         </small>
       </div>
+
+      <div class="form-group">
+        <%= form.label(:perse_api, 'Perse API') %>
```
ldodds, line 80: Could add a note similar to the `text-muted` section above to explain this will enable loading from Perse.

The `alert-warning` in this section refers to n3rgy, might want to update that to "for n3rgy and perse integrations"

lib/data_feeds/perse_api.rb
```diff
@@ -0,0 +1,32 @@
...
+      connection.get('/meterhistory/v2/realtime-data', params).body
+    rescue Faraday::Error => e
+      EnergySparks::Log.exception(e, mpan:, from_date:, status: e&.response_status, body: e&.response_body)
+      {}
+    end
+
+    def self.meter_history_readings(mpan, from_date)
+      meter_history_realtime_data(mpan, from_date)['data']
+        &.select { |data| data_ok(data) }
+        &.map { |data| [data['Date'], (1..48).map { |i| data["P#{i}"].to_f }] } || []
```
ldodds, line 25: Are there ever any missing or blank half-hourly readings? We'll want missing/blank to be left as nil so we end up storing what was returned via the API. Otherwise they get set to zero here.

app/jobs/perse_reload_job.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class PerseReloadJob < ApplicationJob
+  queue_as :default
+
+  def perform(meter, notify_email)
+    result = Amr::PerseUpsert.perform(meter)
```
ldodds, line 7: This just runs the loader rather than triggering a reload of the last 14 months

app/services/amr/perse_upsert.rb
ldodds: as the show.html.erb uses the same query, probably worth extracting a method?

app/services/amr/perse_upsert.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Amr
+  module PerseUpsert
+    def self.perform(meter, reload: false)
+      config = AmrDataFeedConfig.find_by!(identifier: 'perse-half-hourly-api')
+      date = 14.months.ago
+      date = latest_reading_date(config, meter) || date unless reload
```
urbanwide, line 8: Is it possible to add a test to check that the right date is used if reload is / isn't provided?

urbanwide (commented):
Looks good - just a few small questions / comments

app/services/amr/perse_upsert.rb
tbhi: added to https://github.com/Energy-Sparks/energy-sparks/blob/af6a502a7f0ef04c43b97da5c4433d5dabc02934/spec/system/meter_management_spec.rb#L302 to check the reload button doesn't use the last reading date - does that seem ok?

app/jobs/perse_reload_job.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class PerseReloadJob < ApplicationJob
+  queue_as :default
+
+  def perform(meter, notify_email)
+    result = Amr::PerseUpsert.perform(meter, reload: true)
+    N3rgyReloadJobMailer.with(to: notify_email, meter:, result:).complete.deliver
```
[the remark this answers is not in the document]
tbhi: believe they're used exclusively so seems ok

app/services/amr/perse_upsert.rb
urbanwide: Yes. In an ideal world you'd probably want to add a test to perse_upsert_spec.rb but ok if you're happy it's been done here.

urbanwide (approved):

PR #4123, opened 2024-12-16
academic year for tariff/costs page

tbhi:


ldodds (approved):
Looks good. Although testing this I've found a fairly serious bug in the gas costs pages. I'll create a PR for that based off this one.

See comment on the baseload page re: releasing.

PR #4124, opened 2024-12-19
make baseload advice longterm table use academic years

tbhi:


ldodds (approved):
Changes to the calculations look correct and have briefly tested locally so I think this is OK to merge.

However the original plan was for the academic year changes for the long term, baseload and tariffs pages to be behind the same feature flag so they can be enabled/disabled together. Looks like it'll be fiddly to refactor to that now, so suggest we just focus on getting this released and ensuring its been manually tested on test server before prod. 

So merge this and other PRs tomorrow for release Monday

I'd like to later refactor the baseload service to move the calculations you've added into a service in the analytics so we're retaining a standard API for the basic analysis. This will later allow us to move to generating more of these metrics overnight. The `services/school/advice/*` classes are already fairly complicated.

It doesn't look like the `BaseloadAnnualBreakdownService` is actually being used so that could be refactored to look something like this:

```
BaseloadAnnualBreakdownService
 def initialize(meter) # real or aggregate meter

 # returns hash of 
 # period: (a SchoolPeriod), 
 # partial:, 
 # baseload_kw:, 
 # baseload_usage: (a CombinedUsageMetric)
 # 
 # can be extended to support :year, :academic_year
 def annual_baseload_breakdown(period: :fixed_academic_year)

 # as per current code
 def enough_data?
 def data_available_from
 
end 
```

PR #4142, opened 2025-01-06
Todos displaying

urbanwide:
### Viewing programme and audit activities and actions

- [x] change the programme type show page to list both activities and actions. List the pupil activities first, then the adult. Use “Pupil-led activities” as title for pupil activities and “Energy saving actions” for adult. The page should check off items that are done on either list
- [x] Ensure that the audit page checks off items that have been completed, and also notes when the whole audit has been completed, as we do with programmes

### Activity / action completed pages

- [x] ensure that when a user completes an adult activity, the completed page notes that a programme is completed if they’ve done so. This is currently only done for activities
- [x] ensure that when a user completes an adult action OR pupil activities, the complete page notes if they’ve completed any of the adult OR pupil activities in a programme and prompts them to join. Currently this only checks for pupil activities

### Prompts

- [x] Prompt to complete programme: update the notices shown on, e.g. pupil and adult dashboard, and recommendations pages, that say things like “You have completed 5/8 of the activities in the Get Energised! programme” to include counts of actions too. See existing prompts for completing audits which uses similar language already. The scores for completion should cover both pupil and adult activities
- [x] Complete programme spec
- [x] Audit Prompt needs updating to work with todos
- [x] Join prompt: Update the join prompt to take note of actions as well as activities
- [x] School#suggested_programme_type spec

### Activites and action index pages
- for the programmes block on the actions page, the programmes in that list should only include those that have any adult actions.
- update the /intervention_type_groups index page to also have a programmes block like the activities page.

### Programme index page

- Update the programme index page to include some idea of the scope of the programme, e.g as a sentence above the one indicating the number of points to be scored:
“This programme is intended for pupils” (if there are no adult actions)
“This programme is intended for adults” (if there are no pupil activities)
“This programme is intended for the whole school” (if there are both).

### Admin

- update /admin/reports/intervention_types so that is has a column for programmes, like the activity type version
- update /admin/programme_types/12/programmes (for example)
- Change import script to also import programme action activity from this school year if this needs doing


app/models/audit.rb
```diff
@@ -94,6 +94,10 @@ def completed?
     observations.audit_activities_completed.any?
   end
 
+  def tasks_completed_on
+    observations.audit_activities_completed.last.at
```
ldodds, line 98: I assume there's only ever one of these so taking the last is fine, rather than sorting by `at`.

app/models/activity.rb
```diff
@@ -64,4 +64,8 @@ def display_name
   def points
     observations.sum(:points)
   end
+
+  def recorded_on
```
ldodds, line 68: To me "recorded on" would be the date the activity was recored, so `created_at` not `happened_on`.

config/locales/views/schools/schools.yml
```diff
@@ -354,7 +355,25 @@ en:
...
+          completed_message_html:
+            one: You've completed all the tasks in the <strong>%{title}</strong> programme. Mark it done to score <strong>%{count}</strong> bonus point?
+            other: You've completed all the tasks in the <strong>%{title}</strong> programme. Mark it done to score <strong>%{count}</strong> bonus points?
+            zero: You've completed all the tasks in the <strong>%{title}</strong> programme. Mark it as complete?
+          message_html:
+            one: You've recently completed a task that is part of the <strong>%{title}</strong> programme. Do you want to enrol in the programme?
+            other: You've recently completed %{count} tasks that are part of the <strong>%{title}</strong> programme. Do you want to enrol in the programme?
         message_html: You have completed <strong>%{programme_activities_count}/%{activity_types_count}</strong> of the activities in the <strong>%{programme_type_title}</strong> programme
+        progress:
+          message_actions_only_html: You have completed <strong>%{completed_actions_count}/%{total_actions_count}</strong> of the actions in the <strong>%{title}</strong> programme
```
ldodds, line 368: Seems inconsistent to refer to activities/actions/or both when reporting progress, when the rest of the wording only refers to "tasks". It'd be more consistent to just use "tasks" everywhere I think.

spec/components/previews/toggler_dropdown_component_preview/examples.html.erb
```diff
@@ -0,0 +1,18 @@
...
+    <h2>title: 'Title'</h2>
+
+    <%= render TogglerDropdownComponent.new(title: 'Title') do %>
+      Content in here
+    <% end %>
+
+    <h2>surround: false</h2>
+    <p>Just renders content as if component not there</p>
+
+    <%= render TogglerDropdownComponent.new(title: 'Title', surround: false) do %>
```
ldodds, line 13: I get an error as there's no `surround` keyword defined in the constructor

ldodds: After I fixed this to see the examples, I noticed that if you toggle the second one then the first one opens. Think the target attributes need to adjusting so that we could have multiple on a page (although not sure how much we'll be using this in user facing pages).

app/models/activity_type.rb
```diff
@@ -142,6 +142,10 @@ def count_existing_for_academic_year(school, academic_year)
     school.activities.where(activity_type: self).where(happened_on: academic_year.start_date..academic_year.end_date).count
   end
 
+  def public_type
+    'activity'
```
ldodds, line 146: maybe a symbol? (I just ended up double checking that public_type wasn't used to display text, rather than being used in an i18n key).

ldodds (approved):

app/models/activity.rb
urbanwide: This method is not used now so have removed it

spec/components/previews/toggler_dropdown_component_preview/examples.html.erb
urbanwide: Thanks - I changed the name of the parameter just before the final commit and forgot about the preview...

Have fixed the second toggle issue though thanks!

config/locales/views/schools/schools.yml
urbanwide: Soo... do you mean always refer to them together? So rather than splitting them out as we do now, always combine the two?
Rather than:
`You have completed <strong>%{completed_activities_count}/%{total_activities_count}</strong> of the activities and <strong>%{completed_actions_count}/%{total_actions_count}</strong> of the actions in the <strong>%{title}</strong> programme
`We have this instead:
You have completed <strong>%{completed_tasks_count}/%{total_tasks_count}</strong> of the tasks in the <strong>%{title}</strong> programme

ldodds: Yes. I thought that's what'd decided when you asked the question on Trello?

You had this as the example: ”You've recently completed 6 tasks that are part of the Cut Out the High Carbon Commute programme. Do you want to enrol in the programme?” of the preferred approach I'd read that as an example of how we'd frame all the messages?

urbanwide:
Hi @ldodds - have made changes as suggested. While I was there, I also noticed there was no specific text for prompts if no tasks had been completed yet, so I added some in as it looked strange without it.

i.e. previously we had:

'You have completed 0/6 of the tasks from your recent energy audit. Complete the others to score 165 points and 50 bonus points for completing all audit tasks'

Now we have:
'You haven't yet completed any of the tasks recommended in your recent energy audit. If you complete them, you will score 165 points and 50 bonus points for completing all audit tasks'

Also added similar for programmes. Let me know if the text needs changing though.

Thanks.

PR #4150, opened 2025-01-09
Fix gas cost chart when there is a single meter

ldodds:
Testing #4123 I noticed a long standing issue with the gas costs chart.

The issue only affects schools with a single gas meter and only on the 2 year comparison chart.

The cost pages use charts named with an `electricity_` prefix. The default definition of that is to show the electricity costs. However the charts override that by specifying a meter (either a real gas/electricity meter or the aggregate one), which means the charts then shows the costs for that meter.

When a school has a single meter, the 2 year comparison chart was not specifying a meter, so it was defaulting to the electricity costs.

I've fixed this by updating the partials to build a chart_config for the aggregate (gas or electricity) meter so the right values are used.

The naming is a hold-over from previous versions of these pages and ideally needs tidying up. We should do that as a separate exercise, but there's a lot of other chart configurations to tidy as well.

### DIFF
--- a/app/views/schools/advice/electricity_costs/_analysis.html.erb
+++ b/app/views/schools/advice/electricity_costs/_analysis.html.erb
@@ -45,6 +45,6 @@
     show_school_total: true,
     fuel_type: @advice_page.fuel_type %>
   <% if @analysis_dates.months_of_data > 23 %>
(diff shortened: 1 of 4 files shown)


tbhi (approved):
surprised no erblint complaints

PR #4164, opened 2025-01-13
Fix DCC Meter option on form

ldodds:
The DCC meter status was recently changed from a boolean to an enum.

However the meter creation and editing form was still setup as a checkbox. This meant that the status wasn't showing or updating correctly. The status was being reset to "no" after an update. I think this might also have been triggering unnecessary re-checks of the status. 

I've updated the form and the spec to ensure the values are available.


ldodds:
@tbhi can you review this for todays release please?

tbhi (approved):

PR #4169, opened 2025-01-15
SolisCloud jobs & UI

tbhi:


app/views/schools/solis_cloud_installations/_form.html.erb
```diff
@@ -0,0 +1,11 @@
+<%= simple_form_for [school, installation] do |form| %>
+  <%= render 'shared/errors', subject: installation, subject_name: 'solis cloud api feed' %>
+  <%= form.input :api_id, as: :string, label: 'API ID',
+                          hint: 'The code that authorises us to access data for this site.' %>
```
ldodds, line 4: "The Solis Cloud API Key"?

app/views/schools/solis_cloud_installations/_form.html.erb
```diff
@@ -0,0 +1,11 @@
...
+  <%= form.input :api_secret, as: :string, label: 'API Secret',
+                              hint: 'The code that authorises us to access data for this site.' %>
```
ldodds, line 6: "The Solid Cloud API Secret"?

ldodds (approved):
The admin interface seems to work OK. Left a couple of comments and sent you details of an error encountered with the upserter.

After a successful import its not immediately clear which meter is for which station without looking at the metadata. Wondering if we can make that any clearer for admins/users? Could use the station id perhaps? I can't see a friendlier name in the metadata unless thats available some other way?

PR #4180, opened 2025-01-21
Add fields to school group user export

ldodds:
Adds extra fields to the school group user download so its similar to the main report.

Includes school type, whether they are active, funder, etc.

### DIFF
--- a/app/controllers/admin/school_groups/users_controller.rb
+++ b/app/controllers/admin/school_groups/users_controller.rb
@@ -49,6 +49,11 @@ module Admin
           csv << [
             'School Group',
             'School',
+            'School type',
+            'School active',
+            'School data enabled',
+            'Funder',
+            'Region',
             'Name',
             'Email',
             'Role',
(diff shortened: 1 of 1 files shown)


app/controllers/admin/school_groups/users_controller.rb
```diff
@@ -73,7 +78,20 @@ def produce_csv(school_group, group_admins, school_users)
...
+          if school
+            school&.active? ? 'Yes' : 'No'
+          else
+            ''
+          end,
+          if school
+            school&.data_enabled? ? 'Yes' : 'No'
+          else
+            ''
+          end,
```
tbhi, line 92: perhaps `school ? (school.data_enabled? ? 'Yes' : 'No') : ''`

tbhi (approved):

app/controllers/admin/school_groups/users_controller.rb
ldodds: This is what I had originally, but Rubocop wants if statements:

```
[Correctable] Style/NestedTernaryOperator: Ternary operators must not be nested. Prefer `if` or `else` constructs instead.
```

PR #4182, opened 2025-01-22
Restyle programme page prompts

urbanwide:
@ldodds some test restyling the programme page prompts (shown at the top) and todo items (also shown on audit page). I used the prompt component for the prompt at the top then thought I would try it for the todo items themselves as they would line up nicely with it. See what you think. There is still some i18n to put in and tests to fix, but will sort that after feedback.

Not enrolled yet:
Screenshot 2025-01-23 at 11 27 05

Enrolled, nothing done yet:
Screenshot 2025-01-23 at 11 26 55

Enrolled, some progress made:
Screenshot 2025-01-23 at 11 26 33

Completed programme:
Screenshot 2025-01-23 at 11 26 45


config/locales/common.yml
```diff
@@ -73,6 +73,7 @@ en:
       yes_label: 'Yes'
     line_graphs: Line graphs
     pie_charts: Pie charts
+    points_html: "%{count} points"
```
ldodds, line 76: does this need `_html`?

PR #4191, opened 2025-01-24
Link to edit dashboard message from icon

urbanwide:
Was using the interface and felt like I should be able to edit a dashboard message by clicking on the info icon. Could be wrong though! This meant the info icon is added in a greyed out form for when a message is not set.
Admin school group / view page
Screenshot 2025-01-24 at 00 14 00

Admin school group / index page
Screenshot 2025-01-24 at 00 13 26

### DIFF
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -201,6 +201,11 @@ input, progress {
   color: $grey-light;
 }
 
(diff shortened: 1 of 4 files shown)


ldodds (approved):
Looks fine to me. Might want to call this out on the team call so people are aware.

PR #4192, opened 2025-01-24
Todos programme, activities and action index pages

urbanwide:
**Activities and action index pages**

- [x] - update the /intervention_type_groups index page to also have a programmes block like the activities page.
- [x] - for the programmes block on the actions page, the programmes in that list should only include those that have any adult actions.

I have not gone to town with the spec for the above, as I'm assuming these pages will need re-skinning at some point and therefore re-speccing.

Could do with some new text for inside the programme block now on activities and action index pages. Currently we have: `Short programmes of related activities that pupils can work through step-by-step to achieve greater impact.` Obviously now programmes can contain activities or actions and can be carried out by either pupil or adults. Maybe something like?
`Short programmes of related tasks that pupils or adults can work through step-by-step to achieve greater impact.
`
image

- [x] - Update recommendations logic - was using old audit activities and actions relationships

**Programme index page**

- [x] - Update the programme index page to include some idea of the scope of the programme, e.g as a sentence above the one indicating the number of points to be scored: “This programme is intended for pupils” (if there are no adult actions) “This programme is intended for adults” (if there are no pupil activities) “This programme is intended for the whole school” (if there are both).

Have added this to programme index and programme show pages. Also added an icon but not convinced there are icons that work for pupils / adults / whole school. See what you think:

Screenshot 2025-01-26 at 21 53 36
Screenshot 2025-01-26 at 21 53 27
Screenshot 2025-01-26 at 21 53 20

There is some redesign of these pages (programmes, activities / actions / recommendations)  to be done but would like to chat on how we'd like to go about it first and perhaps do it in another PR later on.

**Other**

- [x] Only allow a programme to be repeated if school has not already completed it this academic year.

Old logic was allowing repeat if school had completed this programme in a previous academic year.
Which meant that - if the school had previously completed the programme in a different academic year, then completed it again this year, that when clicking repeat after this section completion (yes I know it's getting unlikely this will happen), if activities / actions had been completed this year already, the progress of those are automatically added when restarting a programme, meaning the programme would be automatically completed again. This would have allowed the school to continuously complete programmes by doing nothing, just clicking restart.


app/views/programme_types/_intention.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if programme_type.activity_type_todos.none? %>
+  <%= render IconComponent.new(name: 'chalkboard-user') %> <%= t('programme_types.intention.adults') %>
+<% elsif programme_type.intervention_type_todos.none? %>
+  <%= render IconComponent.new(name: 'book-open-reader') %> <%= t('programme_types.intention.pupils') %>
+<% else %>
+  <%= render IconComponent.new(name: 'users') %> <%= t('programme_types.intention.all') %>
```
ldodds, line 6: There's a school icon https://fontawesome.com/icons/school?f=classic&s=solid

app/views/programme_types/_intention.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if programme_type.activity_type_todos.none? %>
+  <%= render IconComponent.new(name: 'chalkboard-user') %> <%= t('programme_types.intention.adults') %>
```
ldodds, line 2: Could use users or maybe people-group https://fontawesome.com/icons/people-group?f=classic&s=solid if you switch to using a school icon.

I think the pupil icon works

ldodds (approved):

ldodds:
For wording, how about something generic: "Short programmes of related tasks that can be worked through step-by-step to achieve greater impact"?

re: re-design. Suggest we get this set of features done before spending more time on this as the home page, etc updates will be more impactful.

When it comes to revisiting, I think we could use more components here, e.g. 

- a component for displaying a single activity/action/programme as a "card" or row in results
- a component for showing a list of items, e.g. for recommendations or other card-deck use cases
- a component for the individual task items, that handles their state
- maybe one for an icon with a note, as you're using for intention and the programme status
- in the dashboard redesign I did away for using card-deck and some of the flex options as they don't work very well on mobile, I found that using Bootstrap columns, with different breakpoints was generally enough
- the styling of the card-decks should lift from the new wireframes which might mean: separately coloured backgrounds/callouts, different box styling (new design has more rounded corners), etc

app/views/programme_types/_intention.html.erb
urbanwide: Thanks for these suggestions, have made the changes

urbanwide:
Thanks for the review and the discussion around the redesign @ldodds. That all makes good sense and I've added it to a separate Trello card so we don't lose it. Do you want a last check of this before I merge?

PR #4193, opened 2025-01-25
Engaged School Report changes

tbhi:


app/models/school.rb
```diff
@@ -749,6 +749,10 @@ def self.school_list_for_login_form
                                             'school_groups.name as school_group_name').where(visible: true).order(:name)
   end
 
+  def data_visible?
```
ldodds, line 752: still worth adding but SchoolCreator.make_data_enabled! enforces that a school can only be made data_enabled if its first visible. I guess the flags could be adjusted separately though, e.g. if school is made not visible after being made data enabled.

ldodds (approved):

PR #4194, opened 2025-01-27
Mailchimp changes tracking

ldodds:
To avoid having a very large PR, I'm breaking the Mailchimp synchronisation changes up into smaller PRs.

The overall set of changes will have:

1. Code to track changes in our database that must be pushed to Mailchimp (THIS PR)
2. Code to populate our database with current subscriber status in Mailchimp ([HERE](https://github.com/Energy-Sparks/energy-sparks/pull/4210))
3. Code to push changes to Mailchimp via the AudienceManager
4. Some admin reports to help monitor things

The key changes in this PR are:

- adding a timestamp (`mailchimp_fields_changed_at`) to all of the models that hold data that is pushed to Mailchimp for a given User (Mailchimp contact)
- a new `MailchimpUpdateable` concern which adds callbacks on models to ensure that the timestamp is populated correctly before saving
- an additional timestamp and status on the User model to tracking when the user was updated in Mailchimp and the subscriber status
- a new scope `User.mailchimp_update_required` which lists users that need to have data pushed to Mailchimp

Maintaining the `mailchimp_fields_changed_at` timestamp is mostly done by using `ActiveRecord::Dirty` and looking for changes to specific fields. As the the Scoreboard `name` property is managed via Mobility, we can't tell if attribute has changed so I've enabled dirty record tracking for Mobility so that works. But its only enabled for this model and attribute currently so should minimise impact elsewhere.

There are some User model associations (e.g. cluster schools, alert contacts) which have some custom code/callbacks that directly update the `mailchimp_fields_changed_at` timestamp on meaningful changes.

Overall this is simpler than the approach I was originally working on!


app/models/user.rb
```diff
@@ -84,6 +91,22 @@ class User < ApplicationRecord
 
   scope :alertable, -> { where(role: [User.roles[:staff], User.roles[:school_admin], User.roles[:volunteer]]) }
 
+  scope :mailchimp_update_required, -> do
```
urbanwide, line 94: Whilst I like the light approach on the callback side - definitely good that no extra database writes are done there but I do worry that this query is rather large. Assume it is something that would be run in a batch job in the middle of the night?

urbanwide (approved):
Looks good but I worry a bit about the rather large query in the scope mailchimp_update_required, assume would be ok if run out of hours though?

app/models/funder.rb
```diff
@@ -4,10 +4,15 @@
...
-#  id   :bigint(8)        not null, primary key
-#  name :string           not null
+#  id                          :bigint(8)        not null, primary key
+#  mailchimp_fields_changed_at :datetime
+#  name                        :string           not null
 #
 class Funder < ApplicationRecord
+  include MailchimpUpdateable
+
+  MAILCHIMP_FIELDS = [:name].freeze
```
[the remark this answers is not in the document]
ldodds: Yes, that's nicer. I can add that.

app/models/user.rb
ldodds: Planning to run daily initially, but maybe more frequently if needed.

I'm not too worried about it as in my db (with statuses set on ~1300 users, but no updates pushed) it runs in 28ms. The query is mostly LEFT JOINS which should be optimised well in the database.

The query is not fetching anything from the joined tables, so doesn't need to pull anything back from the database other than the User attributes. 

Can see how it runs in practice though.

urbanwide: That sounds good then 👍

PR #4196, opened 2025-01-27
Todos sure up

urbanwide:
Spotted a couple of things that needed changing after a good poke around. 

Firstly when observations are 'removed' via the frontend (via the ObservationRemoval service), if not in current academic year, visible is set to false, rather than it being removed. If they are in this academic year, they are destroyed, which automatically removes completed todos etc, which is how I assumed it worked all the time. So now this takes note of the visible flag in a number of places and removes completed todos when visible is set to false.

Also made sure that when picking the latest recording for a todo, there is a possibility that more than 1 recording occurs on the same day. Added code to make sure we always pick the last one created. It probably would have done this anyway, just wanted to be intentional.

### DIFF
--- a/app/models/intervention_type.rb
+++ b/app/models/intervention_type.rb
@@ -79,7 +79,7 @@ class InterventionType < ApplicationRecord
   before_save :copy_searchable_attributes
 
   def actions_for_school(school)
-    observations.for_school(school)
+    observations.visible.for_school(school)
   end
 
   # override default name for this resource in transifex
--- a/app/models/todo.rb
+++ b/app/models/todo.rb
@@ -59,7 +59,7 @@ class Todo < ApplicationRecord
     when 'ActivityType'
       school.activities.where(activity_type: task)
     when 'InterventionType'
-      school.observations.intervention.where(intervention_type: task)
+      school.observations.intervention.visible.where(intervention_type: task)
     else
       raise StandardError, 'Unsupported task type'
     end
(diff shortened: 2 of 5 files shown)


urbanwide:
What I don't get is why if an observation is this calendar year then it is removed but in a previous calendar year it is made invisible. If it were to be to retain scores, don't our scoreboards only count for this calendar year anyway?

urbanwide:
In my local DB there are only 14 observations that are not visible. It seems like it was intentional to add this functionality but can't work out why. It's a bit inconsistent.

ldodds:
> What I don't get is why if an observation is this calendar year then it is removed but in a previous calendar year it is made invisible. If it were to be to retain scores, don't our scoreboards only count for this calendar year anyway?

You can retrieve older versions of scoreboards, e.g `/scoreboards/north-west-england?previous_year=true`. I think previously you could also request them for any previous academic years but we removed that as it wasn't that useful.

I can only think the visibility is there to allow scores to be retained, as the observation values are still counted. This was [the original commit](https://github.com/Energy-Sparks/energy-sparks/commit/5b0b54b54e039143599b524114608dc0cfbd9a05#diff-224459a707b142a57ce9d518843a0a5f54d8f44b8599be9e2cef515532f53f86).

It's obviously been pretty rare to retrospectively remove items. When we're recently done updates because, e.g a school has recorded an activity multiple times, we just set the observation to zero. Which is better than removing or hiding anything.

urbanwide:
Ahh ok, didn't realise you could view a previous year's scoreboard, that would make more sense now. But yes, seems like behaviour we perhaps don't need anymore if we are setting score to zero if we don't want it to count...

Anyways, will merge in. It's been useful to discuss this. Thanks!

PR #4204, opened 2025-01-28
Missing data report

tbhi:


db/migrate/20250128164211_add_missing_reading_window_to_amr_data_feed_config.rb
```diff
@@ -0,0 +1,6 @@
+class AddMissingReadingWindowToAmrDataFeedConfig < ActiveRecord::Migration[7.1]
+  def change
+    add_column :amr_data_feed_configs, :missing_reading_window, :integer
+    change_column_default :amr_data_feed_configs, :missing_reading_window, from: nil, to: 5
```
ldodds, line 4: Can you not just add `null: true, default: 5` to the `add_column` statement?

spec/mailers/previews/admin_mailer_preview.rb
```diff
@@ -2,4 +2,9 @@ class AdminMailerPreview < ActionMailer::Preview
   def issues_report
     AdminMailer.with(email_address: 'test@blah.com', user: User.admin.last).issues_report
   end
+
+  def missing_reading_window
+    missing = [[AmrDataFeedConfig.where.not(missing_reading_window: nil).first, 10.days]]
```
ldodds, line 7: Minor, but after initially running migration all of the configs have a null window so there's an error here

ldodds (approved):

db/migrate/20250128164211_add_missing_reading_window_to_amr_data_feed_config.rb
tbhi: am thinking it's better to set all the existing ones to nil and then have 5 as the default for any new ones?

PR #4205, opened 2025-01-29
Fix test flake

urbanwide:


spec/support/shared_examples/recommendations_service.rb
```diff
@@ -208,7 +208,7 @@
       end
 
       it 'tops up from them' do
-        expect(energy_use).to eq(audit_task_types.take(5))
+        expect(audit_task_types).to include(*energy_use)
```
tbhi, line 211: would something like `expect(energy_use).to all(be_in(audit_task_types))` work to preserve `expect(actual)`?
is it an ordering issue?

urbanwide: Didn't know you could do that, thank you!
Yes, the order isn't specified (or required, otherwise would add it to the code), I think that's why it is flakey

tbhi (approved):

PR #4209, opened 2025-01-30
Home page feature and placeholder

urbanwide:
This does a number of things:
* Creates a new feature flag :new_home_page and allows the new home to be switched on by it
* There is a new home page in place but it's only got a few html blocks on the page from the design - it's not component-ised or translated yet, just so we had something to look at as a starter. You'll notice I picked a random photo from the repository until we have a better one.
* Moved the classes used by the current home page (and other pages using the home.html.erb layout) to be scoped to .home-page - and all other pages apart from the home page use these still. The new home page has it's classes scoped to .home.
* Tidied up some css and also removed some classes that are no longer used.
* Took the plunge and removed / commented out the css that changes the header text sizes for different viewport sizes. See what you think.
* Changed some header margins to be more consistent
* Removes the last line that refers to the old feature flag: LOCALE_SWITCHER_BUTTONS
* Changed the navigation-margin to be very slightly bigger so the page doesn't disappear underneath top nav.

### DIFF
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -38,6 +38,11 @@ p, .trix-content div {
   background-image: linear-gradient(to right, $blue-light 0%, $blue-pale 100%);
 }
 
+.fa-li.fa-check {
+  color: $positive-dark;
+  padding-bottom: 100px;
+}
+
 div.padded-row {
   padding-top: 30px;
   padding-bottom: 30px;
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -18,7 +18,7 @@ address {
   font-size: initial;
 }
 
-p {
+p, .trix-content div {
   @include font(f6);
 }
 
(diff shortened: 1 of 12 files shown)


urbanwide:
Thanks for the feedback Leigh. 

Hero: If you look at the design of the Hero, the image breaks out of the container to be flush right with the side of the page but the content on the left hand side, is lined up inside with the container (as is the logo in the nav etc). This is what I tried to achieve with the alignment as I have it at the moment. So I guess this would mean a big gap to the left, if the surrounding container has one too (as your screen is displaying).
Screenshot 2025-02-04 at 15 39 35
I have just made a change to make the font and buttons bigger on larger screens as a starting point but will have a think on what to do about the alignment as above, maybe break out to the left on larger screens...

Scopes: I've made the scopes as a way of stopping the old styles from interfering with the new ones, definitely hope to put more specific styles into components as they arise.

Header top/bottom margins: I've just nudged the header margins a bit so the switch from h1 to h6 is now more gradual, it doesn't look too different to what is there now tbh but is more consistent than it was before.

Will follow up on the other points tomorrow...

urbanwide:
> I believe there's also a few places where we're forcing headings in trix content to be styled correctly? IIRC I overrode some stylings for some of the new components. Not sure if those are needed or not still.

The equivalence component does some overriding of the trix-content h1 size generally, and also overrides the div and trix-content div for public displays:

```
.equivalence-component.vertical {
  .trix-content {
    h1 {
      font-size: $f3;
      line-height: 34px;
      padding-bottom: 10px;
    }
  }
}

// Used for public display version of equivalence
// Force the text to match the header height and colours
// so more readable from a distance
.equivalence-component.public-displays.vertical {
  .trix-content div, div {
    @include font(f1);
    font-family: $header-font-family;
    font-weight: $header-font-weight;
    color: $blue-very-dark;
    padding-bottom: 10px;
  }
  .illustration-wrapper {
    img {
      height: 500px;
    }
  }
}
```
On production, a small screen looks like this (non-public display):
Screenshot 2025-02-05 at 11 09 57

In this branch, it looks like this (heading sizes untouched):
Screenshot 2025-02-05 at 11 09 49

The public-display version looks the same at different sizes.

---

The dashboard also overrides .trix-content but now looks better without different sizes, it was like this before:
Screenshot 2025-02-05 at 11 35 37


--- 
Advice pages headings are bigger now on smaller viewports:

New:
Screenshot 2025-02-05 at 11 39 02

Old:
Screenshot 2025-02-05 at 11 40 38

Perhaps we can chat about this more in our 1-2-1 tomorrow if it's going ahead?

PR #4230, opened 2025-02-05
Update mailchimp when email address is changed

ldodds:
Mailchimp uses email address as its primary key.

It is possible to update the email address of a contact if the existing address is used to make the API call, but the new one is passed in as part of the update.

This PR:

- adds a callback to User so that we trigger an immediate update in Mailchimp (via a job) if a users email address is changed
- removes email address as a watched field, as no longer necessary
- revises the AudienceManager to support passing in old email address
- fixes a loop hole in the status checker so that if email addresses arechanged in mailchimp, we remove the statuses on our side


app/jobs/mailchimp/email_updater_job.rb
```diff
@@ -0,0 +1,11 @@
+module Mailchimp
+  class EmailUpdaterJob < ApplicationJob
+    def perform(user:, original_email:)
+      contact = Mailchimp::Contact.from_user(user)
+      Mailchimp::AudienceManager.update_contact(contact, original_email)
+    rescue => e
+      Rollbar.log(e, job: :audience_updater)
```
tbhi, line 7: is this necessary as Log.exception calls rollbar too?

spec/models/user_spec.rb
```diff
@@ -645,4 +644,34 @@
...
+          expect(Mailchimp::EmailUpdaterJob).not_to receive(:perform_later)
+          user.update!(email: 'new@example.org')
+        end
+      end
+
+      context 'when user is in mailchimp' do
+        let(:user) { create(:user, email: email, mailchimp_status: :subscribed) }
+
+        it 'updates mailchimp' do
+          expect(Mailchimp::EmailUpdaterJob).to receive(:perform_later).with(user: user, original_email: email)
```
tbhi, line 671: I would have preferred the test to have run the job then check the user rather than mock

tbhi (approved):

app/jobs/mailchimp/email_updater_job.rb
ldodds: Should have paid closer attention to what `Log.exception` was doing. Removed duplicates.

spec/models/user_spec.rb
ldodds: Not sure if this is what you meant but have updated the spec. Am now:

- allowing the job to run, but stubbing the Mailchimp API instead
- have changed the job to set the `mailchimp_updated_at` date on the user, which is probably should have been anyway
- checking that the timestamp has been updated after the job runs

PR #4233, opened 2025-02-06
Build grid component

urbanwide:
Ready for a look. Some components may need further refinement, but leaving as is for now.

This introduces some new components. The main grid component, and some element and card components.
Element components are basic ones such as image, paragraph, button, header and block.
Card components tend to have slots containing element components.

The GridComponent can have many 'cells', which can be currently of types Cards::Stats, Cards::Feature, PromptList, Elements::Image, Icon and Elements::Block. I expect other types to be added here.

Use of this setup can be seen on the home page:
home/index.html.erb

I expect to build more or extend existing cards for the other home page content that needs adding.

See what you think @ldodds, thanks 🙏


app/views/home/index.html.erb
```diff
@@ -1,67 +1,53 @@
...
-
-    <div class="row">
-      <div class="col-xl-3 col-sm-6 col-12 pb-4">
-        <div class='rounded text-center bg-light p-4 h-100'>
-          <h4>Save energy</h4>
-          <h3>10%</h3>
-          <p class='small'>Typical energy and carbon savings seen by schools in their first year with Energy Sparks.</p>
-        </div>
-      </div>
+    <%= render Cards::StatementComponent.new do |statment| %>
```
ldodds, line 19: typo, should be statement?

urbanwide:
Yes that makes sense, thanks for the review.

Was also in two minds about html element based components but thought for the purposes of using them for slots, it’s quite handy. Definitely not wanting to use them for everything...

Agree it might be a good idea to use render in some of them, rather than a template. Good idea.

In terms of your last comment on adding a layout module, do you mean like this:

```
Layout/GridComponent
Layout/Cards/StatsComponent
Layout/Cards/FeatureComponent etc
```

Or flatter like this?:

```
Layout/GridComponent
Layout/StatsComponent
Layout/FeatureComponent etc
```

I agree with you on the home page hero. I thought it best to keep it as close to the design as possible for now so it could be discussed - or happy to try some alternatives?

ldodds:
For the Layout component, I think either works. I'd use the nested style if we're going to have more Cards or if there are other (possibly existing) components that might also be grouped.

re: home page. Maybe build out all the rest of the page, then we can come back to that. Wondering if we're going to have similar issue on the other marketing type pages as well.

PR #4239, opened 2025-02-10
Allow clearing of school.bill_request

tbhi:


ldodds:
Its fine like this. Generally they're sending a single request and want to clear that, rather than tracking requests for specific meters.

ldodds (approved):

PR #4241, opened 2025-02-10
add school urn validation to show error when it exceeds database limit

tbhi:


app/models/school.rb
```diff
@@ -283,7 +283,8 @@ def self.engaged(date_range)
   end
 
   validates :name, :address, :postcode, :website, :school_type, presence: true
-  validates :urn, presence: true, uniqueness: true
+  validates :urn, presence: true, uniqueness: true,
+                  numericality: { greater_than_or_equal_to: 10_000, less_than_or_equal_to: (2**31) - 1 }
```
ldodds, line 287: Checking the database, some schools (around 30) are already in the system with URNs that are invalid according to the lower bound here. Looks like some have been setup with a DfE reference, e.g. 7008 rather than a URN. So we can't make those invalid. 

A couple have single digit numbers. I think this is where the team have set an invalid URN to allow a school that has previously been with us to be onboarded again.

So we need to loosen the restriction here to not make any existing schools invalid. 

Where the URN is too large the validation catches the problem but it doesn't really give the user a lot of feedback ("must be less than or equal to 2147483647"). Could we display a more bespoke message if this validation has failed? E.g. "The URN or SEED you have supplied appears to be invalid"?

tbhi: would it be worth correcting the first few and have a minimum allowing the rest?

ldodds: Suggest we maybe do that as a separate exercise and just handle the integer too large issue for now.

Later in the year I want to look at importing the DfE school database, so rationalising ids would be a good thing to tackle as part of that.

ldodds (approved):

PR #4251, opened 2025-02-12
New user profile pages

ldodds:
This PR adds a new `/profiles` route to allow users to manage their own accounts and preferences.

I've got some additional improvement to make (e.g. managing Mailchimp preferences, some admin options) but this feature covers the basics, so want to get this reviewed first.

There's a feature flag which toggles adding a user icon to the second nav which links to the pages.

Included some redacted screenshots below.

### Account page

Screenshot from 2025-02-13 15-26-31

### Account edit

Screenshot from 2025-02-13 15-26-44

### Password update

Screenshot from 2025-02-13 15-26-53

### Alert management

Screenshot from 2025-02-13 15-27-02

### DIFF
--- a/app/controllers/users/contacts_controller.rb
+++ b/app/controllers/users/contacts_controller.rb
@@ -1,0 +1,40 @@
+module Users
+  class ContactsController < ApplicationController
+    load_resource :user
+    load_and_authorize_resource through: :user
+
+    def index
+      @schools = if @user.group_admin?
+                   @user.school_group.schools.visible.by_name
+                 elsif @user.has_other_schools?
+                   @user.cluster_schools.by_name
+                 else
+                   [@user.school]
+                 end
+      @show_clusters = @schools.any? { |s| s.school_group_cluster.present? }
+      render :index, layout: 'dashboards'
+    end
+
+    # Sometimes creating the contact is failing as its invalid...?
+    # trying to touch! the user model when inserting.
+    def create
+      @contact = Contact.create(
+        user_id: @user.id,
+        school_id: params[:school_id],
+        name: @user.display_name,
+        email_address: @user.email,
+        staff_role: @user.staff_role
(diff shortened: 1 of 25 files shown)


urbanwide:
I have a quick question about the url - it feels like it should be /profile for your own profile?

Second question would be the left nav - 'My account' links to profile 'dashboard', 'Change password' links to change password form but there is no link to the 'Update account' page. It feels like it should have one to me, especially as there is one for Change password.

Looking more at the code now.

ldodds:
Rails paths are plural by default. Or are you suggesting a `/profile` link that would redirect? I want to keep the `/profiles` path so that admins can access all the profiles.

I can see what you mean about "Update account". I was in two minds about adding one. I can add that in and an extra profile home link.

ldodds:
I am planning on adding a redirect link that can be targeted from alerts/help pages, etc.

app/controllers/users/contacts_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+                 elsif @user.has_other_schools?
+                   @user.cluster_schools.by_name
+                 else
+                   [@user.school]
+                 end
+      @show_clusters = @schools.any? { |s| s.school_group_cluster.present? }
+      render :index, layout: 'dashboards'
+    end
+
+    # Sometimes creating the contact is failing as its invalid...?
```
urbanwide, line 18: Do we know why creating is failing? Seems like something that needs ironing out? Is that why you are using Contact.create here and then saving a second time?

urbanwide (approved):
Looks good - a couple of comments / questions

app/controllers/users/contacts_controller.rb
ldodds: Will revisit this later, think its related to the active record callbacks

PR #4252, opened 2025-02-12
Try lookbook

urbanwide:
Can now be found at `/admin/components` and also in the admin index

Turns out I'd put the gem in the in the `group :development, :test` rather than just the `group :development` as recommended. It's now in the `group :development, :production` so we can use it on the live boxes.

I've also changed the name of the transient variable `log` in the meters factory (and tested it locally), in case we do want to switch this on in test for whatever reason (docs don't mention doing this though).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -131,7 +131,6 @@ group :development, :test do
   gem 'guard-rspec', require: false
   gem 'guard-rubocop', require: false
   gem 'knapsack'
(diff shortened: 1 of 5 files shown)


ldodds (approved):
Can you hold off on merging until after the release and then put on to test tomorrow? That `log` variable issue makes me suspicious so want it on test for a few days before we push it live.

PR #4254, opened 2025-02-14
Add additional flag to meter to indicate if its on manual reads

tbhi:


app/controllers/admin/reports/manual_reads_controller.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+module Admin
+  module Reports
+    class ManualReadsController < AdminController
+      include ActionView::Helpers::UrlHelper
+      include ApplicationHelper
+
+      class Column
```
ldodds, line 9: Probably worth extracting this out at some point e.g. in app/classes. Have been wondering how we can simplify creating the many different CSV/HTML reports we're producing.

ldodds (approved):

PR #4277, opened 2025-02-20
Improve automated conversion to kwh when loading data

tbhi:
wonder if it would be better to make convert_to_kwh an enum rather than the extra boolean


lib/tasks/deployment/20250224143128_improve_automated_conversion_to_kwh.rake
ldodds, line 1: We also need to find each of these meters, identify any scaling meter attribute that is currently in use (so not replaced or deleted) and that has an open end date. The end date needs to be replaced with the current date. This will ensure that any loading after this release will not have the scaling applied twice (once during loading and once as a validation).

PR #4279, opened 2025-02-20
Add a review page to highlight potential issues with school configuration before its made data enabled.

ldodds:
Adds a new school specific page that attempts to highlight potential issues with the school configuration.

The page is available to admins from the Manage School menu and can be used to check configuration before and after a school is made data visible. The initial set of checks are based on some recent issues and may be extended over time.

The "data visible" toggle button now take the admin to this page instead, when enabling a school, so they can review setup before making the data visible on the site. Making a school no longer data enabled remains the same.


ldodds:
@urbanwide this is now done, can you give it a review please?

urbanwide (approved):

PR #4285, opened 2025-02-24
Fix organic contact interests

ldodds:
Fixes an issue with creating the list of emails associated with each contact in the CSV export.

Internally we manage the list of interests as a hash of name/id to true/false for each preference. But this has to be turned into a comma-separated list for the CSV export.

The interests for "organic" contacts was not being converted to a string. Bug crept through because Hash responds to `include`. So have added a check for the type of object as well as its contents.

### DIFF
--- a/app/services/mailchimp/csv_exporter.rb
+++ b/app/services/mailchimp/csv_exporter.rb
@@ -133,14 +133,8 @@ module Mailchimp
       contact.locale = 'en'
       contact.tags = non_fsm_tags(existing_contact).join(',')
 
-      # Turn existing interests into a hash to make it easier to set
-      # defaults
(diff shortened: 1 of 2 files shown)


app/services/mailchimp/csv_exporter.rb
```diff
@@ -133,17 +133,24 @@ def copy_contact(existing_contact, add_default_interests: false)
       contact.locale = 'en'
       contact.tags = non_fsm_tags(existing_contact).join(',')
 
+      # Turn existing interests into a hash to make it easier to set
+      # defaults
       interests = if existing_contact[:interests].present?
```
tbhi, line 138: `(existing_contact&.dig(:interests)&.split(',') || []).index_with { true }`

PR #4296, opened 2025-02-25
Replace use of component helper

urbanwide:
Replace the use of the component helper.

This highlights the extent to which we are using components - it's a big PR! I've done a fairly deep review of the changes myself but will benefit from another review. Thanks


ldodds (approved):
Did some spot checks across the files, but haven't checked every line. So relying on the tests to catch any problems.

PR #4299, opened 2025-02-27
Add carousel testimonials section

urbanwide:


config/locales/views/home/home.yml
```diff
@@ -53,7 +53,19 @@ en:
         header: Empower pupils
         subtext: Number of participating schools using Energy Sparks with their eco-team to inform impactful action
     sub_title: A unique school-specific energy management tool and education programme
+    testimonials:
```
ldodds, line 56: Aside: we'll want to move these to some content management eventually. Need to decide whether to add that once the homepage is done or after implementing the other pages.

app/assets/stylesheets/headers.scss
```diff
@@ -61,121 +61,149 @@ h1, h2, h3, h4, h5, h6 {
...
-    padding-bottom: 5px;
-  }
+@mixin responsive-typography {
+  @include responsive-typography-scale-up;
+}
 
-  h2 {
-    @include font(f5);
-  }
+@mixin responsive-typography-scale-up {
```
ldodds, line 68: Sent you screenshot but this is scaling up too much at least for the carousel on the homepage. Thinking its only the hero text on homepage where we might scale the typography as rest of the site works for with h1-h6 as is.

app/components/application_component.rb
```diff
@@ -4,9 +4,22 @@ class ApplicationComponent < ViewComponent::Base
 
   attr_reader :id, :classes
 
-  def initialize(id: nil, classes: '')
+  # Structuring the initialize method in this manner offers flexibility for future enhancements
```
ldodds, line 7: :+1:

app/components/notice_component.rb
```diff
@@ -1,26 +1,21 @@
...
 
-  def initialize(status:, classes: nil, style: :normal)
+  def initialize(status:, style: :normal, **_kwargs)
+    super
     @status = status
-    @classes = classes
     @style = style
     validate
+    add_classes(status)
+    add_classes(@style == :compact ? ' p-3' : ' p-4')
```
ldodds, line 12: Much nicer!

app/components/school_comparison_component/school_comparison_component.html.erb
```diff
@@ -1,6 +1,4 @@
-<div
-  id="<%= @id %>"
-  class="school-comparison-component-callout-row row mt-2">
```
ldodds, line 3: This breaks the layout as the naming convention here is different (can't remember why) so the styles in the component need updating. See, e.g. baseload insights page. 

The group version is OK I think as the `-callout-row` class is not on the outer element.

Highlights benefit of previews as sometimes easier to check the previews rather than the pages where component is being used. Don't think we have them for these components.

ldodds (changes requested):
Need to fix the school comparison component.

I think the equivalences on pupil dashboard are OK, but they're using `.dashboard-equivalences-component` for the javascript. See carousels.js

config/locales/views/home/home.yml
urbanwide: You were talking about moving the yaml to the database a bit ago. Is this what you mean?

app/assets/stylesheets/headers.scss
urbanwide: Ok, will remove the scaling here. Always interesting seeing how things look on other people's screens!

config/locales/views/home/home.yml
ldodds: No, I mean that we should have a Testimonials model (or similar) with admin forms so they can be managed in the app, as we do with Case Studies. The model would sync to transifex.

app/components/school_comparison_component/school_comparison_component.html.erb
urbanwide: This is a good spot, thanks. I've wrapped the component in a div to put the classname in, and renamed the classes used to have more concise names, scoped to this classname. Hope this is ok.

I've also added basic previews for this and the school group comparison component

ldodds (approved):

PR #3276, opened 2023-12-07
Chart preview form

ldodds:
Provides a simple form for previewing charts using a school's current data. Intended to help support debugging and reviewing chart configurations. 

Allows admin to pick any data enabled school and any chart configuration. Also allows tweaking of component parameters.

Plenty of improvements that could be made but useful in its current form.

PR #4308, opened 2025-03-06
next school button for easier checking pages of different schools

tbhi:
anything like this already? any other options?

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -860,6 +860,10 @@ class School < ApplicationRecord
     end
   end
 
+  def next_slug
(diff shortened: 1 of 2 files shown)


ldodds:
When I'm testing, e.g. charting I'm usually look for schools with different volumes of data, e.g. more or less than a year, two years, with and without the needed fuel types, etc.

Might be better off with a way to preview specific charts for specific schools. I've got a draft open PR to do that which I can revive.

tbhi:
replaced by https://github.com/Energy-Sparks/energy-sparks/pull/3276

PR #4309, opened 2025-03-06
new analytics with various chart fixes

tbhi:
- [x] fix end date on academic year


ldodds (approved):

app/services/date_service.rb
```diff
@@ -4,4 +4,14 @@ module DateService
   def self.fixed_academic_year_end(date)
     Date.new(date.year + (date.month < 9 ? 0 : 1), 8, 31)
   end
+
+  def self.subtitle_end_date(chart_type, date)
```
ldodds, line 8: Possible alternative to hard-coding: the `timescale` in the chart configuration can be used to check whether the series use `fixed_academic_year`.

`ChartDataValues` has access to the configuration as its returned from the analytics within `chart[:configuration]` but the `MeterSelectionChartComponent` would need to look it up using `ChartManager`

ldodds (approved):

PR #4314, opened 2025-03-11
Tools section

urbanwide:


config/locales/views/home/home.yml
```diff
@@ -64,6 +64,19 @@ en:
       read_case_study: Read case study
     title: We help schools become more energy efficient and fight climate change.
     tools:
+      cards:
+      - badge: Make a differece
```
ldodds, line 68: Typo

app/components/layout/cards/tools_component.rb
```diff
@@ -0,0 +1,12 @@
+module Layout
+  module Cards
+    class ToolsComponent < ApplicationComponent
```
ldodds, line 3: Nitpick: "Tools" seems like an unusual name for the component and the section. The text isn't really about tools we offer? I know is says "Data-driven tools" in the heading above, but I don't think what we're listing are tools, or that the component needs to be limited to that.

Could we use something more descriptive. E.g. a StatementComponent? Or BadgedStatementComponent? Or ProductFeatureComponent?

ldodds (changes requested):

app/components/layout/cards/tools_component.rb
urbanwide: Makes sense, thanks

urbanwide: Have renamed StatementComponent to SectionHeader (which made more sense anyway), so I could then rename ToolsComponent to StatementComponent. I've renamed the tools yaml section to features

ldodds (approved):

PR #4327, opened 2025-03-11
Add audits section to front page

urbanwide:


config/locales/views/home/home.yml
```diff
@@ -5,6 +5,9 @@ en:
     more_case_studies: More case studies
     title: Case Studies
   home:
+    audits:
```
ldodds, line 8: The original wireframes had this section as being about audits, but the new content is just about Energy Sparks in general rather than audits so might want to rename

app/views/home/index.html.erb
```diff
@@ -89,6 +89,19 @@
...
+
+    <%= render Layout::GridComponent.new(
+          cols: 2,
+          classes: 'bg-blue-light rounded-12 p-4 mb-4',
+          cell_classes: 'my-auto'
+        ) do |grid| %>
+      <%= grid.with_feature_card do |feature| %>
+        <%= feature.with_header(title: t('home.audits.title'), level: 3) %>
+        <%= feature.with_description { t('home.audits.description_html', count: School.visible.count) } %>
+        <%= feature.with_button(t('home.enrol_your_school'), 'energy-audits', style: :primary) %>
```
ldodds, line 101: not sure this is the intended final link. Can you update to /find-out-more for now?

ldodds (approved):

config/locales/views/home/home.yml
urbanwide: This is a fair point, thanks

PR #4338, opened 2025-03-17
Remove volunteer role

ldodds:
Removes the volunteer role as its no longer needed. Volunteers are flagged as staff users with a staff role of "Parent or Volunteer".

This simplifies some of the code and CanCanCan abilities.


ldodds:
> Quick question (not done a full review yet) - I have a couple of users with the volunteer role in my local database, have these been changed on live already or do they need migrating?

Released yesterday: https://github.com/Energy-Sparks/energy-sparks/pull/4336

urbanwide (approved):

PR #4339, opened 2025-03-18
Add organisations section to front page

urbanwide:


app/components/application_component.rb
```diff
@@ -16,13 +16,53 @@ class ApplicationComponent < ViewComponent::Base
...
+    add_theme(theme)
   end
 
   def add_classes(classes)
     @classes = class_names(@classes, classes)
   end
+
+  # may put this in a card component base class but want to avoid too many levels of inheritance if possible
+  # tempted to cascade the theme down
+  def add_theme(theme)
```
ldodds, line 32: as not all (most?) of the components won't be themable in quite this way, would this be better in a mixin if you don't want to add a base class for cards?

app/components/layout/cards/feature_component/feature_component.scss
```diff
@@ -1,9 +1,8 @@
 .layout-cards-feature-component {
-  &.responsive {
+  &.main {
```
ldodds, line 2: as this impacts the design, worth adding a separate preview?

ldodds (changes requested):
Left a couple questions, but I think we need to sort out the heights of the cards. Because of the content differences, they all have different heights on different screen sizes. Might need a min height or something?

ldodds:
Screenshots of heights

Screenshot from 2025-03-18 11-14-12
Screenshot from 2025-03-18 11-13-50

urbanwide:
Sorry @ldodds , this should have been draft - not finished sorting the heights out yet or previews, need some spec too.

urbanwide:
Just noticed my rounded corners have gone, just looking why now 😬

ldodds (approved):
Looks good, love the previews

PR #4340, opened 2025-03-18
Rework existing Trix extensions and styling into a component

ldodds:
We currently have several customisations for the Trix editor, all implemented slightly differently:

- the user facing editor for recording activities and actions has the code and strike-though options removed. This is done in CSS (`display: none`), via a wrapper class
- the user facing editor has larger buttons, again handled via CSS
- the url field is customised via a trix-initialise callback
- for the activity type form, we add a custom chart button. This is currently done by embedding the button and trix dialog content into page using a partial, adding a class to the surrounding trix element, and some Javascript to update the Trix editor

This PR removes this inconsistency by:

- adding a Forms::TrixComponent component which provides a cleaner interface for selecting between customisations, e.g. `controls: :simple` or `controls: :advanced`
- rationalises the CSS, as there was some old/unneeded classes
- moves hiding the code/strikethrough buttons into Javascript, removing the buttons directly from the UX rather than just hiding them
- moves creation of the chart button into JS. The list of charts is now set as a data attribute on the `div.forms-trix-component` wrapper. This removes the need to add a wrapper class and include HTML in the page. The Javascript contains the button and dialog templates
- reworks the Javascript adding a `TrixCustomiser` class to encapsulate working with Trix a bit more
- adds a revised Heading button which replaces the standard H1 option with a dialog that lets users select from H1-H6

The component has initially just been added to the Activity, ActivityType forms as these are where we are using customised editors. As well as the new Cms::Section forms. Spec helpers have been added to confirm the right versions of the editors are being displayed (this wasn't tested before).

Elsewhere the admin interface for now just uses the default options. If we just need a basic Trix component then we can just use `f.rich_text_field` and then use the component is there's any customisations.

### DIFF
--- a/app/assets/javascripts/trix_customisations.js
+++ b/app/assets/javascripts/trix_customisations.js
@@ -11,7 +11,7 @@ function addHeadingAttributes() {
 }
 
 addEventListener("trix-initialize", event => {
-  var customiser = new TrixCustomiser(event.target)
+  const customiser = new TrixCustomiser(event.target)
   customiser.addUrlValidation()
 
   var components = $(event.target).parents('.forms-trix-component');
@@ -40,7 +40,6 @@ document.addEventListener("trix-action-invoke", function(event) {
 
 class TrixCustomiser {
   constructor(element) {
-    console.log(element.toolbarElement)
     this.element = element
   }
 
--- a/app/components/forms/trix_component.rb
+++ b/app/components/forms/trix_component.rb
@@ -19,7 +19,7 @@ module Forms
     end
 
     def data_attributes
-      return { chart_list: charts } if charts
+      { chart_list: charts } if charts
     end
 
     class << self


app/views/admin/activity_types/_form.html.erb
```diff
@@ -46,31 +44,48 @@
...
       <%= f.label ActivityType.human_attribute_name(:school_specific_description) %>
-      <%= render 'admin/shared/preview_section', scope: "school-specific-description", preview_url: admin_activity_type_preview_path, no_timings: true, no_weightings: true, locales: [locale] do |locale| %>
-        <%= f.rich_text_area t_field(:school_specific_description, locale), class: 'chart-list' %>
+      <%= render 'admin/shared/preview_section',
+                 scope: 'school-specific-description',
+                 preview_url: admin_activity_type_preview_path,
+                 no_timings: true,
+                 no_weightings: true,
+                 locales: [locale] do |locale| %>
+                  <%= render(Forms::TrixComponent.new(form: f, field: t_field(:school_specific_description, locale),
```
ldodds, line 64: This is the only functional change here, to use the new component.

tbhi (approved):

tbhi:
wonder if chart button is a bit small?
image

PR #4341, opened 2025-03-20
Use LDZ calorific values during gas meter conversions

tbhi:


ldodds (changes requested):

ldodds (approved):

ldodds:
Revised calculations look fine, but some specs failing still

PR #4342, opened 2025-03-21
Front page feedback

urbanwide:
A few changes based on the feedback from Sarah & Claudia

This is based on https://github.com/Energy-Sparks/energy-sparks/pull/4339/ which will need reviewing / merging first


app/components/elements/image_component.rb
```diff
@@ -22,8 +23,15 @@ def setup_classes
       add_classes('d-none d-md-block') if @collapse
     end
 
+    def style
+      style = ''
+      style += "width: #{@width};" if @width
+      style += " height: #{@height};" if @height
```
urbanwide, line 29: Didn't end up using these changes but left them here in case they are needed later.

PR #4343, opened 2025-03-24
Set role when adding existing user as school admin

ldodds:
When we add an existing user as a school admin, we're not currently setting the role correctly. The school is just added to the users list of schools, but their role is not updated.

Fix the issue in the controller and add a database fix to correct around 20 records.

### DIFF
--- a/spec/system/schools/users_spec.rb
+++ b/spec/system/schools/users_spec.rb
@@ -346,6 +346,20 @@ describe 'School admin user management' do
           expect(other_school_admin.cluster_schools_for_switching).to eq([school])
         end
 
+        context 'when other user is staff' do
+          let!(:other_school_admin) { create(:staff, name: 'Other admin') }
+
+          it 'adds the user' do
+            click_on 'Add an existing Energy Sparks user as a school admin'
+            fill_in 'Email', with: other_school_admin.email
+            click_on 'Add user'
+            expect(page).to have_content(other_school_admin.name)
+            other_school_admin.reload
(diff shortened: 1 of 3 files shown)


spec/system/schools/users_spec.rb
```diff
@@ -346,6 +346,20 @@
           expect(other_school_admin.cluster_schools_for_switching).to eq([school])
         end
 
+        context 'when other user is staff' do
+          let!(:other_school_admin) { create(:staff, name: 'Other admin') }
+
+          it 'adds the user' do
+            click_on 'Add an existing Energy Sparks user as a school admin'
+            fill_in 'Email', with: other_school_admin.email
+            click_on 'Add user'
```
urbanwide, line 355: I would split this into a before block and at least one test here as follows:

```
before do
  click_on 'Add an existing Energy Sparks user as a school admin'
  fill_in 'Email', with: other_school_admin.email
  click_on 'Add user'
end

it 'adds the user' do
  expect(page).to have_content(other_school_admin.name)
  other_school_admin.reload
  expect(other_school_admin.cluster_schools_for_switching).to eq([school])
  expect(other_school_admin.role).to eq 'school_admin'
end

```

urbanwide (approved):
Looks good - with comment about the spec

spec/system/schools/users_spec.rb
ldodds: I've reworked the spec a bit to reduce overlap between the tests.

PR #4351, opened 2025-03-26
improve energy tariff value validation message

tbhi:
fixed tariff
image
variable
image


tbhi:
had problems with shoulda-matcha on with this when setting the message in the validation on the model as shoulda-matcha is using the error message to filter tests from the different less than and greater than constraints so when the same message is used for both is gets confused about which checks apply to which constraint.

ldodds (approved):

PR #4367, opened 2025-04-02
use puma preloading

tbhi:


ldodds (approved):
We used to have this setting on as it helped when we had less memory on the server, but got [removed in one of the Rails upgrades](https://github.com/Energy-Sparks/energy-sparks/commit/ffa8d6148610044b2d71c174463e6784bd4b88db#diff-24409379bdb75ad446bb4e2c18fd4cced1a263b99a3ff96fc3777d8fd8faeab9)?

PR #4369, opened 2025-04-03
Add blog section to front page

urbanwide:
* I need to fix the rspec for the BlogService (rewrote it without updating spec)
* Need a discussion about the formatting @ldodds : Some content possibly needs chopping or we need to restrict the amount of tags, or move them down under the title.

NB have made the badge text a tad lighter as I feel it's closer to the original design. See what you think.

image


urbanwide:
Some layout options (some with more spacing than others) @ldodds :

I think I favour either (no 2 or 5 for me): 

category, title, date / author
OR
title, date / author, category

1. Screenshot 2025-04-04 at 15 36 50
2. image
3. Screenshot 2025-04-04 at 15 57 36
4. Screenshot 2025-04-04 at 15 37 26
5. Screenshot 2025-04-04 at 15 39 20

app/services/blog_service.rb
```diff
@@ -0,0 +1,101 @@
...
+    cached_feed&.dig(:items)
+  end
+
+  # Can be run from cron if we don't want to tie up the front-page request
+  def update_cache!
+    Rails.cache.write(key, fetch_feed) if fetch_feed?
+  end
+
+  # Can be run in cron to decouple from this tie-ing up the front page request
+  def fetch_feed
```
ldodds, line 64: I think you can use the Faraday Retry middleware to do a fix number of retries with a back-off. We're using that elsewhere. https://github.com/lostisland/faraday-retry

ldodds (approved):
Left some comments about the cache/cron behaviour and using Faraday for the retries. Happy if you want to refactor that as a separate PR.

For the layout options, I like 5 too.

app/services/blog_service.rb
```diff
@@ -0,0 +1,101 @@
...
+    @key = "blog:#{url}"
+    @retries = retries
+    @force_fetch = force_fetch
+  end
+
+  def cached_feed
+    Rails.cache.fetch(key)
+  end
+
+  def cache_out_of_date?(period)
```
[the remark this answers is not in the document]
urbanwide: This is a far simpler way of doing things, thanks. I was doing it the other way as I wanted to keep the old cache entries when they had become expired (for the case when fetching a newer version failed). Had not spotted race_condition_ttl - would have needed something like that on reflection.

Anyways, the new version hopefully matches your suggested implementation 👍🏻

app/services/blog_service.rb
urbanwide: Thanks - I should have thought to check existing ways of doing this in the codebase. Makes things easier, thanks.

urbanwide:
Thanks @ldodds, ready for a re-review if you have a minute, thanks 🙏

app/components/elements/image_component/image_component.scss
```diff
@@ -17,4 +17,10 @@ img.elements-image-component {
     width: 100%;
     height: 100%;
   }
+
+  @include media-breakpoint-up(lg) {
```
ldodds, line 21: can be removed?

ldodds (approved):

PR #4388, opened 2025-04-11
Add search for new support pages

ldodds:
Implements the searching for the `Cms::Section` model

* Refactor and tidy the `Searchable` concern to tidy the query, use better escaping and parameterise some of the assumptions around fields to be searched
* Allow `show_all` parameter in `Searchable.search` to allow searching across unpublished content
* Switch to using `websearch_to_tsquery` rather than `to_tsquery`, as the former parses search terms more like one would expect from a web search. As a side effect this is likely to improve the existing activity and intervention type search as currently multiple keywords are being treated as a phrase search, rather than separate keywords 
* Add route and search form, and a component for handling formatting of search results

### DIFF
--- a/app/models/cms/section.rb
+++ b/app/models/cms/section.rb
@@ -62,8 +62,8 @@ module Cms
       private
 
       def searchable_filter(show_all: false)
-        if show_all == true
-          %|"#{table_name}"."published" in ('true', 'false')|
+        if show_all
+          'true' # include everything
         else
           %|"#{table_name}"."published" = 'true'|
         end
--- a/app/models/cms/section.rb
+++ b/app/models/cms/section.rb
@@ -61,8 +61,12 @@ module Cms
     class << self
       private
 
-      def searchable_filter
-        %|"#{table_name}"."published" = 'true'|
+      def searchable_filter(show_all: false)
+        if show_all == true
+          %|"#{table_name}"."published" in ('true', 'false')|
+        else
+          %|"#{table_name}"."published" = 'true'|
+        end
       end
 
       def searchable_body_field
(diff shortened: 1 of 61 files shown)


app/models/concerns/searchable.rb
```diff
@@ -1,75 +1,90 @@
...
+# searchable_body_field - name of field containing body text
+# searchable_metadata_fields - array of additional field names to query
+# searchable_filter - WHERE clause that filter the matched records, e.g. to those visible to end user
 module Searchable
   extend ActiveSupport::Concern
 
+  # Disable cop as want SQL to be formatted for readability. Its already a complex query and splitting
+  # into fragments will make it less readable.
+  #
+  # rubocop:disable Metrics/BlockLength
```
tbhi, line 12: put the disable comment on the next line to apply to the whole block?

app/models/cms/section.rb
```diff
@@ -56,5 +57,25 @@ def set_default_position
         self.position = max_position ? max_position + 1 : 0
       end
     end
+
+    class << self
+      private
+
+      def searchable_filter(show_all: false)
+        if show_all == true
```
tbhi, line 65: `== true` not needed?

app/models/cms/section.rb
```diff
@@ -56,5 +57,25 @@ def set_default_position
...
+          %|"#{table_name}"."published" in ('true', 'false')|
```
tbhi, line 66: could be simplified to "true"?

tbhi (approved):

app/models/concerns/searchable.rb
ldodds: the disable only works in this position?

PR #4396, opened 2025-04-15
Add Transifex Synching for CMS module

ldodds:
Include the `TransifexSynchronisable` in the `Cms::Base` class and implement `tx_resources` to only synch published content.

Noticed that the `Scoreboard` and `Testimonial` classes were not in the loader so have added those.

### DIFF
--- a/app/models/testimonial.rb
+++ b/app/models/testimonial.rb
@@ -31,8 +31,5 @@ class Testimonial < ApplicationRecord
   validates :image, :title_en, :name, :quote_en, :organisation, :category, presence: true
 
   scope :active, -> { where(active: true) }
-
-  def self.tx_resources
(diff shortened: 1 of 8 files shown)


app/models/testimonial.rb
```diff
@@ -31,4 +31,8 @@ class Testimonial < ApplicationRecord
   validates :image, :title_en, :name, :quote_en, :organisation, :category, presence: true
 
   scope :active, -> { where(active: true) }
+
+  def self.tx_resources
+    active.order(:id)
+  end
```
tbhi, line 37: would `scope :tx_resources, -> { active.order(:id) }` be better?

tbhi (approved):

PR #4400, opened 2025-04-16
Add admin controls to support pages

ldodds:
Add admin controls to the support page views, to make it easier to review, edit and publish content.

I've added a new component to handle display of a panel of admin buttons with an optional status.


app/views/cms/pages/_section_admin_buttons.html.erb
urbanwide, line 1: This looks very similar to app/views/cms/pages/_admin_buttons.html.erb, could they share the same partial but with a parameter? Same for the categories one?

urbanwide (approved):
Looks good - made a couple of suggestions 👍

app/views/cms/pages/_section_admin_buttons.html.erb
ldodds: Reworked this to use a single partial with polymorphic paths and conditional sections :+1:

PR #4402, opened 2025-04-17
Refactor advice pages to support view caching

ldodds:
The advice pages currently rely on loading the MeterCollection from the Rails cache. 

This is currently used:

- in several `before_action` to check, e.g. whether date is up to date, whether the page can be shown, etc
- to calculate variables for rendering the page

All of that currently happens before the views which makes it difficult to introduce caching. Adding view caching means that the MeterCollection will still be loaded and calculations run.

This PR refactors the advice page framework:

- rework the before_action code, so that its less reliant on the MeterCollection in a number of places. E.g. some of the key data is available via the `School:Configuration` model)
- reworks the Baseload controller, services and views so that calculations are lazily run (i.e. when the are first accessed in views, with memoized results) rather than being called from the controller actions, so we pass a service to the view rather than variables.

This move things into the direction of adding caching of templates and/or partials.

Other controllers and services will also need refactoring.

To come in later PRs:

* actually introducing caching, starting with baseload page
* working through other controllers, services and views to refactor and introduce caching. Starting with heating control and out of hours pages
* rework other before actions, e.g. to lazily load and/or cache alerts or other objects used in pages.
* revisit some of the date logic in AnalysisDates as the last full week methods don't always seem to produce correct dates, so reliant on charts loading.


app/controllers/schools/advice/baseload_controller.rb
```diff
@@ -42,37 +43,28 @@ def aggregate_meter
       end
 
       def set_economic_tariffs_change_caveats
-        @economic_tariffs_change_caveats = build_economic_tariffs_change_caveats
```
urbanwide, line 45: Could you initialise the service here i.e.
```
@economic_tariffs_change_caveats = Costs::EconomicTariffsChangeCaveatsService.new(
          meter_collection: aggregate_school, fuel_type: @advice_page.fuel_type.to_sym
        )
```
      
Then call the service in the method on the service in the view, if it's not nil / been set? i.e. in app/views/schools/advice/_how_have_we_analysed_your_data.html.erb

```
<% if (caveats = @economic_tariffs_change_caveats&.calculate_economic_tariff_changed) %>
  <%= t('advice_pages.how_have_we_analysed_your_data.your_electricity_tariffs_html',
        last_change_date: short_dates(caveats.last_change_date),
        rate_before_gbp_per_kwh: format_unit(caveats.rate_before_£_per_kwh, :£),
        rate_after_gbp_per_kwh: format_unit(caveats.rate_after_£_per_kwh, :£),
        percent_change: format_unit(caveats.percent_change, :percent)) %>
<% end %>
```

urbanwide (commented):
Looks like a good approach (although I can't claim to be an expert on caching but I can see where you are going with it). Made a few comments but apologies if I have the wrong end of the stick!

app/views/schools/advice/advice_base/insights.html.erb
```diff
@@ -1,24 +1,28 @@
-<%= render 'schools/advice/advice_page', school: @school, advice_pages: @advice_pages, advice_page: @advice_page,
-                                         tab: @tab, data_warning: @data_warning do %>
+<%= cache_if @cache_view, [@school.id, @school.name.parameterize, @tab, @school.validation_cache_key].join('-') do %>
```
tbhi, line 1: sufficient to use `@school`? Think it'll use `@school.updated_at` then.  Believe template name is automatically used in the cache key so don't need @tab?

also wonder if we also want a long (day or week?) expires_in as otherwise not sure cache file would be cleaned up

ldodds: I'm going to revert this change here and introduce a separate PR with caching changes. Need to think a bit more about caching the entire partial or sections, e.g. the nav, _insights, _analysis, etc.

You're right about the keys. I think @school may be sufficient so long as we're updating the record to set the validation_cache_key. Hadn't realised about the template being included.

app/controllers/schools/advice/baseload_controller.rb
ldodds: The service can't be created in the controller as it accepts the meter_collection as a parameter in the constructor. The idea is to defer having any references to the meter collection until the view rendering so everything is lazy loaded. I would have refactored this service like the others, but it's in the analytics and i was trying to keep this self-contained. So going to keep in the view for now.

ldodds:
@tbhi could you do a final look at this and give it a :+1:? I'll add the actual caching to the baseload page in a follow on PR as this is big enough already.

app/classes/schools/analysis_dates.rb
```diff
@@ -0,0 +1,110 @@
...
+      analysis_date > (Time.zone.today - 30.days)
+    end
+
+    def months_of_data
+      ((analysis_end_date - analysis_start_date).to_f / 365 * 12).floor
+    end
+
+    def months_analysed
+      months = months_of_data
+      months > 12 ? 12 : months
```
tbhi, line 77: `[months, 12].min`?

app/services/schools/advice/baseload_service.rb
```diff
@@ -16,6 +17,24 @@ def multiple_electricity_meters?
...
+      def current_baseload
+        @current_baseload ||= begin
+          average_baseload_kw_last_year = average_baseload_kw(period: :year)
+          average_baseload_kw_last_week = average_baseload_kw(period: :week)
+
+          previous_year_average_baseload_kw = previous_period_average_baseload_kw(period: :year)
+
+          previous_week_average_baseload_kw = previous_period_average_baseload_kw(period: :week)
+
+          OpenStruct.new(
```
tbhi, line 29: `ActiveSupport::OrderedOptions.new.merge`?

app/services/schools/advice/baseload_service.rb
```diff
@@ -77,67 +98,77 @@ def academic_year(start_date, end_date)
...
-        calculate_seasonal_variation(aggregate_meter, asof_date)
+        @seasonal_variation ||= calculate_seasonal_variation(aggregate_meter, asof_date)
       end
 
       def seasonal_variation_by_meter
         return {} unless electricity_meters.count > 1
 
-        electricity_meters.each_with_object({}) do |meter, variation_by_meter|
-          variation_by_meter[meter.mpan_mprn] = calculate_seasonal_variation(meter, meter.amr_data.end_date, true)
+        @seasonal_variation_by_meter ||= begin
```
tbhi, line 138: begin now needed?

spec/system/schools/advice_pages/baseload_spec.rb
```diff
@@ -68,7 +58,7 @@
             average_baseload_kw: average_baseload_kw,
             average_baseload_kw_benchmark: average_baseload_kw_benchmark,
             annual_baseload_usage: usage,
-            baseload_usage_benchmark: usage,
+            #            baseload_usage_benchmark: usage,
```
tbhi, line 61: remove?

tbhi (approved):

PR #4418, opened 2025-04-28
run analytics spec with rest

tbhi:


ldodds:
Before we merge this and the other PR down, I'd like to get it deployed to test. I know its unlikely to perform any differently, but want to make sure there aren't any obscure issues that aren't being exercised by the tests.

PR #4427, opened 2025-04-29
Front page hero tweaks

urbanwide:
A few tweaks to the home page hero as follows:

* Remove responsive typography
* Switch image to the left
* Add more padding to hero text on xl viewports
* Add a min height for text area on xl viewports
* Remove padding to text area on sm viewports

Does look better on my screen. Image still takes up lots of vertical space on very wide screens (probably looks awful on @ldodds's :-) ). Open to other suggestions!

### DIFF
--- a/app/components/elements/image_component/image_component.scss
+++ b/app/components/elements/image_component/image_component.scss
@@ -1,15 +1,16 @@
 img.elements-image-component {
   &.stretch {
+    position: absolute;
     height: 100%;
     object-fit: cover;
   }
 
   &.right {
-    right: 0;
+    left: 0;
   }
 
   &.left {
-    left: 0;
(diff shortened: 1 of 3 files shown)


urbanwide:
Put back to draft, just tweaking something

ldodds:
The layout looks a lot better, but the image doesn't look quite right.

Here's how it looks on my screen

Screenshot from 2025-05-06 13-07-04

Basically from around 1200 pixels size the image gets slowly stretched until the top is cut off. Its caused by the fixed width I think, but without that its not the right size.

ldodds:
Can this be closed now?

urbanwide:
Yes it can, sorry!

PR #4439, opened 2025-05-07
Remove feature flags for advice index and page nav

ldodds:
The PR removes the feature flag controlling display of the new version of the advice index and nav which allows some of the templates and partials to be simplified. The `_nav` partial is then refactors the partial so that its more self-contained.

Have removed the call to cache the partial, as this will need to be done for the whole page, as the nav state is different for different pages (highlighting).


ldodds:
Realised there's a bug here: with a cached navbar, the highlighting is broken as you move between pages. So I think that means we can't cache and reuse the same content across the various pages. Key will need to include the page.

app/views/schools/advice/show.html.erb
```diff
@@ -1,19 +1,51 @@
...
-  <div class="row mt-4">
-    <div class="col">
-      <h2><%= t('advice_pages.index.show.title') %></h2>
-      <%= render PromptComponent.new(status: :none) do |c| %>
-        <% c.with_link { link_to t('advice_pages.index.show.notice_link'), priorities_school_advice_path(@school) } %>
-        <%= t('advice_pages.index.show.notice_html') %>
+<% content_for :page_title, "#{t('advice_pages.index.title')} | #{@school.name}" %>
+
+<% content_for :dashboard_header do %>
+  <%# i18n-tasks-use t('advice_pages.index.title') %>
```
tbhi, line 4: are these needed?

PR #4441, opened 2025-05-08
Calculate SECR compatible carbon emissions

tbhi:


app/models/meter_monthly_summary.rb
```diff
@@ -0,0 +1,121 @@
...
+      :incomplete
+    elsif month_readings.to_set(&:status).to_a == %w[ORIG]
+      :actual
+    else
+      :corrected
+    end
+  end
+
+  private_class_method def self.from_solar_meter(meter, meter_collection_meter, period_start, period_end)
+    return if meter_collection_meter.nil?
```
ldodds, line 82: Wonder if we need a Rollbar warning or something here? There's something wrong if we think there's a solar array but then can't find the meter.

ldodds (approved):
Looks good. I was wondering about the performance. For example, i think you could push more of the `from_main_meter` code into SQL as I think you could do all of the grouping and aggregation in the query? E.g group by the month, sum the readings and count days/status codes. That's what I was thinking originally. Obviously makes the code very different for the mains and solar. Or, as we're using the meter collection already for solar, to just read the main meter consumption from there to save the database hit? Probably worth trying on test server to see.

app/models/meter_monthly_summary.rb
```diff
@@ -0,0 +1,129 @@
...
+  validates :quality, presence: true
+  validates :total, presence: true
+  validates :year, presence: true, uniqueness: { scope: %i[meter_id type] }
+
+  def self.create_or_update_from_school(school, meter_collection)
+    today = Time.zone.today
+    meter_ids = school.meters.to_h { |meter| [meter.mpan_mprn, meter.id] }
+    Periods::FixedAcademicYear.enumerator(start_date(today, 2), today).filter_map do |period_start, period_end|
+      (meter_collection.heat_meters + meter_collection.electricity_meters).each do |meter|
+        meter = meter.sub_meters[:mains_consume] if meter.storage_heater?
```
tbhi, line 37: is this the correct way to get the original readings?

app/models/meter_monthly_summary.rb
```diff
@@ -0,0 +1,120 @@
...
+  validates :quality, presence: true
+  validates :total, presence: true
+  validates :year, presence: true, uniqueness: { scope: %i[meter_id type] }
+
+  def self.create_or_update_from_school(school, meter_collection)
+    today = Time.zone.today
+    meter_ids = school.meters.to_h { |meter| [meter.mpan_mprn, meter.id] }
+    Periods::FixedAcademicYear.enumerator(start_date(today, 2), today).filter_map do |period_start, period_end|
+      (meter_collection.heat_meters + meter_collection.electricity_meters).each do |meter|
+        meter = meter.sub_meters[:mains_consume] if meter.sub_meters.key?(:mains_consume)
```
tbhi, line 37: changed here

ldodds (approved):

PR #4447, opened 2025-05-08
Build education workshop page from new components

urbanwide:
Builds education workshops page. 

Also - adds xxl viewport, for more flexibility on wider screens

Fiddles with margins / padding a bit more elsewhere and I expect to build this more into standard component behaviour as we build more pages.

Also changes the 2 col layout to break to 1 col on md size screens and lower (tablets). I think this looks better than the squished / misaligned version with 2 columns on this size.

Keen to get this on test if ok'd with you @ldodds as Sarah is reviewing the other pages as we speak.


ldodds (approved):
Looks good. Design-wise it reinforces for me that the hero images should not be photos.

urbanwide:
I agree about the heros. Thanks for the review

PR #4454, opened 2025-05-12
Time shift alert email delivery

ldodds:
Used the Mailgun [o:deliverytime](https://www.mailgun.com/blog/email/tips-tricks-scheduling-email-delivery/) parameter to time shift the alert email deliveries.

Delivery time is set for 15 minutes after the call to send the email, to help offset the peak of Wednesday morning load caused by email generation from the additional load caused by the increase in (automated) traffic from the emails being delivered.

In theory all emails should be delivered around 15 minutes after the last `GenerateSubscriptionsJob`.

### DIFF
--- a/spec/mailers/alert_mailer_spec.rb
+++ b/spec/mailers/alert_mailer_spec.rb
@@ -20,6 +20,14 @@ RSpec.describe AlertMailer do
         expect(email.mailgun_headers['X-Mailgun-Tag']).to eql 'alerts'
       end
 
+      it 'sends an email with a mailgun deliverytime option' do
+        AlertMailer.with(email_address: email_address, school: school, events: []).alert_email.deliver_now
+        expect(ActionMailer::Base.deliveries.count).to be 1
+        deliverytime = email.mailgun_options[:deliverytime]
+        expect(deliverytime).not_to be_nil
+        expect(DateTime.rfc2822(deliverytime)).to be_between(Time.zone.now, Time.zone.now + 15.minutes)
+      end
+
       it 'specifies a subject' do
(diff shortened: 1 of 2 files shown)


spec/mailers/alert_mailer_spec.rb
```diff
@@ -20,6 +20,14 @@
         expect(email.mailgun_headers['X-Mailgun-Tag']).to eql 'alerts'
       end
 
+      it 'sends an email with a mailgun deliverytime option' do
+        AlertMailer.with(email_address: email_address, school: school, events: []).alert_email.deliver_now
+        expect(ActionMailer::Base.deliveries.count).to be 1
+        deliverytime = email.mailgun_options[:deliverytime]
+        expect(deliverytime).not_to be_nil
```
tbhi, line 27: don't need this if checking it below?

spec/mailers/alert_mailer_spec.rb
```diff
@@ -20,6 +20,14 @@
...
+        expect(DateTime.rfc2822(deliverytime)).to be_between(Time.zone.now, Time.zone.now + 15.minutes)
```
tbhi, line 28: perhaps something like `be_within(1.minute).of(15.minutes.from_now)`?

tbhi (approved):

spec/mailers/alert_mailer_spec.rb
ldodds: I tried that syntax which triggers:

```
TypeError:
  expected numeric
# ./spec/mailers/alert_mailer_spec.rb:28:in `block (4 levels) in <top (required)>'
```

PR #4456, opened 2025-05-13
minimal rubocop config

tbhi:


.rubocop.yml
```diff
@@ -34,359 +33,24 @@ plugins:
...
-  Enabled: false
-RSpec/NestedGroups:
-  Enabled: false
-RSpec/VariableName:
-  AllowedPatterns:
-    - _£_
-RSpec/ScatteredSetup:
-  Enabled: false
-RSpec/ItBehavesLike:
-  EnforcedStyle: 'it_behaves_like' # rather than 'it_should_behave_like'
```
urbanwide, line 384:
```
RSpec/ItBehavesLike:
  EnforcedStyle: 'it_behaves_like' # rather than 'it_should_behave_like'
RSpec/MultipleExpectations:
  Max: 12 # max expectations per test
RSpec/LetSetup:
  Enabled: false
```

I remember adding these.
RSpec/ItBehavesLike - to make sure we don't use it_should_behave_like, just it_behaves_like
RSpec/MultipleExpectations - good for when checking for multiple form fields on a page - can also be quicker to have multiple expectations in one test (less setup/tare down).
RSpec/LetSetup - I use this pattern, be it right or wrong, I feel like it reads / writes better on the whole. Can be tricky to debug sometimes. The alternative can lead to quite a bit of code duplication.

urbanwide (commented):
Definitely worth a try! I left a comment about a few of the cops I added a while back 👍🏻

PR #4464, opened 2025-05-19
Bootstrap mappings 5 -> 4

urbanwide:


ldodds (approved):

ldodds:
Separately we need to generate a todo list of all files that contain deprecated Bootstrap 4 classes and move to Bootstrap 5. Maybe we can tick these off in group, e.g. font classes, badge classes, etc.

PR #4487, opened 2025-05-28
Serve active storage attached images and files via CDN

ldodds:
Currently when serving images attached to, e.g. activity types, programmes, etc we use the default Active Storage option of serving the content via a redirect: the application request results in a redirect to S3.

We also directly proxy some content (e.g. attachments for case studies, jobs, downloadable resources) ourselves. This was done to provide stable links that can be used in emails, etc.

This has limitations:

- when content is served via an active storage redirect / S3 the content is cacheable in the users browser, but not publicly. We would get some minor benefit if this content could be served via the CDN instead. The ActiveStorage proxy option allows content to be publicly cached indefinitely
- the direct proxying of content via the application has the same issues in that its not cacheable. During Mailchimp sends the case study and resource content is re-downloaded multiple times.

Following guidance given in the ActiveStorage document about [serving content via a CDN](https://guides.rubyonrails.org/active_storage_overview.html#proxy-mode) this PR:

- changes how we link to attached images for activity types, programmes, etc to use a new routing helper to generate links that will proxy content via our CDN instead. This allows the images to be publicly cached
- removes the old StorageHelper code used to directly proxy attached content for case studies and other downloadable resources in favour of redirecting to a CDN based link instead, so the content is publicly cached and traffic is moved off the application

I've also tidied up some of the links to use link helpers.

I've not switched the active storage configuration to proxy mode by default as not sure that's needed. E.g. Audit attachments are not public so no benefit to caching in the CDN. For that content a redirect is still the better option.

Had to add `application/ld+json` to the `erb_lint` config to allow through the Schema.org job markup on the jobs page.


.erb_lint.yml
```diff
@@ -1,6 +1,11 @@
 ---
 EnableDefaultLinters: true
 linters:
+  AllowedScriptType:
+    allowed_types:
+      - 'application/json'
+      - 'application/ld+json'
```
tbhi, line 7: worth having a comment about the job page here?

tbhi (approved):

PR #4502, opened 2025-06-04
Tidy up group and school admin associations

ldodds:
1. Some users who were group admins have been school admins. But they are still linked to a school group.
2. Some users were school admins, linked to multiple schools, but have become group admins. But they are still linked to the schools in their group.
3. Some users are school admins but are linked to all schools in their group. They should be group admins.

2 & 3 is causing some issues with the engagement/user tracking reports as those users who should be group admins are having their logins attributed to all schools in the group.

This PR

- has an after party task that performs the updates. I've manually checked the queries are returning right users, but second pair of eyes on the logic would be helpful
- adds an active record callback to user to try and fix these issues automatically in future

### DIFF
--- a/app/controllers/users/contacts_controller.rb
+++ b/app/controllers/users/contacts_controller.rb
@@ -14,7 +14,7 @@ module Users
       # if user has been signed up for alerts from other schools, then include these
       # so they can unsubscribe
       user_contact_schools = @user.contacts.map(&:school).select(&:visible)
-      @schools = @schools + user_contact_schools if (user_contact_schools - @schools).any?
+      @schools = @schools | user_contact_schools
       @show_clusters = @schools.any? { |s| s.school_group_cluster.present? }
       render :index, layout: 'dashboards'
     end


ldodds:
I've got a log of the user ids that will be affected by this in case there's an issue, will put on Trello

lib/tasks/deployment/20250603135846_group_admins_with_cluster_schools.rake
```diff
@@ -0,0 +1,62 @@
+# rubocop:disable Metrics/BlockLength
+namespace :after_party do
+  desc 'Deployment task: group_admins_with_cluster_schools'
+  task group_admins_with_cluster_schools: :environment do
+    puts "Running deploy task 'group_admins_with_cluster_schools'"
+
+    # Find group admins who have also been manually linked to schools, e.g. were
+    # originally school admins. Remove all those associations as they're unnecessary
```
urbanwide, line 8: On my local machine I found a user (Jasmine Newhouse) with a cluster school in a different group to her school group, might not be the same on live though. Assume we are ok to remove cluster school relationships in these cases, as it shouldn't be possible for a user to be a group_admin and also have a cluster school in a different group?

urbanwide (approved):
Looks good, had a thorough check. Left one question.

lib/tasks/deployment/20250603135846_group_admins_with_cluster_schools.rake
ldodds: This is unusual but in practice I don't think it'd be doing anything. As a group user they should be getting group admin specific permissions so I don't think that relationship would do anything. I'll take another look though.

ldodds: I've confirmed that, when this user logs in they get a group admin view and don't have school admin privileges for the other school in their list. However they are linked directly to a school which is impacting what page they see when they sign on, so will update the migration to handle that too.

urbanwide (approved):
Yep looks good 👍🏻

app/controllers/users/contacts_controller.rb
```diff
@@ -11,6 +11,10 @@ def index
                  else
                    [@user.school]
                  end
+      # if user has been signed up for alerts from other schools, then include these
+      # so they can unsubscribe
+      user_contact_schools = @user.contacts.map(&:school).select(&:visible)
+      @schools = @schools + user_contact_schools if (user_contact_schools - @schools).any?
```
urbanwide, line 17: Could you do this here instead?

`@schools = @schools | user_contact_schools`

PR #4507, opened 2025-06-04
School page short-links

ldodds:
Adds a new set of short-links that attempts to redirect a user to the right location for their school. These are intended to be used from the new support pages, but might be used elsewhere (e.g. in newsletters).

The syntax is: `/s/:path`

If a user is logged in as a pupil or school user, we redirect them to the appropriate place. 

If `:path` is `dashboard` then we redirect to the school dashboard. If it is `pupils` we redirect to the pupil dashboard. Otherwise the path is assumed to be a school relative link. E.g. `/s/advice` would redirect to something like `/s/some-primary-school/advice`.

This allows all the school relative admin pages, advice, etc to be targeted.

If the user is not currently signed in, then they are prompted to login and are then redirected. Group admins or school admins linked to multiple schools will see an interstitial page where they can choose a school to visit. This seemed better than just failing, or picking an arbitrary school. 

Internal admins just get sent to a random data enabled school. Shows the links are working.

Happy to take suggestions for improving the design/text/etc. 

I've sanitized the path when it is used directly, but maybe there's other approaches? E.g. I wasn't sure if there was another way to generate the full range of routes using helpers, e.g. users, edit, advice, advice/baseload/analysis, etc.

### DIFF
--- a/app/controllers/redirects_controller.rb
+++ b/app/controllers/redirects_controller.rb
@@ -1,0 +1,38 @@
+class RedirectsController < ApplicationController
+  skip_before_action :authenticate_user!
+
+  helper_method :school_redirect_path
+
+  def school_page_redirect
+    unless current_user.present?
+      store_location_for(:user, "/s/#{params[:path]}")
+      redirect_to new_user_session_path, notice: I18n.t('users.index.redirect') and return
+    end
+    user_role = current_user.role.to_sym
+    if user_role == :group_admin || (user_role == :school_admin && current_user.has_other_schools?)
+      @path = params[:path]
+      @schools = if current_user.group_admin?
+                   current_user.school_group.schools.visible.by_name
+                 else
+                   current_user.cluster_schools.visible.by_name
+                 end
+      render :choose_school, layout: 'dashboards'
+    else
+      school = user_role == :admin ? School.data_enabled.sample : current_user.school
+      redirect_to school_redirect_path(school, params[:path])
+    end
+  end
+
+  private
+
+  def school_redirect_path(school, path)
+    case path
+    when 'dashboard'
+      school_path(school)
+    when 'pupils'
+      pupils_school_path(school)
+    else
+      "/schools/#{school.slug}/#{path}"
+    end
+  end
+end
--- a/app/controllers/redirects_controller.rb
+++ b/app/controllers/redirects_controller.rb
@@ -1,31 +1,37 @@
 class RedirectsController < ApplicationController
-  skip_before_action :authenticate_user!
+  layout 'dashboards'
 
+  skip_before_action :authenticate_user!
   helper_method :school_redirect_path
 
   def school_page_redirect
-    unless current_user.present?
-      store_location_for(:user, "/s/#{params[:path]}")
-      redirect_to new_user_session_path, notice: I18n.t('users.index.redirect') and return
-    end
-    user_role = current_user.role.to_sym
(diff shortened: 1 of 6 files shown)


urbanwide:
This is a great idea :-) Just wondering if the link has to be so short and if we could use /school instead of /s (as it's more readable to the end user?) and also more follow the current link patterns i.e.:
/school/ (for dashboard)
/pupils/school/ (for pupil dashboard)
/school/advice (and so on for other links)

I had thought it was a bit confusing that admins get sent to a random school, but listing an interim page of all schools would also be bit clunky. But maybe a flash message saying what is happening might make things less confusing?

ldodds:
@urbanwide could you finish off this PR for me? And then email the team with some examples of the new links and how they'll behave for school, group and ES admins?

re: your points:

* happy to use something longer than `/s` was just focusing on making them short. `/school` and `/schools` feels like it might be a bit easy to mix up though? We may later need some group specific urls, which I was going to set up as `/g` but maybe there's a common path, e.g. `/r/school` and `/r/group` that clarifies they are separate to existing urls.

* happy to change how the dashboard and pupil links work, but my thinking was that for most pages, you can tell someone they can take the suffix after the school slug and use that. This doesn't apply to pupils because of the arbitrary path difference (which I think we should change, to `/schools/:id/pupils` at some point, btw!) so went with a suffix. And, because all the destinations were then "named", I added "dashboard" for the adult dashboard. But I think allowing a bare `/` to hit the dashboard too ensures there's always a fallback

* for admins I wanted some behaviour to avoid reports of broken urls and to give them behaviour closer to most users. I wouldn't expect them to be using them that often so went with random. It could also be a random one of their managed schools. I agree a flash message might be helpful, so long that its clear its an admin only message, not one displayed to everyone

app/controllers/redirects_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  skip_before_action :authenticate_user!
+  helper_method :school_redirect_path
+
+  def school_page_redirect
+    new_session and return if current_user.nil?
+
+    if current_user.group_admin? || (current_user.school_admin? && current_user.has_other_schools?)
+      choose_school
+    elsif current_user.admin?
+      redirect_to school_redirect_path(School.data_enabled.sample, params[:path]),
```
tbhi, line 13: wonder if showing choose_school for all schools would be better?

config/routes.rb
```diff
@@ -61,6 +61,8 @@
 
   get 'cms/youtube_embed/:id', to: 'cms/youtube_embed#show'
 
+  get '/user/school(/*path)', to: 'redirects#school_page_redirect', as: :school_page_redirect, constraints: { path: /.*/ }
```
tbhi, line 64: wonder about
/my/path
/my/school/path
/redirect/path
/shortened/path
but as you say can change later

tbhi (approved):

app/controllers/redirects_controller.rb
urbanwide: That would be a lot of schools to show... but do see your point. We compromised on putting a notice up showing the admin user had been redirected to a random school.

urbanwide:
@ldodds This has been merged in but you might want to review what was decided as not sure of your original thinking (see comments above). I've not sent the email to the team yet (they're not waiting for it, have checked with Claudia) as we thought it best to not confuse folks if there is a chance we might change the format soon after.

PR #4536, opened 2025-06-23
Remove new_home_page feature flag

urbanwide:


app/views/home/index.html.erb
```diff
@@ -1,245 +1,201 @@
-<% cache_if current_user.nil?, I18n.locale, expires_in: 1.hour do %>
+<% cache ['home/index', I18n.locale], expires_in: 1.hour do %>
```
tbhi, line 1: don't think you need the 'home/index' part as it believe it would have a key based on the template already

are we ok with logged in users not using new blog entries, testimonials etc until the 1 hour cache expiry?

tbhi (approved):

PR #4543, opened 2025-06-25
deal with negative readings

tbhi:
- nil negative readings (currently not only gas - best for both?)
- don't overwrite the reason for nilling when substituting later so can see the original reason


app/services/aggregation/validate_amr_data.rb
```diff
@@ -0,0 +1,1288 @@
...
+      #
+      # The corrections are applied in a later step
+      process_meter_attributes
+
+      # Meter readings from the DCC can include some known values for 'bad data'?
+      # Remove these individual HH readings, or in some cases entire days of data
+      # in which these values occur
+      remove_dcc_bad_data_readings if @meter.dcc_meter
+
+      remove_negative_readings
```
tbhi, line 87: added here

ldodds (approved):

PR #4551, opened 2025-06-30
try batches for upsert to reduce db memory usage

tbhi:


ldodds (approved):
OK lets try this, test does look better. We might want to tweak the 100 row limit? I don't have a feel for our average file size I suspect it's the larger files causing the issues. Wondering whether a large size might be better initially, but happy to try this.

PR #4552, opened 2025-07-01
Zero out overnight generation data for solar meters

tbhi:


ldodds (changes requested):
`solar_pv_real_metering?` returns true if an electricity meter is linked to metered solar data. So this is zeroing out all the electricity consumption, rather than anomalies in readings for a solar generation meter. The correction should only be automatically applied to solar pv meters only. 

Can you add a basic spec to confirm this is applied only to generation meters.

ldodds (approved):

PR #4562, opened 2025-07-04
Add template caching to a few more pages

urbanwide:
Add caching to a few more pages:

* training
* newsletters
* team
* funders
* school statistics

The top two will be redesigned soon anyway. The bottom 3 hit the database, so did these at the same time.
Added some minimal spec / refactored existing spec. The whole home_spec.rb file needs a refactor, which I'll perhaps look at when doing the new training and newsletter pages.


ldodds:
One minor thing: as these pages don't have any user specific content, you could cache the contents for all users exception admins, not just non-logged in users.

ldodds (approved):

PR #4579, opened 2025-07-11
 fix handling of solis cloud data without all time periods 

tbhi:


tbhi:
replaced with https://github.com/Energy-Sparks/energy-sparks/pull/4586

PR #4588, opened 2025-07-15
Revise home page, pricing page and navbar links

urbanwide:


ldodds:
Just had a quick look over the code, seems fine. Other changes also seem sensible, copying the UTM params makes sense for now, assuming they're preserved in the routing redirects.

Case study page:

- in the hero use same buttons as the home page, rather than "Enrol"
- in the testimonial just have "Read case study"

Footer should have same links as the main nav so think thats OK.

The Enrol page is due to be replaced. Don't worry about that for now.

ldodds (approved):

PR #4592, opened 2025-07-17
Fix issues with gas substitution code producing large values during temperature compensation

ldodds:


tbhi (approved):
all seems reasonable

PR #4596, opened 2025-07-18
Watch a demo workflow

urbanwide:
https://docs.google.com/document/d/1ZyCdjj-ZtqNKSDWPv12NvYgm8O6oUNaZpg4edKA3BVg/edit?tab=t.0

It's still not quite finished but here's a list of outstanding things that I can think of:

* [x] You'll want to look at the design of these pages. I've done a quick once over of the watch-demo and more-information form pages - not sure about the image on the right, looks a bit out of place as it is. I updated the school_demo thank you page as we discussed, but the video looked too small. So I made the video wider / bigger - but not sure you can see the next steps underneath still - but I left it like this so you could see). Also updated the group_demo thank you page.
* [x] The school_demo thank you page and email are missing links to the support pages
* [x] The spec for the CampaignMailer#school_demo needs finishing (it's partially done - see campaign_mailer_spec.rb)
* [x] Have put the placeholders (school_info and group_info) in the controller etc for when the new more information flow is implemented.
* [x] Do the links on the new school_demo email need utm_params?
* [x] Also note the removal of the last find_out_more translations / erb

### DIFF
--- a/spec/mailers/campaign_mailer_spec.rb
+++ b/spec/mailers/campaign_mailer_spec.rb
@@ -97,6 +97,20 @@ RSpec.describe CampaignMailer do
     it 'sends to correct email' do
       expect(email.to).to eq(['jane@example.org'])
     end
+
+    it 'includes links to demos' do
+      expect(body).to have_link(href: short_demo_video_campaigns_url)
+      expect(body).to have_link(href: long_demo_video_campaigns_url)
+    end
+
+    it 'includes links to next steps' do
+      expect(body).to have_link(href: enrol_our_school_url)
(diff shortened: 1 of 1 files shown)


app/views/newsletters/index.html.erb
```diff
@@ -31,7 +31,7 @@
                   card.with_image(src:
               cdn_link_url(newsletter.image),
                                   rounded: :top, classes:
-              'border-light bg-light border-top border-left border-right')
+              'border-light border-top border-left border-right')
```
urbanwide, line 34: Didn't mean to leave the bg-light in here, just the border-light, so removed it.

ldodds (approved):

PR #4624, opened 2025-07-31
Calculate data for simplified school targets

tbhi:


ldodds (approved):
Calculations look good. The GenerateProgressService always uses the most recent target. If a target has expired, but data is lagging, and school has set a new target then we won't be updating the old target to backfill data. So might need to change how the service runs or perhaps better, when they can create a new target (e.g. after target data AND after we've got full data for that period), I think currently they can create a new target as soon as its expired regardless of the coverage of the usage data.

PR #4649, opened 2025-08-12
Add task for importing establishment data

ethobat:
Add a rake task `school:import_establishments` to import establishment data from the zip file obtained by `school:download_gias_data` into the table `Establishments`,
The downloaded zip file contains two files, edubasealldata[timestamp].csv and links_edubasealldata[timestamp].csv. This task only uses the first file.

### DIFF
--- a/app/models/lists/establishment.rb
+++ b/app/models/lists/establishment.rb
@@ -76,11 +76,11 @@ module Lists
     def self.import(csv_str, batch_size)
       rows = CSV.parse(csv_str, headers: true)
 
-      # Keys are headers from the CSV that match a database column when converted
+      # Array of pairs mapping headers from the CSV that match a database column when converted
       headers_to_attributes = rows.first.headers.filter_map do |h|
         [h, convert_header(h)] if Establishment.column_names.include?(convert_header(h))
-      end.to_h
-      headers_to_attributes['URN'] = 'id' # URN is the only header mapped to a column that isn't just the header in snakecase
+      end
+      headers_to_attributes.append(['URN', 'id']) # URN is the only header mapped to a column that isn't just the header in snakecase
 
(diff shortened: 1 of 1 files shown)


spec/models/lists/establishment_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'rails_helper'
+
+module Lists
+  describe Establishment do
```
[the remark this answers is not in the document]
ethobat: I didn't think it was worth writing a test for import_from_zip because it just calls two methods that have their own tests (read_data_csv_from_zip and import). Writing tests like this is new to me so I might be wrong.

what's the benefit of creating the zip dynamically instead of using a fixture?

db/migrate/20250807120146_create_establishments.rb
```diff
@@ -0,0 +1,53 @@
+class CreateEstablishments < ActiveRecord::Migration[7.2]
+  def change
+    # rubocop:disable Metrics/BlockLength
+    create_table(:lists_establishments) do |t|
```
tbhi, line 4: think you could put the rubocop:disable on this line then it would just apply to the block and the rubocop:enable wouldn't be needed

app/models/lists/establishment.rb
```diff
@@ -0,0 +1,115 @@
...
+    end
+
+    private_class_method def self.upsert_batch(batch)
+      puts "Upserting batch of #{batch.length} entries"
+      upsert_all(batch, unique_by: 'id')
+    end
+
+    private_class_method def self.create_from_row(row, headers_to_attributes)
+      ret = {}
+      headers_to_attributes.each_key do |h|
```
tbhi, line 97: wonder if similar to
```
headers_to_attributes.to_h do |header, attribute|
  [attribute, attribute_types[attribute].cast(row[header])]
end
```

tbhi (approved):

PR #4655, opened 2025-08-15
Create new Target advice pages

tbhi:


tbhi:
should there be a storage heater advice page also?

config/locales/views/advice_pages/electricity_target.yml
```diff
@@ -0,0 +1,5 @@
+---
+en:
+  advice_pages:
+    electricity_target:
+      page_title: Target for reducing your electricity use
```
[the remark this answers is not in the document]
tbhi: revised back to original and tests seem fine now

app/controllers/schools/advice/advice_base_controller.rb
```diff
@@ -71,22 +72,22 @@ def set_page_title
...
+          { name: @advice_page_title, href: advice_page_path(@school, @advice_page) }
         ]
       end
 
-      def if_exists(key)
-        full_key = "advice_pages.#{@advice_page.key}.#{key}"
-        if I18n.exists?(full_key, I18n.locale)
-          t(full_key)
-        end
+      def if_exists(key, section: nil)
```
ldodds, line 86: What I've been doing in recent code is using `default: nil`, e.g. `I18n.t(full_key, default: nil)` avoids throwing an exception if key has not been set.

app/controllers/schools/advice/advice_base_controller.rb
```diff
@@ -98,17 +99,19 @@ def set_tab_name
...
       end
 
       def check_authorisation
-        if @advice_page && @advice_page.restricted && cannot?(:read_restricted_advice, @school)
-          redirect_to school_advice_path(@school), notice: 'Only an admin or staff user for this school can access this content'
-        end
+        return unless @advice_page && @advice_page.restricted && cannot?(:read_restricted_advice, @school)
+
+        redirect_to school_advice_path(@school),
+                    notice: 'Only an admin or staff user for this school can access this content'
```
ldodds, line 109: Missing translation.

app/controllers/schools/advice/base_target_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      end
+
+      def set_page_subtitle
+        super(section: 'target')
+      end
+
+      def percent_change(current_consumption, target_consumption)
+        (current_consumption - target_consumption) / target_consumption.to_f
+      end
+      # helper_method :percent_change
```
ldodds, line 49: Not used?

ldodds:
Left a few comments on the code, rest of updates look good but:

I think there's still something not right with the behaviour when school has limited data. Testing with `redscope-primary-school`. Target starts 1st July 2025. Their gas data runs 5 Jun 2025 - 11 Aug 2025. The gas target page after regeneration says "Data for your new target should be available tomorrow.".

That message is fine if the target hasn't generated yet, but in this case they don't have enough historical data. I'll add a card in for capturing monthly historical figures but there needs to be a better interim message. See earlier comment:

> When a school has limited gas/electricity data, I suggest we add a custom "Not enough data to run analysis" page. The current target report falls back to showing their progress for the last week, this isn't ideal but gives them some feedback. Can you suggest some wording / content that includes that and which confirms what target has been set, when we'll have enough data, along with links to the long term / recent progress pages so they can explore their usage that way?

Testing with `cheddar-grove-primary-school` who have a target that has expired 2025-09-02. The electricity advice page has a progress report but doesn't indicate that the target has actually expired and that we're waiting for latest data to calculate the final result. There's a recent data warning that should appear automatically on advice pages when data is >30 days old which isn't the case here yet, but I think for expired targets this window needs to be shorter and/or have a custom message, e.g. to indicate we're waiting for data updates before giving final results. Also, their gas data doesn't have enough historical data but similar to redscope is says the analysis will be available tomorrow. But in this case there won't ever be a report.

app/views/schools/advice/base_target/_monthly_table.html.erb
```diff
@@ -0,0 +1,38 @@
...
+      <th><%= t('advice_pages.target.analysis.monthly_progress.this_year') %></th>
+      <th><%= t('advice_pages.target.analysis.monthly_progress.percentage_change') %></th>
+      <th><%= t('advice_pages.target.analysis.monthly_progress.on_target') %></th>
+    </tr>
+  </thead>
+  <tbody>
+    <% consumption.each do |month| %>
+      <tr>
+        <td><%= target_strftime(Date.new(month[:year], month[:month])) %></td>
+        <td><%= format_unit(month[:previous_consumption], :kw) %></td>
```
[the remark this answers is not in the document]
tbhi: changed

tbhi:
> Have revised to show limited data page if any previous months are nil - previously would only show if all months had missing data. Perhaps it would be if any previous months have missing data but currently missing field is for current or previous month.

rechecked code and previous month is set to nil if any missing so should be ok with this

PR #4669, opened 2025-08-21
Establishment matcher

ethobat:
Add an establishment_id column to Schools and use URNs to match them to Establishments. Also import establishment link data to get the most up-to-date establishment for successfully matched schools.


app/models/concerns/csv_importable.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    def import(csv_str, batch_size)
+      rows = CSV.parse(csv_str, headers: true)
+
+      # Array of pairs mapping headers from the CSV that match a database column when converted
+      headers_to_attributes = rows.first.headers.filter_map do |h|
+        [h, convert_header(h)] if column_names.include?(convert_header(h))
+      end
+      headers_to_attributes = headers_to_attributes.nil? ? @csv_special_columns : headers_to_attributes.union(@csv_special_columns)
```
ldodds, line 24: Rather than use `@csv_special_columns` and `@csv_name_starts_with` I think it's better to define these as methods. E.g. CSV Exportable relies on `csv_attributes` which has an empty definition in the module which is then overridden in the importing classes (`def self.csv_attributes...end`)

app/models/lists/establishment_link.rb
```diff
@@ -0,0 +1,33 @@
...
+    include CsvImportable
+    self.table_name = 'lists_establishment_links'
+
+    @csv_name_starts_with = 'links_edubasealldata'
+    @csv_special_columns = [['URN', 'establishment_id'], ['LinkURN', 'linked_establishment_id']]
+
+    belongs_to :establishment, class_name: 'Lists::Establishment'
+    belongs_to :linked_establishment, class_name: 'Lists::Establishment'
+
+    def successor?
```
ldodds, line 29: I think you can define this as a scope, e.g.:

```
scope :successor, -> { where('link_type LIKE 'Successor%') }
```
This uses the SQL LIKE query to match the contents of the `link_type` column. `Successor%` means any string that starts with "Successor". Then in Establishment you can use code like: `links.successor.first`

In this case there's not a huge difference performance wise but it avoids instantiating any EstablishmentLink objects that we don't need.

lib/tasks/schools/match_establishments.rake
```diff
@@ -0,0 +1,16 @@
+namespace :school do
+  desc 'Update establishment_id column for all active Schools in England and Wales'
+  task :match_establishments, [:path] => :environment do |_t, _args|
+    stats = { perfect: 0, la_plus_en: 0, unmatched: 0 }
+    update = {}
+
+    School.active.where(country: [0, 2]).find_each do |sch| # Only do active schools in England and Wales
```
ldodds, line 7: You can use :england and :wales here rather than the numbers, which will be clearer.

lib/tasks/schools/match_establishments.rake
```diff
@@ -0,0 +1,16 @@
...
+      update[sch.id] = { establishment_id: Lists::Establishment.find_establishment_id_for_school(sch, stats) }
```
ldodds, line 8: Rather than just return the ids, you can return the Establishment. When Rails does the update it will automatically use the id of the object. So suggest changing the Establishment method and the code here.

lib/tasks/schools/match_establishments.rake
```diff
@@ -0,0 +1,16 @@
...
+      raise 'Something bad happened!' unless update.key?(sch.id)
```
ldodds, line 9: What errors does this capture? `find_establishment_id_for_school` will return nil if there's no match and `update` will have the school id added regardless?

spec/factories/lists/establishment_link.rb
```diff
@@ -0,0 +1,8 @@
+FactoryBot.define do
+  factory :establishment_link_successor, class: 'Lists::EstablishmentLink' do
```
ldodds, line 2: Maybe just `:establishment_link`?

spec/factories/lists/establishment.rb
```diff
@@ -3,6 +3,14 @@
     establishment_name { 'sample' }
     postcode { 'AB12 34C' }
     number_of_pupils { 100 }
-    last_changed_date { 1.month.ago }
+    open_date { DateTime.current }
+  end
+
+  factory :closed_establishment, class: 'Lists::Establishment' do
```
ldodds, line 9: If you nest the `:closed_establishment` factory within `:establishment` then FactoryBot uses the outer factory to define all the main fields, then you can customise to add a close date, which saves some repetition. See users_factory and look at `:user` and `:pupil`.

spec/factories/lists/establishment_link.rb
```diff
@@ -0,0 +1,8 @@
...
+    link_name { 'Sample' }
+    link_type { 'Successor' }
+    establishment_id { 100000 }
```
ldodds, line 5: You can just write `establishment` here and FactoryBot will create the relationship using the factory with the same name. It means there will then be actual relationships in the database.

For `linked_establishment_id` you can use something like:

```
association :linked_establishment, factory: :establishment
```

spec/models/lists/establishment_spec.rb
```diff
@@ -62,5 +38,37 @@ module Lists
         end
       end
     end
+
+    describe 'uses links correctly' do
```
ldodds, line 42: I think you could make this a bit easier to read by removing the hard-coded ids and creating the relationships in the assertions, e.g.

```
it 'finds successor' do
  establishment = create(:closed_establishment)
  successor = create(:establishment_link_successor, establishment: establishment)
  expect(establishment.successor).to eq(successor)
end
```
You could also add another `link_type` and confirm successor find the right relationships.

ldodds (changes requested):

ldodds (approved):

PR #4677, opened 2025-08-28
Autofill from onboarding URN + tests

ethobat:
Add a field for the school's URN when creating onboarding data. When the user is prompted to fill in the school's details, the URN is used to find an establishment and automatically fill out as much of the form as possible.

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -128,6 +128,18 @@ class School < ApplicationRecord
   HEATING_TYPES = %i[gas electric oil lpg biomass underfloor district_heating ground_source_heat_pump
                      air_source_heat_pump water_source_heat_pump chp].freeze
 
+  GOR_CODE_TO_REGION = {
+    'A' => :north_east,
+    'B' => :north_west,
+    'D' => :yorkshire_and_the_humber,
+    'E' => :east_midlands,
+    'F' => :west_midlands,
+    'G' => :east_of_england,
+    'H' => :london,
+    'J' => :south_east,
+    'K' => :south_west
+  }.freeze
+
   friendly_id :slug_candidates, use: %i[finders slugged history]
 
   delegate :holiday_approaching?, :next_holiday, to: :calendar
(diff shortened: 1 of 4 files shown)


app/models/school.rb
```diff
@@ -878,6 +878,89 @@ def pupil_numbers_ok?
     end
   end
 
+  def self.from_onboarding(onboarding)
```
ldodds, line 881: You can initialize an object from a hash, and latter add additional attributes using `assign_attributes`, so would be clearer to write something like:

```
school = School.new({
 data_enabled: false,
 ...
})

return school if school.establishment.nil?

school.assign_attributes({
 establishment: ...,
 address: address_from_establishment(school.establishment)
 ...
})

```

app/models/school.rb
```diff
@@ -878,6 +878,89 @@ def pupil_numbers_ok?
...
+
+    est = sch.establishment
+    sch.name =                            est.establishment_name
+    sch.address =                         address_from_establishment(est)
+    sch.postcode =                        est.postcode
+    sch.website =                         est.school_website
+    sch.school_type =                     school_type_from_phase_of_education_code(est.phase_of_education_code)
+    sch.number_of_pupils =                est.number_of_pupils
+    sch.percentage_free_school_meals =    est.percentage_fsm
+    sch.region =                          region_from_gor_code(est.gor_code)
```
ldodds, line 901: looking at the data I see they're using a gor_code of "W" to indicate Wales. Could you also update the `country` field to be Wales in this case, otherwise England?

app/models/school.rb
```diff
@@ -878,6 +878,89 @@ def pupil_numbers_ok?
...
+  end
+
+  def self.concatenate_address(elements)
+    return elements.filter(&:present?).join(', ')
+  end
+
+  def self.region_from_gor_code(gor_code)
+    case gor_code
+    when 'A'
+      return regions[:north_east]
```
ldodds, line 920: Can just use `:north_east` ActiveRecord will automatically find the right number for the enum.

Could also have done this by specifying the mapping as a hash:

```
# A constant in the class
REGIONS = {
 'A' => :north_east
}.freeze
REGIONS[gor_code]
```

ldodds (changes requested):
Looks good. Couple of code styling tweaks plus additional of setting `country` field.

ldodds:
One more suggestion: SchoolOnboarding has a country field too. Can you default that based on the Establishment if there is one, e.g. if gor_code is W then Wales, otherwise England?

You can do this in the model by adding a `before_save` callback. See `before_save :update_local_distribution_zone` in School for a similar example.

ldodds (approved):

PR #4681, opened 2025-09-01
Revise analytics benchmarking values with new annual values

tbhi:

PR #4682, opened 2025-09-01
Sort out the priority numbering for case studies

urbanwide:
* Adds missing position field to case studies admin interface
* Also sorts out the rounding used for the text portion of a case study on the case studies index. Not sure what happened there as I'm sure it was right. It is currently like this on production:
image


urbanwide:
hmm appears to be broken in master too. I'll have a look (but unlikely to be done today unfortunately).

tbhi:
fix is https://github.com/Energy-Sparks/energy-sparks/pull/4681/files#diff-492dc19656a2b0fd2e90a522dc2cad4fe9eaa5ace5cccb8853192644f3346034

urbanwide:
Thank you @tbhi

PR #4683, opened 2025-09-01
Dont check meters for archived and deleted schools

ldodds:
Admins are getting notified about meters for archived schools, but we shouldn't be checking these.

Revise the scope used to check meters against the DCC to exclude inactive schools. This will ignore both archived and deleted schools.

### DIFF
--- a/app/models/meter.rb
+++ b/app/models/meter.rb
@@ -91,13 +91,7 @@ class Meter < ApplicationRecord
   scope :dcc, -> { where(dcc_meter: %i[smets2 other]) }
   scope :consented, -> { dcc.where(consent_granted: true) }
   scope :not_recently_checked, -> { where('dcc_checked_at is NULL OR dcc_checked_at < ?', 7.days.ago) }
-  scope :meters_to_check_against_dcc, -> do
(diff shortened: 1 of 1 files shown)


app/models/meter.rb
```diff
@@ -91,7 +91,14 @@ class Meter < ApplicationRecord
   scope :dcc, -> { where(dcc_meter: %i[smets2 other]) }
   scope :consented, -> { dcc.where(consent_granted: true) }
   scope :not_recently_checked, -> { where('dcc_checked_at is NULL OR dcc_checked_at < ?', 7.days.ago) }
-  scope :meters_to_check_against_dcc, -> { main_meter.not_dcc.not_recently_checked }
+  scope :meters_to_check_against_dcc, -> do
+    main_meter
+      .not_dcc
+      .not_recently_checked
+      .joins(:school)
+      .where.not(schools: { active: false })
```
tbhi, line 99: could we reuse the school scope like `Meter.main_meter.not_dcc.not_recently_checked.joins(:school).merge(School.active)`?

tbhi (approved):

PR #4693, opened 2025-09-04
Revise dashboard messages and alert email notices for new target advice page

tbhi:


spec/services/alerts/generate_email_notifications_spec.rb
```diff
@@ -232,4 +232,33 @@ def create_email_alert(fuel_type)
...
+
+    it 'shows achieving targets' do
+      create(:school_target, :with_monthly_consumption, school:, current_consumption: 1000)
+      shared_before
+      expect(email_text).to \
+        include("Priority alerts\n" \
+                'Well done, you are making progress towards achieving your target to reduce your electricity usage!')
+    end
+
+    it 'does not show an expired target' do
```
tbhi, line 258: maybe should say when a new target needs setting?

ldodds:
* `redscope-primary-school` doesn't have enough data to calculate their gas target but the prompt says its being achieved.
* The older version of the alerts didn't show the dashboard messages if a school's data is not recent (> 30 days old) but the new version still shows them. We can't make an accurate assessment of progress if the data is out of date. This is what the `check_recent` flag in ProgressSummary was doing so we could trigger determination of pass/fail in different contexts. This should be added to the dashboard and alert prompts.

PR #4696, opened 2025-09-05
Selectable Charts

ldodds:
New component that can be configured to allow a user to choose between different charts and schools then see the results on the same page.

Intended to be the basis for the new group charts page to allow group admins to quickly check charts for individual schools

- [ ] Better preview
- [ ] Chart titles and subtitles
- [ ] Dynamic links to relevant school analysis page
- [ ] Defaults from page parameters
- [ ] Disabling schools based on fuel type configuration
- [ ] Specs for component
- [ ] Page that includes component
- [ ] Tidy JS to use less JQuery
- [ ] Rename 'usage-charts` to something else?
- [ ] Chart footnotes not working
- [ ] configuration field has hard-coded chart name, specify default elsewhere?
- [ ] setSelectorState has hard-coded electricity param
- [ ] Put safe set of charts somewhere.

### DIFF
--- a/app/assets/javascripts/select.js
+++ b/app/assets/javascripts/select.js
@@ -2,6 +2,8 @@
 
 $(document).ready(function() {
 
+  $.fn.select2.defaults.set("sorter", function(datas) { return datas.filter((data) => !data.element.hidden); });
+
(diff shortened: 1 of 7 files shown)


tbhi (commented):
are there other options than select2?

PR #4710, opened 2025-09-10
Rework target comparison reports

tbhi:


db/views/comparison_electricity_targets_v03.sql
```diff
@@ -0,0 +1,22 @@
...
+         SUM((consumption ->> 4)::float) AS current_year_target_kwh
+  FROM school_targets, jsonb_array_elements(electricity_monthly_consumption) consumption
+  WHERE (consumption ->> 5)::boolean = false
+  GROUP BY school_targets.id
+)
+SELECT school_targets.school_id,
+       -school_targets.electricity AS current_target,
+       school_targets.start_date AS tracking_start_date,
+       totals.*,
+       ((totals.current_year_kwh - totals.current_year_target_kwh) / totals.current_year_target_kwh) AS current_year_percent_of_target_relative
```
tbhi, line 19: think this matches the previous calculation but not the percentage calculation on the advice pages

ldodds: If there's a discrepancy then we should be showing the same figures as now used on the new advice pages.

PR #4719, opened 2025-09-17
try query without sorting

tbhi:


ldodds:
The sorting is there to ensure that more recent data, which may be loaded via a different config (e.g. manual vs automatically provided), is always used in preference to older data.

PR #4720, opened 2025-09-17
Error generating meter monthly summaries

tbhi:
add for `'E0H1' => { name: 'Missing electricity - holiday set to zero' }`

Wonder if would be best to default to :incomplete


tbhi:
Seems like going through the other codes as would be best as separate task?

Made further modifications to move SOL0 to actual based on description "Zero pv generation data prior/post to pv install".

Also moved incomplete to before estimated as seems like incomplete should override estimated.

PR #4722, opened 2025-09-18
Fix timeline counts

urbanwide:
Noticed the counts in the left nav of the timeline was one out for a group. Realised the counting for the current year was up to the end of the year and not just today (as we are actually displaying). Also make sure we are counting only visible observations.

### DIFF
--- a/app/controllers/timeline_controller.rb
+++ b/app/controllers/timeline_controller.rb
@@ -62,6 +62,6 @@ class TimelineController < ApplicationController
   end
 
   def observation_counts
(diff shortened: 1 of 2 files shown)


urbanwide:
NB I also noticed that when comparing DATE columns to timestamps in PostgreSQL, the DATE is treated as midnight at the start of that day. This means that any times later on that day can be excluded unless the end date is adjusted, for example by adding an interval, to ensure the full day is included.

This doesn't matter so much with observations, as the time recorded is typically at 00:00:00 anyway. However there are occasional temperature observations that are throughout the day, so worth correcting.

PR #4723, opened 2025-09-19
New school group comparison reports

ldodds:
Screenshot from 2025-09-19 17-29-53

Adds a new comparison report index and alternate layout for school group reports.

Uses a different, simpler layout than the main comparison report index. Its organised based on theme and the links will appear depending on what fuel types we have data for. E.g. heating related reports won't display if none of the schools have gas data.

Most of the work is in the component which tries to handle all of the logic around:

- only showing visible reports, so we can still hide them if necessary
- only showing reports that match the fuel types for the group
- giving some flexibility around specifying link text

The new component is also used to add some additional links to related reports from the group advice pages.


ldodds:
Thinking about this last night, would a small improvement be to remove list styling for the main list, bold the list items (and single links) and add some padding? E.g.

Screenshot from 2025-09-25 08-44-33

app/views/school_groups/advice/comparison_reports.html.erb
```diff
@@ -0,0 +1,152 @@
...
+            <% list.with_named t('school_groups.advice.comparison_reports.links.baseload_variation'),
+                               fuel_type: :electricity,
+                               reports: {
+                                 seasonal_baseload_variation:
+                                  t('school_groups.advice.comparison_reports.links.seasonal'),
+                                 weekday_baseload_variation:
+                                  t('school_groups.advice.comparison_reports.links.weekday')
+                               } %>
+          <% end %>
+        <% end if @fuel_types.include?(:electricity) %>
```
urbanwide, line 100: Not seen this syntax before, had to look it up :-) Not sure if I like it or not, but better now I know what it does :-)

app/controllers/comparisons/base_controller.rb
```diff
@@ -156,5 +163,22 @@ def included_schools
       filter = filter.accessible_by(current_ability, :show) unless include_invisible
       filter.pluck(:id)
     end
+
+    def switch_layout
+      params[:group] == 'true' ? 'dashboards' : 'application'
+    end
+
+    def set_school_group_vars
+      @school_group_layout = params[:group] == 'true'
```
urbanwide, line 172: This looks like a mixed bag of things to do all in one method. I'd be thinking about splitting it out into functional groups

urbanwide (commented):
Looks good. A few things that you may or may not wish to change :-)

app/views/school_groups/advice/comparison_reports.html.erb
ldodds: It's just `x if y`. I've not used it that much in templates but Tristan suggested in a few places to reduce wrapping in `<% if... %><% end %>`. Maybe something to discuss around our preferences.

app/views/school_groups/advice/comparison_reports.html.erb
```diff
@@ -0,0 +1,152 @@
...
+        </div>
+      </div>
+
+      <%= render Layout::GridComponent.new(id: 'report-grid',
+                                           cols: 2,
+                                           cell_classes: 'mb-4',
+                                           component_classes: %w[h-100]) do |grid| %>
+        <% grid.with_block cell_classes: 'pb-2', classes: %w[p-4 bg-light rounded] do %>
+          <h4><%= t('school_groups.advice.comparison_reports.lists.long_term') %></h4>
+          <%= render SchoolGroups::ComparisonReportListComponent.new(school_group: @school_group,
```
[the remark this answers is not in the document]
ldodds: Have changed the translations.

The reports doesn't change that often, I think it'll be easier now to extract some of this out into a more manageable configuration now the UX is closer to what we want. I also want to move the main set of reports towards this style so we can maybe push the configuration into the database by improving the existing models.

urbanwide (approved):
See last comment about top margin on report page (sorry didn't spot this earlier).

PR #4726, opened 2025-09-21
Create group dashboard learn more component

urbanwide:
Currently looks like this:

image

Has no spec as yet. Put it here for discussion @ldodds. Wasn't sure about the list of schools... Currently uses select2, which I guess we are phasing out at some point as it's jquery.


urbanwide:
Hi @ldodds, here's a restyled version. Not sure about the colours and I appreciate it doesn't line up with the block below very well right now. Both new blocks have rounded-xl corners (like the marketing pages), which aren't typically found on the dashboards, so wondering if they should just be lightly rounded instead (though I tried it and it looks a bit odd). 

The groups savings component below is wider than the alerts too, so one option would be to make the groups savings component less wide to line up with reminders and alerts and now the new learn more component.

image

urbanwide:
Latest. Am using the same rounding as reminders and alerts for dashboard consistency. I wonder if the school dropdown box should go on the left and the advice page link on the right? Feels more similar to the school dashboard with the advice page links on the right:

image

Also updated the school dashboard:

image

ldodds (approved):
Looks good, think we can leave wording as is pending any internal feedback.

Minor comment: maybe change the preview to allow selecting of groups without active schools to preview the not data enabled version and/or add an option for that.

PR #4735, opened 2025-09-23
Add ability for schools to add historical monthly usage figures

tbhi:
- just model and form

### DIFF
--- a/spec/system/school_groups/school_engagement_spec.rb
+++ b/spec/system/school_groups/school_engagement_spec.rb
@@ -14,6 +14,7 @@ describe 'School group engagement' do
 
   context 'when signed in' do
     before do
+      travel_to(Date.new(2025, 11))
       sign_in(create(:admin))
       create(:school, data_enabled: false, school_group: school.school_group)
       create(:user, school:, last_sign_in_at: Date.new(2025, 5, 1))


app/controllers/schools/manual_readings_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+      if @school.save
+        redirect_to school_manual_readings_path(@school), notice: t('common.saved')
+      else
+        render :show
+      end
+    end
+
+    private
+
+    def manual_months
```
ldodds, line 34: Could we remove the reliance on having an existing target? I want to use this manually supplied data to improve long term use pages for schools with limited data. Also, we might want to prompt schools to add this before they've set a target. Maybe just default to allowing up to a year of data?

config/locales/views/schools/schools.yml
```diff
@@ -240,6 +240,11 @@ en:
       timeout_modal:
         still_watching_your_live_data_usage: Still watching your live data usage?
         temporarily_paused_message_html: You've been viewing your live data for %{timeout_interval} minutes. We've temporarily paused updates. Click Restart to continue viewing
+    manual_readings:
+      completed: This table needs to be completed to calculate your target starting %{start_date}.
+      enough_data: We have enough data from your meters so you don't need to enter any readings manually.
+      intro: This table allows you to provide total kWh values for the whole school using data from your monthly bill or sourced from your energy supplier portal.
```
ldodds, line 246: "We have limited energy data for your school. To allow you to set targets to reduce your energy use you can use this form to provide historical usage data. You will need to check either your monthly bill or your energy supplier portal and then provide the total kWh used for each month.

If we are later able to get access to historical half-hourly readings from your energy supplier then we will automatically use these instead."

app/views/schools/manual_readings/show.html.erb
```diff
@@ -0,0 +1,42 @@
...
+          <tr>
+            <td>
+              <% if reading_form.object.month %>
+                <%= reading_form.hidden_field :month %>
+                <%= I18n.l(reading_form.object.month, format: '%B %Y') %>
+              <% else %>
+                <%= reading_form.input :month, label: false, discard_day: true %>
+              <% end %>
+            </td>
+            <td><%= if @months[:electricity]&.include?(reading_form.object.month)
```
ldodds, line 32: Maybe need a note or something here rather than a gap, e.g. "Not required"

app/views/schools/manual_readings/show.html.erb
```diff
@@ -0,0 +1,42 @@
+<h1><%= t('manage_school_menu.manage_manual_readings') %></h1>
+<% if @current_target.nil? %>
+  <p><%= t('schools.manual_readings.set_target_html', path: new_school_school_target_path(@school)) %></p>
+<% elsif @months.values.flatten.empty? %>
```
ldodds, line 4: Could we not link to the form if there's no need to use it? Saves adding menu items, etc that may not be needed?

Rather than checking what months are needed every time, then we could add a flag to the FuelConfiguration during regeneration to decide whether adding manual readings is worthwhile, then show the links?

It might be useful for internal admins to see previously supplied data though.

db/schema.rb
```diff
@@ -1902,6 +1902,17 @@
     t.index ["urn"], name: "index_schools_on_urn", unique: true
   end
 
+  create_table "schools_manual_readings", force: :cascade do |t|
+    t.bigint "school_id", null: false
+    t.date "month", null: false
+    t.integer "electricity"
```
ldodds, line 1908: Need to be floats

app/views/schools/manual_readings/show.html.erb
```diff
@@ -0,0 +1,42 @@
...
+      <thead>
+        <tr>
+          <th><%= t('common.labels.date') %></th>
+          <% @months.each do |fuel_type, months| %>
+            <th><%= t("common.#{fuel_type}") unless months.empty? %></th>
+          <% end %>
+        </tr>
+      </thead>
+      <tbody>
+        <%= f.simple_fields_for :manual_readings, @readings do |reading_form| %>
```
ldodds, line 22: Fields need to allow for float inputs. Usage might be 17.5 kWh for example. 

Also don't use numeric fields, as the UX adds the controls for increasing and decreasing by one, which isn't great for data entry. Users are rarely going to use those.

app/controllers/schools/manual_readings_controller.rb
ldodds: Ideally we want at least a year (13 months to allow for comparisons) for each fuel type. This could be used for targets or elsewhere.

For targets, while we have schools that currently have limited data and data, when setting new targets you shouldn't be able to set one in the past, so the start date for those shouldn't be earlier than the current month.

app/controllers/schools/manual_readings_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+    load_and_authorize_resource through: :school
+
+    before_action :set_breadcrumbs
+
+    def show
+      @meter_start_dates = %i[electricity gas].select { |fuel_type| @school.configuration.fuel_type?(fuel_type) }
+                                              .index_with do |fuel_type|
+        @school.configuration.meter_start_date(fuel_type)
+      end
+      return if @meter_start_dates.empty?
```
urbanwide: Presume this will only return here if there are no fuel types configured for the school. Does this behave when rendering the page? Assume it give the "your school has enough data message" which wouldn't quite be right. Depends if it is possible for this to happen I guess.

app/controllers/schools/manual_readings_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+        show
+        render :show
+      end
+    end
+
+    private
+
+    def build_manual_readings
+      end_date = @meter_start_dates.values.compact.max || Date.current
+      return if end_date < 1.year.ago
```
urbanwide: Please could we have some documentation of the logic here and what the code is trying to achieve?
i.e. Are we trying to ensure we have 13 months of readings, if we don't have the actual readings, fill in with manual readings. 
Looks like we are finding the latest start time for existing meter data across fuel types. If this is over a year ago, then return as presume this means we have enough data? Is this true though if we're missing data at the end? Should it be 13 months?

spec/system/schools/manual_readings_spec.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
```
urbanwide: Could test setup be done in before blocks and expect statements go in 'it' block where possible please?

urbanwide:
I was thinking that we might want to change the interface behaviour when deleting the contents from readings fields. You'd expect when deleting a value from a field and then saving, that it would remain empty, but the old value then re-appears. If this worked, then we wouldn't need the "Remove" checkbox from the side.

urbanwide:
I was wondering if we could add a few things to the interface. I found myself checking the dashboard (once I'd worked out where it was) to see what data already exists. It would be good if we could in some way be more explicit about the data we have for a school already and the data we are asking for as it's not clear why here.

Whether this be a reminder that "We already have Electricity data from Oct 2020 to Aug 2025 and Gas data from June 2025 to Sept 2025, please fill in the missing or partial months below where possible" and/or a more visual reminder as below (have no idea how practical it would be to add the figures in. The idea of the grey boxes is that they are the non-editable / disabled. They don't have to be grey either, just an idea.
Screenshot 2025-10-15 at 23 25 30

app/controllers/schools/manual_readings_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+        @school.configuration.meter_start_date(fuel_type)
+      end
+      return if @meter_start_dates.empty?
+
+      build_manual_readings
+      @readings = @school.manual_readings.sort_by(&:month)
+    end
+
+    def update
+      @school.update(resource_params)
```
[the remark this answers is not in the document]
tbhi: :+1:

app/services/date_service.rb
```diff
@@ -12,4 +12,14 @@ def self.subtitle_end_date(chart_config, date)
       date
     end
   end
+
```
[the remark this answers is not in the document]
tbhi: :+1:

spec/system/schools/manual_readings_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+RSpec.describe 'manual readings' do
+  let(:school) { create(:school, :with_fuel_configuration, :with_meter_dates) }
+
+  before do
+    travel_to(Date.new(2025, 8, 15))
+    Flipper.enable(:manual_readings)
+    sign_in(create(:school_admin, school:))
+  end
+
+  it 'is on the manage school menu' do
```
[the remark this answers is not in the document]
tbhi: moved

app/controllers/schools/manual_readings_controller.rb
tbhi: > Ideally we want at least a year (13 months to allow for comparisons) for each fuel type. This could be used for targets or elsewhere.
> 

wonder if allowing input for 2 years of data would be better?

tbhi:
added to the PR to split missing field in the calculated target consumption field so it is clearer whether current or previous data is missing which meant also updating other users of the field

spec/system/schools/manual_readings_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+  context 'without a target' do
+    it 'has electricity only' do
+      visit school_manual_readings_path(school)
+      expect(form_input_values).to eq([[], *month_input_values(2024, 7, 12, [nil])])
+      complete_form
+      expect(school.manual_readings.order(:month).pluck(:month, :electricity, :gas)).to \
+        eq((0..12).map { |i| [Date.new(2024, 7) + i.months, 5, nil] })
+    end
+
+    it 'with gas' do
```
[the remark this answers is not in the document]
tbhi: thanks, reworked, think looks clearer

config/locales/views/schools/schools.yml
```diff
@@ -638,6 +644,7 @@ en:
       metering:
         manage_meters: View and manage your school electricity and gas meters
         manage_tariffs: Manage current and historical energy tariffs for your school
+        manual_readings: Enter readings manually where we have missing data
```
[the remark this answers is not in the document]
tbhi: added with missing word track?

PR #4739, opened 2025-09-25
Rework recent use table for school groups

ldodds:
Screenshot from 2025-09-27 12-31-08

Reworks the recent usage table show on the school group advice page. Moves from having a single table with all fuel types to separate tables in a tabbed view

* Creates a new component for displaying energy use for a specific fuel type, across a range of periods
* Creates a new component for displaying tabbed set of tables for the available fuel types for a school group
* Tables only show data enabled schools with relevant fuel type, not all schools. This reduces number of empty rows
* Adds footnote to the tables, to allow user to see summary of analysis. Currently this consists of some table specific notes, plus some general text that will be added to other group advice pages
* Adds a new monthly reporting period, which covers last full calendar month with usage compared to same month last year
* Removes some old redundant code and calculations from the analytics alert that produces the data

There are some further things to do which will come later:

* Table columns may need to change, e.g. "Last week" -> "Latest week". Need feedback
* The existing code to read and present the summary data for the alert could be simplified. I've not touched this yet other than to make the new monthly data available
* There are some helpers used in partials that are shared with the old code, these can be removed later and moved into the components


app/components/dashboards/group_energy_summary_table_component.rb
```diff
@@ -0,0 +1,51 @@
+module Dashboards
+  class GroupEnergySummaryTableComponent < ApplicationComponent
+    PERIODS = %w[week month year].freeze
```
tbhi, line 3: %i for symbols?

app/components/dashboards/group_energy_summary_table_component.rb
```diff
@@ -0,0 +1,51 @@
...
+    def show_clusters?
+      @show_clusters
+    end
+
+    def render?
+      @schools.any? && @periods.any?
+    end
+
+    def period_label(period)
+      case period
```
tbhi, line 27: maybe I18n.t("common.labels.last_#{period}")

app/components/dashboards/group_energy_summary_table_component.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    private
+
+    def value_for(recent_usage, formatted: true)
+      return nil unless recent_usage
+      case @metric
+      when :usage then formatted ? recent_usage.usage : recent_usage.usage_text
+      when :co2 then formatted ? recent_usage.co2 : recent_usage.co2_text
+      when :cost then formatted ? recent_usage.cost : recent_usage.cost_text
+      when :change then formatted ? recent_usage.change : recent_usage.change_text.gsub(/[^-.0-9]/, '')
```
tbhi, line 45: could drop this one as same as else?

app/components/dashboards/group_energy_summary_table_component/group_energy_summary_table_component.html.erb
```diff
@@ -0,0 +1,58 @@
...
+    <tbody>
+      <% schools.data_enabled.order(:name).each do |school| %>
+        <% if school.configuration.fuel_configuration.fuel_types.include?(fuel_type) %>
+          <tr>
+            <td>
+              <%= link_to school.name, school_path(school) %>
+            </td>
+            <% if show_clusters? %>
+              <td>
+                <%= school.school_group_cluster_name %>
```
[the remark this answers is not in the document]
ldodds: Going to leave for now, as will be easier to turn this into a hyperlink to prompt them to add clusters.

PR #4741, opened 2025-09-25
Add secr to advice page nav plus a few other fixes

urbanwide:
Adds SECR report link to group advice nav.

I am a bit confused about permissions. Should a group admin from a group be able to see another group's secr link and report? I think all school group admins can see any secr report at the moment for any school group. Currently am using can? (:show_management_dash) for the link which only allows group admins to see their own but not sure this is right @ldodds ?

Made some other nav changes while I was there - the nav was highlighting Energy use charts when on any of the top items. i.e.
Screenshot 2025-09-25 at 18 11 38
 For the report section (i.e. a non fuel type section) the current and hover was not rounded (like the fuel type ones were). The section header had a grey hover when the fuel sections didn't. Just made them consistent, hope that's ok.
Also added the ability to show/hide an item based on a boolean (i.e. if visible variable is positive, it will show)

### DIFF
--- a/app/components/page_nav_component.rb
+++ b/app/components/page_nav_component.rb
@@ -75,14 +75,14 @@ class PageNavComponent < ApplicationComponent
   class ItemComponent < ViewComponent::Base
     attr_reader :name, :href, :match_controller, :classes
 
-    def initialize(name:, href:, note: nil, match_controller: false, selected: false, classes: nil, **kwargs)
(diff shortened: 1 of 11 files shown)


ldodds:
re: permissions. Group admins should only be able to see their own SECR reports.

ldodds (commented):
See comment re: SECR reports, other changes look good. I think I fixed the nav issue in another PR but thanks for catching it and fixing the styling problems. Same was probably happening on the school settings and support pages.

PR #4755, opened 2025-10-01
Chart and table footers for school group analysis

ldodds:
Add in the "How did we calculate this data" links for the school group charts page and the school group advice analysis pages so they're displayed under the charts and alongside the savings table.

As the model contains similar content, I've tweaked the introductory text for the savings table to include a link that will trigger the modal instead. Saves repeating the same information

Couple of other tweaks:

- disable modals on the group comparison components included on the insights page. There's a link to a table with all of the information
- add `table-sm` to the group comparison component modal, where it is shown
- fix styling of the additional report links on the advice pages


ldodds:
> wonder about a shared partial for the analysis comparison sections as seems quite similar there

Yes, can do some post release tidying I think.

PR #4763, opened 2025-10-02
Change navbar for groups / group admin

urbanwide:
UPDATED LOGIC for second nav link "left link":

If we are a page in a school context, display the school with link to school dashboard.
Otherwise, if the current user has a school, display “< Back to dashboard” link
Otherwise, if we are on a page in a group context (and we have permission to :compare the group), display the group with link to school group dashboard.

This also adds a "School group dashboard" link to the second nav "dashboards" links, to left of the bar on right hand side. This is shown if we are on a page in a group context (and we have permission to :compare the group)

### DIFF
--- a/spec/system/navigation/second_nav_spec.rb
+++ b/spec/system/navigation/second_nav_spec.rb
@@ -82,7 +82,7 @@ RSpec.describe 'Navigation -> second nav' do
       end
     end
 
-    context 'when current user is a school admin' do
+    context 'when current user is a school admin (i.e. has a school assigned to their user)' do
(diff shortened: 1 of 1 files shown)


ldodds (approved):
I think looks OK. I'm not sure about whether there's a need to do the `:compare` check in both places. 

All of the school group/advice controllers should redirect a user to the group map page if they cannot `:compare` a group. 

So I think this check really just means the group name won't appear on the left on the Map page? In this instance you could still include the group name, but not have a link?

It probably makes sense to not show the "Group dashboard" link on the right if you can't `:compare` the group though.

PR #4766, opened 2025-10-03
correct target calculation with no complete current consumption

tbhi:


ldodds:
Testing with `oasis-academy-aspinal` locally. I have data up to 11th August 2025. Set a target for 1st August and regenerated. Insights table, analysis tables look correct.

I then updated the target to 1st July and regenerated. The analysis table looks correct but the insights table is reporting a 20% reduction. 

We can't tell if they're meeting their target as we don't have a full month of recent data. And in my dev database the data is now >30 days out of date. Either scenario should be enough to not produce a % change (and by extension a dashboard message, but those don't seem to be appearing).

tbhi:
> I then updated the target to 1st July and regenerated. The analysis table looks correct but the insights table is reporting a 20% reduction.
> 
> We can't tell if they're meeting their target as we don't have a full month of recent data. And in my dev database the data is now >30 days out of date. Either scenario should be enough to not produce a % change (and by extension a dashboard message, but those don't seem to be appearing).

would there not be July's data?

will look into hiding % if >30 days out of date

ldodds:
> would there not be July's data?

Yes, you're right, we should use July here if there was recent data.

PR #4917, opened 2025-12-04
pagy upgrade

tbhi:
https://ddnexus.github.io/pagy/guides/upgrade-guide/
-  Pagy::Backend -> Pagy::Method
- replace custom nav template with pagy.series_nav(:bootstrap) which looks to generate the same
- update intializer


config/locales/pagy.yml
ldodds, line 1: As you've renamed the file, you also need to update `.tx/confg` or re-run the command to generate it.

config/locales/cy/pagy.yml
ldodds, line 1: Moving it here breaks the convention that all the Welsh files are in the cy directory. The Transifex config assumes the file will be written to that directory.

We might just rely on the Pagy defaults, but think we want the option to tweak those ourselves so need to make sure we can round-trip to Transifex like the other files, so maybe put into standard directories?

ldodds: See `rake i18n:generate_tx_config`

tbhi:
had issues with pluralisation for welsh (needs key pagy.p11n for pagy i18n - set to OneOther for English but don't think has the correct one for Welsh) so switched to standard i18n which I believe has support for this already - this also restores the original file locations

tbhi:
not sure Welsh pluralisation actually works in rails either - https://github.com/svenfuchs/rails-i18n/pull/1178 - looks to fall back to OneOther like english

tbhi:
seems ok to go with the standard gem for now which will continue the use of OneOther for both languages but hopefully that will get resolved with a new version of the rails-i18n gem

ldodds (approved):

PR #5142, opened 2026-03-18
Automatically update pupil numbers using DfE data

tbhi:
- [x] make school pupil number field readonly to non admins
- [x] update free school numbers
- [x] don't update if admin has set numbers
- [x] set start date to start of the previous academic year
- [x]  admin report that lists active schools that we aren’t automatically updating along with a reason; partial schools, admin attribute
- updated FloorAreaPupilNumbersBase to work with floor area and pupil number to allow only setting pupil numbers


lib/tasks/schools/update_pupil_numbers.rake
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+namespace :school do
+  desc 'Update pupil numbers based on establishment data'
+  task update_pupil_numbers: :environment do
+    Rails.logger = Logger.new($stdout) if Rails.env.development?
+    School.active.joins(:establishment).where(full_school: true).find_each do |school|
```
ldodds: I think this would be better in a scope on school.

ldodds (changes requested):
See inline comments for initial feedback

But there's another aspect of the PupilNumberUpdater which I'm not sure is quite right. 

The older version carried forward any floor area changes into the  newly created attributes, which I don't think this is now doing? E.g. `/admin/schools/woodhall-primary-school/school_attributes` the floor area would need to be carried forward to any updates to the pupil numbers. The new version would add an end date to the existing attribute which means the floor area change would be reset.

There's validation on the meter attribute requiring a user to add both a floor area and a pupil number in order to create or update an attribute, so I'd expect all(?) existing attributes to have both values currently.

It's obviously not ideal that this was originally created as a single attribute that covers two independently varying measures.

lib/tasks/schools/update_pupil_numbers.rake
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+namespace :school do
+  desc 'Update pupil numbers based on establishment data'
+  task update_pupil_numbers: :environment do
+    Rails.logger = Logger.new($stdout) if Rails.env.development?
+    School.active.joins(:establishment).full_school.find_each do |school|
+      # GIAS receives information in January from the previous September
+      start_date = school.academic_year_for(1.year.ago)&.start_date
```
tbhi, line 9: perhaps if today is before January the start date should be 2 years ago?

app/services/schools/pupil_number_updater.rb
```diff
@@ -1,71 +1,69 @@
...
-      floor_area = most_recent_expired_floor_area
-
-      input_data = {
-        'start_date' => today,
-        'end_date' => nil,
-        'floor_area' => floor_area,
-        'number_of_pupils' => pupil_count.to_s
-      }
+    def should_create_attribute?(number_of_pupils, start_date, attribute, data)
+      (attribute.created_by_id.nil? || data[:end_date] <= start_date) &&
```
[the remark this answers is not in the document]
tbhi: believe it should be doing this

app/views/admin/reports/base_meter_reports/_report_filters.html.erb
```diff
@@ -13,8 +13,8 @@
                          class: 'form-control-sm' %>
         </span>
         <span class="nowrap">
-          <%= label_tag :user, nil, class: 'small' %>
-          <%= select_tag :user,
+          <%= label_tag :admin, nil, class: 'small' %>
+          <%= select_tag :admin,
                          options_from_collection_for_select(User.admins_by_name, :id, :display_name, params[:user]),
```
ldodds: Should be `params[:admin]`

ldodds: Also the renamed parameter breaks `/admin/reports/admin_user_meter_report`

app/views/admin/reports/pupil_number_updates/index.html.erb
```diff
@@ -0,0 +1,28 @@
...
+    <tr>
+      <th>School Group</th>
+      <th>Admin</th>
+      <th>School</th>
+      <th>Reason</th>
+      <th data-orderable="false"></th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @schools.by_name.each do |school|
```
ldodds: By group then by school name would be a UX improvement

app/services/schools/pupil_number_updater.rb
```diff
@@ -1,71 +1,74 @@
...
+    DATE_FORMAT = '%d/%m/%Y'
+    AUTOMATED_REASON = 'Automated pupil number update using DfE data'
 
     def initialize(school)
       @school = school
     end
 
-    def update(pupil_count, reason = 'Automated pupil number update')
+    def update(number_of_pupils, percentage_free_school_meals, start_date,
+               reason_prefix = 'Automated pupil number update')
```
ldodds: Can use the constant as the default `reason_prefix`?

lib/tasks/schools/update_pupil_numbers.rake
ldodds: Currently they have the January 2025 census data in the database. covering the period Sept 2024-August 2025 AY. The database was will have been updated to show that data in July/August 2025, so towards the end of the academic year.

Then from around July/August 2026, we can assume that the database will be updated to use the January 2026 census (covering Sept 2025-August 2026).

So I think don't we should be looking at 2 years ago, It's more possible that they might update the data in August 2026 and we misattribute that to the Sept 2024-August 2025 academic year?

Unless I misunderstood the explanation on their site?

ldodds (changes requested):
Question about time period still to be answered, couple of minor issues with broken reports.

lib/tasks/schools/update_pupil_numbers.rake
tbhi: sounds right

app/services/schools/pupil_number_updater.rb
tbhi: removed default to keep more explicit

ldodds (approved):

PR #5167, opened 2026-03-31
Update case studies to allow a group or school to be selected per case study

urbanwide:
This does quite a bit (sorry) and will need a manual test (and maybe reviewing on test too), due to simple_form and select2 changes. See below:

* Updates case studies admin interface to allow a school or school group to be assigned to a case study
* Updates Impact Report page to include only related testimonials or case studies. It retrieves case studies related to the school group or its schools, and displays associated testimonials for each case study. If there are no associated testimonials, it displays the case study itself.
* Adds a new case study component for displaying as an alternative to a testimonial
* Updates case studies frontend and admin interface to work with Bootstrap 5, which has led on to:
  * The Simple Form configuration was generating Bootstrap 4 tags in the output. Installed the new Bootstrap 5 config (simple_form_bootstrap5.rb) and prefixed the old Bootstrap 4 wrappers with bs4_, moving them into simple_form_bootstrap4.rb. These are then included in the main Simple Form config. A BootstrapSwitcherFormBuilder (hooked in via simple_form_helper.rb) switches between the namespaced wrappers based on Current.bs5. Probably could find a better way given time but seems to work as is.
  * Updates locale_tabs for Bootstrap 5
  * Updates select2 so it uses the bootstrap 5 theme when needed. Have upgraded the javascript to the latest version (for bootstrap 5) but it still uses jquery.
  * Updates issues forms to work with Bootstrap 5, now simple form is working.


app/helpers/application_helper.rb
```diff
@@ -590,4 +590,21 @@ def bootstrap_path(path = '')
...
+    klass = entity.is_a?(Class) ? entity : entity.class
+
+    return 'users'    if klass == SchoolGroup
+    return 'school'   if klass == School
+    return 'download' if klass == DataSource
+
+    ''
+  end
+
+  def entity_icon(entity)
```
urbanwide: This replaces issueable_icon, as I wanted to use the same icons

ldodds (approved):
Looks good. What extra manual testing is needed to test the simple form changes? I tried out a select2 elsewhere on the site but haven't manually checked any other forms.

urbanwide:
> Looks good. What extra manual testing is needed to test the simple form changes? I tried out a select2 elsewhere on the site but haven't manually checked any other forms.

Thank you. It would be good if you could check a few complex forms on the site really. Just to check they display as expected, as the simpleform changes have potential to affect all form layouts. I have checked a few of them but would be good to get your eye over some of them that you might know better.

ldodds:
Noticed a problem with a dynamically inserted select box on my new bulk licence editing form:

Screenshot from 2026-04-20 11-25-30

Row is added via create_licence.js.rb which ends up calling licence_row_component.html.erb to render the new row, so something changed re: the form markup. You don't have the data to test this so I can help fix if you have any suggestions?

I've reviewed a bunch of other forms (just opening them to check formatting) across the site both user and admin facing and haven't noticed any other errors yet. The above makes me wonder if there's any other forms with dynamically added sections. I've checked the school times page and the manage audts page which both look OK. Can't think of any others atm.

PR #5170, opened 2026-04-01
Admin dashboard quick links

LukeGouldson:
Adds quick links for 'select school group', 'select school', 'find mpxn' and 'find urn' to the admin dashboard show page


ldodds (approved):
A small bug in the redirect for the school link, and some extra specs needed, but otherwise looks good.

I've included some suggestions for making the markup more explicit and improving the UX a bit.

If you can do those then OK to merge.

tbhi (approved):

PR #5174, opened 2026-04-02
Impact report basic models and configuration admin

urbanwide:
Basic models as discussed @ldodds
Please see the SchoolGroups::ImpactReport service for any calculations and placeholders (this is a temporary location). This is currently used by the frontend to retrieve metrics but code needs adding to generate a ImpactReport::Run with ImpactReport::Metrics (models for this added in the PR, but probably needs discussion with Leigh and some modification to make it fit for purpose). The frontend then needs hooking in to use the ImpactReport::Run and associated Metrics.

Very basic admin interface also added for configurations: http://localhost:3000/admin/impact/configurations


app/models/impact_report/configuration.rb
```diff
@@ -0,0 +1,28 @@
...
+# Indexes
+#
+#  index_impact_report_configurations_on_school_group_id  (school_group_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (school_group_id => school_groups.id)
+#
+
+module ImpactReport
```
ldodds, line 22: Assume this is just the minimum-viable version as there's other options to add, e.g. whether to show analysis, etc? That's fine, just checking.

app/controllers/admin/impact/configurations_controller.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Admin
+  module Impact
+    class ConfigurationsController < AdminController
+      before_action :enable_bootstrap5
+
+      load_and_authorize_resource :school_group, include: :impact_report_configuration, id_param: :id, except: [:index]
+
+      def index
```
ldodds, line 10: It makes sense to have an index so we can compare configurations, although I'd expect in most cases in order to manage the configuration for a group we'd want to do so from their settings page.

This also highlights that we need to think about generating reports for different types of groups, so far I've only thought about MATs, but perhaps we need other group types too. Maybe this should default to organisation type groups initially?

app/views/admin/impact/configurations/index.html.erb
```diff
@@ -0,0 +1,26 @@
+<% content_for :page_title, 'Impact Report Configurations' %>
+<h1>Impact Report Configurations</h1>
+<table class="table table-paged">
+  <thead>
+    <tr>
+      <th>School Group</th>
```
ldodds, line 6: Any time we're listing groups or schools we normally want to show the admin user too.

app/models/impact_report/metric.rb
```diff
@@ -0,0 +1,53 @@
...
+                          users_logged_in_recently: 3,
+                          pupils: 4,
+                          enrolled_schools: 5,
+                          enrolling_schools: 6,
+                          total_saving: 7,
+                          reduced_emissions: 8,
+                          activities: 9,
+                          actions: 10,
+                          points: 11,
+                          featured_school: 12,
```
ldodds: This should be part of the configuration, not a metric. The admin user will pick their featured school for the engagement section and (separately) the analysis section. I was expecting any numbers show in those panels to be "live", e.g. current scores, etc.

app/models/impact_report/metric.rb
```diff
@@ -0,0 +1,53 @@
...
+module ImpactReport
+  class Metric < ApplicationRecord
+    self.table_name = 'impact_report_metrics'
+
+    include Enums::FuelType
+
+    belongs_to :impact_report_run, class_name: 'ImpactReport::Run', inverse_of: :metrics
+
+    enum :metric_category, { overview: 0, energy_efficiency: 1, engagement: 2, potential_savings: 3 }, prefix: :category
+    enum :metric_type, {  visible_schools: 0, # Not sure if this should be the name of a key instead
```
ldodds: Lately I've been using enums that are backed with a Postgres enum as that means we end up storing the keys rather than numbers, which I think is clearer if we end up writing SQL.

I also think we discussed that we might have broader metric types, with detail in the value json. So maybe we just need users as a metric and the value has numbers who have logged in recently, number of pupils, etc.

Similarly a schools metric might have value with enrolled, visible, data visible, enrolling, etc.

total_saving might have a value with co2, kwh, and cost?

Perhaps to start with we just need to define metric types for for the engagement section and for the total savings then add to the list? Probably easier to extend it rather than chop things out?

ldodds (changes requested):
Have left a few comments but I think the only thing that needs tweaking is the metric type, the rest can be improved later.

app/models/impact_report/configuration.rb
urbanwide: Yes. Just the shell

app/models/impact_report/metric.rb
```diff
@@ -31,23 +31,41 @@ class Metric < ApplicationRecord
...
+      overview: 'overview',
+      energy_efficiency: 'energy_efficiency',
+      engagement: 'engagement',
+      potential_savings: 'potential_savings'
+    }.freeze
+
+    enum :metric_category, METRIC_CATEGORIES, prefix: :category
+
+    OVERVIEW_METRICS = {
+      schools: 'schools',
```
urbanwide: I know you didn't mention to add these in initially @ldodds, but perhaps they are stable enough to add now?

app/controllers/admin/impact/configurations_controller.rb
urbanwide: Would you like me to add this to the settings page at this point? Behind the feature flag?

app/models/impact_report/metric.rb
urbanwide: Thanks, have updated the enums to be backed by Postgres and made the labelling broader (was rushing a bit sorry).

In the update, as well as the metrics types for the engagement section and total savings, I've also defined 'overview' ones, as these are reasonably well defined too. Hope this is ok @ldodds ?

app/views/admin/impact/configurations/index.html.erb
urbanwide: I'll put this in my next PR (Further refinements to the admin interface)

app/controllers/admin/impact/configurations_controller.rb
ldodds: You could do, but maybe in follow up PR

ldodds (approved):

PR #5180, opened 2026-04-13
Admin dashboard initial prompts

LukeGouldson:
Adds prompts to dashboard home page for overdue issues, this week's issues, school activations, lagging data sources and data feeds not receiving data. Done as a view component. Links do not use the dashboard controller yet, instead navigating to corresponding existing pages


LukeGouldson:
The AMR prompt seems to take a long time to load. I wondered whether there was a faster way of implementing it

app/components/admin/dashboard_prompt_component.rb
```diff
@@ -0,0 +1,105 @@
...
+        { id: 'weekly-issues', check: prompt_for_weekly_issues?, status: :neutral, icon: 'magnifying-glass',
+          link: 'View Issues', path: admin_issues_path(user: @user),
+          content: "You have #{@weekly_issues} issues due for review in the next week" },
+        { id: 'school-activation', check: prompt_for_school_activation?, status: :neutral, icon: 'school',
+          link: 'View Activations', path: admin_activations_path,
+          content: "You have #{@school_activations} schools awaiting activation" }
+      ]
+    end
+
+    def prompt_for_issues_overdue?
```
ldodds, line 34: I think what you're doing here is fine, but what I would do here is memoize the value returned by the count methods, e.g:

```
def overdue_issues_count
  @overdue_issues ||= user.owned_issues.where.not(review_date: Date.current..).count
end
```

Then always use the methods rather than the member variables (e.g. `"You have #{overdue_issues_count} issues overdue for review"`). Just avoids any potential ordering issues in how methods are called in the class or template where variables are nil or queries are run twice.

app/components/admin/dashboard_prompt_component.rb
```diff
@@ -0,0 +1,105 @@
...
+                                       .count(&:has_schools_awaiting_activation?)
+    end
+
+    def lagging_data_sources_count
+      @lagging_data_sources = DataSource.where(owned_by: user).find_each.count(&:exceeded_alert_threshold?)
+    end
+
+    def missing_data_feed_readings_count
+      now = Time.current
+      @missing_data_feed_readings = AmrDataFeedConfig.enabled
```
ldodds: This is very similar to the query in stopped data feeds, so we should factor this out at some point, but doesn't have to be done here.

It's also quite slow, which makes viewing some of the dashboards take some time. I've added a Trello card to look at optimising this as a separate task. I think it should be possible to push all of the query into the database which might then be more efficient.

ldodds (approved):
Looks good.

PR #5191, opened 2026-04-14
Add import time index and change scope

ldodds:
Small bit of tuning for the import log report:

- Add an index on import_time as we're filtering on that
- Change the `with_warnings` scope to be more efficient.

### DIFF
--- a/app/models/amr_data_feed_import_log.rb
+++ b/app/models/amr_data_feed_import_log.rb
@@ -27,9 +27,7 @@ class AmrDataFeedImportLog < ApplicationRecord
 
   scope :errored,       -> { where.not(error_messages: nil) }
   scope :successful,    -> { where(error_messages: nil) }
-  scope :with_warnings, lambda {
-    where(id: AmrReadingWarning.select(:amr_data_feed_import_log_id))
-  }
+  scope :with_warnings, -> { where(id: AmrReadingWarning.select(:amr_data_feed_import_log_id)) }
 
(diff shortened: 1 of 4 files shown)


ldodds:
Deployed to test to try it out. No longer times out

app/models/amr_data_feed_import_log.rb
```diff
@@ -26,8 +27,11 @@ class AmrDataFeedImportLog < ApplicationRecord
 
   scope :errored,       -> { where.not(error_messages: nil) }
   scope :successful,    -> { where(error_messages: nil) }
-  scope :with_warnings, -> { includes(:amr_reading_warnings).where.not(amr_reading_warnings: { id: nil }) }
-  scope :since,         ->(date) { where('import_time >= ?', date) }
+  scope :with_warnings, lambda {
```
tbhi: think this would fit on one line?

app/models/amr_data_feed_import_log.rb
```diff
@@ -26,8 +27,11 @@ class AmrDataFeedImportLog < ApplicationRecord
...
+    where(id: AmrReadingWarning.select(:amr_data_feed_import_log_id))
```
tbhi: does distinct help?

tbhi (approved):

app/models/amr_data_feed_import_log.rb
ldodds: seems to be worse. will stick with this for now as its marked improvement on current

PR #5192, opened 2026-04-15
My x dashboard links

LukeGouldson:
Adds links to the left nav bar on admin dashboard: school groups, project groups, data sources, data feed configs and issues.


LukeGouldson:
I had two interesting bugs which I solved but there may be better ways

1) The School Groups and Project Groups would both appear highlighted when either one is selected, as they use the same controller AND route (index). Therefore turned off match_controller and added 'match_on_param' which only sets the school groups link to be selected when the value of the group_type is nil (so not when project groups is selected)

2) The Issues tab caused the whole nav bar to have a bullet point list. This is due to bootstrap 5 being enabled only for this page, I think. So now there is an option passed to the page nav for 'list-inline' adds a class to the page nav component's html. I guess the page nav hasn't been updated to bootstrap 5 yet, so my changes only allow the dashboard instance to work, not other instances of PageNav on the site.

ldodds:
Just reviewing the UX. I think now that the links in the navigation are working it would make sense for the prompts to link to those pages where possible (can do for issues, data sources, feeds). This will help tie things together otherwise at the moment you end up losing the left-hand nav.

app/components/navigation/admin_dashboard_component/admin_dashboard_component.html.erb
```diff
@@ -2,15 +2,15 @@
...
+  <% nav_sections.each do |section| %>
+    <% c.with_section options: { match_controller: false } do |s| %>
+      <% s.with_item name: section[:name],
+                     href: section[:path],
+                     match_on_param: section[:match_on_param] %>
+    <% end %>
   <% end %>
 <% end %>
+
+<%# however we don't want the match on param to apply to the whole section, so see if this can be implemented individually for items %>
```
ldodds: left over comment?

ldodds (commented):
Looks good, see my suggestion to revise the prompt links.

ldodds (approved):

PR #5196, opened 2026-04-17
Latest data received and out of date badge

LukeGouldson:
Adds 'latest data recieved' and 'out of date' columns to amr data feed configs index page (and therefore also to dashboards)

### DIFF
--- a/app/views/admin/amr_data_feed_configs/_overview_table.html.erb
+++ b/app/views/admin/amr_data_feed_configs/_overview_table.html.erb
@@ -12,13 +12,12 @@
   </thead>
   <tbody>
     <% configurations.each do |config| %>
-      <% date_time = config.amr_data_feed_readings.maximum(:updated_at) %>
       <tr>
         <td><%= link_to config.description, admin_amr_data_feed_config_path(config) %></td>
         <td class="no-wrap"><code><%= config.identifier %></code></td>
         <td><%= config.owned_by&.name %></td>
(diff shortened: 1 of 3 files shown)


app/views/admin/amr_data_feed_configs/_overview_table.html.erb
```diff
@@ -14,6 +16,16 @@
         <td><%= link_to config.description, admin_amr_data_feed_config_path(config) %></td>
         <td class="no-wrap"><code><%= config.identifier %></code></td>
         <td><%= config.owned_by&.name %></td>
+        <td><%= config.amr_data_feed_readings.maximum(:updated_at) %></td>
```
ldodds: Use `nice_date_times(config.amr_data_feed_readings.maximum(:updated_at))` as it gives a nice display of the time.

With sortable tables the sorting of dates can be problematic as the JS component uses lexical order by default, so we need to specify the sort order for each cell. Especially when we've formatted dates and times. So you should also revise the table cell to specify the value for sorting the column:

```
<td data-order="<%= config.amr_data_feed_readings.maximum(:updated_at).iso8601 %>
 <%= nice_date_times(config.amr_data_feed_readings.maximum(:updated_at)) %></td>
</td>
```

Might want to add a variable for the date-time.

ldodds (approved):
One display issue to address, but otherwise looks good 👍

PR #5198, opened 2026-04-17
Add new Contracts & Licensing index

ldodds:
Adds a new admin index page listing all the pages, reports and tools for the Contracts & Licensing feature.

- New /admin/commercial page with links to all relevant pages and reports
- Adds a new set of contract reports based on `TemporalRange` showing current, expired, expiring, future, recently added and provisional contracts
- Adds a new Unlicensed School report to list all schools without licenses

I still need to rework the licence reports to use similar approach as contracts but will get internal feedback on new pages first.

Invoicing section will be populated in next phase of work.


app/controllers/admin/commercial/contracts_controller.rb
```diff
@@ -71,6 +83,10 @@ def destroy
 
       private
```
urbanwide, line 87: I know it's in the admin interface but as action_name is derived from the request, it might be a good idea to put an extra check in place:

```
ALLOWED_SCOPES = %w[current expired expiring future recent].freeze

def load_contracts(scope)
  raise ArgumentError unless ALLOWED_SCOPES.include?(scope)

  @contracts = ::Commercial::Contract.public_send(scope).by_start_date
end
```

app/controllers/admin/commercial/contracts_controller.rb
```diff
@@ -3,12 +3,26 @@
...
 
+      def current = load_contracts(action_name)
+
+      def expired = load_contracts(action_name)
+
+      def expiring = load_contracts(action_name)
+
+      def future = load_contracts(action_name)
+
+      def provisional = local_contracts(action_name)
```
urbanwide: This should be load_contracts not local_contracts. Perhaps these actions should be exercised in a test too, if not already

app/views/admin/commercial/contracts/_contract_list.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= render Admin::HeaderNavComponent.new do |header_nav| %>
+  <% header_nav.with_header title: title %>
+  <% header_nav.with_button 'Contracts & Licensing', admin_commercial_path %>
+<% end %>
+
```
urbanwide, line 5: Wondering if there is a reason why this doesn't have tabbed navigation at the top - with all the different available filtered contracts in (as per licences)? i.e. future, expiring, current etc

app/views/admin/commercial/contracts/current.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= render 'contract_list', title: 'Current Contracts', contracts: @contracts do %>
+<div class="row">
```
urbanwide: It looks like these row and col divs, and the p tag are all common between these pages. Can they be moved up to the _contract_list.html partial?

```
<div class="row">
  <div class="col">
    <p>
       <%= yield %>
    </p>
  </div>
</div>

```

Or does it even need to yield at all, could just pass description as follows:

```
<%= render 'contract_list', title: 'Current Contracts', contracts: @contracts, description: 'This page lists all current contracts' %>
```

urbanwide (approved):
Looks good - with a few questions / comments

app/views/admin/commercial/contracts/_contract_list.html.erb
ldodds: I decided the tabbed version was too complicated, so am going to remove that. Will simplify the code as its then just a scope and date.

app/controllers/admin/commercial/contracts_controller.rb
ldodds: Embarrassing. Will add specs for this and another couple that I didn't add to system specs.

app/views/admin/commercial/contracts/current.html.erb
ldodds: Went with second option. Wasn't sure hope much tweaking each page might need but can always add that in later.

app/views/admin/commercial/show.html.erb
```diff
@@ -0,0 +1,86 @@
+<%= render Admin::HeaderNavComponent.new do |header_nav| %>
+  <% header_nav.with_header title: 'Contracts & Licensing' %>
+<% end %>
+
+<div class="row">
+  <div class="col-12">
+    <%= render Layout::GridComponent.new(classes: 'mb-4',
```
urbanwide, line 7: Sorry a bit late on this one but just realised you have two grid components on the page. Grid component should be clever enough to just keep adding blocks to the same grid and it will put the next two on new rows without needing two grid components.

Also, have you considered adding a new card type here (rather than using block) and making grid support it?

ldodds: Have already merged, but will look at changes in next revisions to the page.

urbanwide: Ok brill 👍🏻

PR #5200, opened 2026-04-20
Switch out annotate for annotaterb

urbanwide:
The annotate gem does not work with rails 8. This switches to the recommended fork annotaterb.

It's not a straight switch. The configuration is now in annotaterb.yml rather than auto_annotate_models.rake (now removed)

To annotate the models run:
`annotaterb models`

Note there are some changes in the model annotation order, which make sense to me. The sort order is:

```
id
other fields in alphabetical order
created_at
updated_at
othermodel_id
anothermodel_id
```


ldodds (approved):
Does the new gem still run the annotation when a migration is run?

urbanwide:
Yes, it appears to. Thanks

PR #5214, opened 2026-04-22
minimal annotaterb config

tbhi:
diffed with defaults and removed anything not changed then ran `bundle exec annotaterb models --force`

### DIFF
--- a/app/models/amr_data_feed_reading.rb
+++ b/app/models/amr_data_feed_reading.rb
@@ -1,3 +1,11 @@
+# Postgres autovacuum specific settings:
+# See: https://www.postgresql.org/docs/current/runtime-config-autovacuum.html
+# Applied using: ALTER TABLE amr_data_feed_readings SET (X = n)
+# autovacuum_vacuum_cost_delay = 0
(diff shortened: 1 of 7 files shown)


app/models/amr_data_feed_reading.rb
```diff
@@ -1,3 +1,11 @@
+# Postgres autovacuum specific settings:
```
tbhi, line 1: can't see a way of not moving to top

app/models/case_study.rb
```diff
@@ -2,21 +2,18 @@
...
 #
-#  id                :bigint(8)        not null, primary key
-#  organisation_type :string
-#  position          :integer          default(0), not null
-#  published         :boolean          default(FALSE), not null
-#  title             :string
-#  created_at        :datetime         not null
-#  updated_at        :datetime         not null
-#  created_by_id     :bigint(8)
-#  organisation_id   :bigint(8)
```
urbanwide, line 13: organisation_id should be there - this is in master, so maybe worth a merge

urbanwide (approved):

PR #5219, opened 2026-04-24
Provide summary of school specific pricing

ldodds:
Removes the school specific version of the pricing calculator in favour of just showing the current and estimated renewal fees on the school licences page.

Also reworks the pricing calculator to properly calculate an estimated renewal price. Previously it was assuming that schools would stay on same licensing and contract terms. But they may change to a different contract, e.g. if moving from free/funded place to self-funding. The calculator now does a better job of producing an estimate which is now displayed on the licenses page.


app/views/admin/commercial/pricing/show.html.erb
urbanwide: I know this isn't part of this PR, but these notes might be better using `hint:` as a parameter to the inputs, so they display in a standard way.

urbanwide (approved):
Just one comment about cosmetic stuff

PR #5224, opened 2026-04-27
initial metrics for group impact page

tbhi:


.ebextensions/cronjob.config
tbhi, line 1: reordered into sections based on frequency

.ebextensions/cronjob.config
```diff
@@ -34,107 +34,67 @@ files:
...
+            15 8 * * 1 root run-webapp-job issues:send_user_report
+            30 8 * * 1 root run-webapp-job targets:send_first_target
+            32 8 * * 1 root run-webapp-job recent_activities:send_email
+            30 6 * * 3 root run-webapp-job alerts:generate_subscriptions
+            0  1 * * 7 root run-webapp-job alerts:delete_alert_content
+            # ============================================================
+            # Monthly
+            # ============================================================
+            0 12 1  * * root run-webapp-job school:load_and_match_gias_schools
+            0 15 15 * * root run-webapp-job school_groups:generate_impact_reports
```
tbhi, line 92: this is the new job

app/services/school_groups/impact_report.rb
```diff
@@ -111,6 +111,10 @@ def enrolled_schools
       def enrolling_schools
         school_group.onboardings_for_group.incomplete.count
       end
+
+      def active_users
+        users.active.count + school_group.users.active.count
```
tbhi: perhaps last login date also be considered?

urbanwide: See users_logged_in_recently method - I probably should have named it better, apologies

tbhi: renamed users_logged_in_recently to active_users

lib/tasks/school_groups/generate_impact_reports.rake
```diff
@@ -0,0 +1,20 @@
...
+  task generate_impact_reports: :environment do
+    SchoolGroup.find_each do |school_group|
+      next if school_group.assigned_schools.active.empty?
+
+      impact_report_run = ImpactReport::Run.create!(school_group:, run_date: Date.current)
+      report = SchoolGroups::ImpactReport.new(school_group)
+      ImpactReport::Metric::OVERVIEW_METRICS.each do |metric_type|
+        value = report.overview.public_send(metric_type)
+        value = value.count if value.respond_to?(:count)
+        ImpactReport::Metric.create!(impact_report_run:, enough_data: true, metric_category: :overview, metric_type:,
```
ldodds: Missing `number_of_schools`. Should always be recording number of schools used to produce the metric. The basis for counting may vary across the different metrics but for the overview it will be the number of visible schools in the group.

.ebextensions/cronjob.config
urbanwide: Does this equate to 3pm on 15th? Should it be out of school hours (although not so it could clash with a potential deployment)?

lib/tasks/school_groups/generate_impact_reports.rake
```diff
@@ -0,0 +1,20 @@
...
+namespace :school_groups do
+  task generate_impact_reports: :environment do
+    SchoolGroup.find_each do |school_group|
+      next if school_group.assigned_schools.active.empty?
+
+      impact_report_run = ImpactReport::Run.create!(school_group:, run_date: Date.current)
+      report = SchoolGroups::ImpactReport.new(school_group)
+      ImpactReport::Metric::OVERVIEW_METRICS.each do |metric_type|
+        value = report.overview.public_send(metric_type)
+        value = value.count if value.respond_to?(:count)
```
urbanwide: I can't decide if this is a bit risky, in case something under the hood changes. Might be best if everything called returns the final metric (rather than having to call count on it) ultimately.

spec/services/school_groups/impact_report_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+
+      it 'excludes users from non-visible schools' do
+        hidden_user = create(:user, school: create(:school, visible: false, school_group:))
+        expect(overview.users).not_to include(hidden_user)
+      end
+    end
+
+    describe '#active_users' do
+      it 'returns the count of users who have logged in within the last three months' do
+        active_user   = create(:user, school: visible_school, last_sign_in_at: 1.month.ago)
```
urbanwide: I am not sure if this is a matter of style but I prefer test setup to be done in a before block or lets and 'it' blocks to contain just tests.

(same in a few other places in this file)

lib/tasks/school_groups/generate_impact_reports.rake
urbanwide: I don't know if this is a style thing, but I'd prefer this to be written:
`impact_report_run.metrics.create!(enough_data: true, metric_category: :overview, metric_type:, value:)`

(also with number_of_schools as mentioned by Leigh)

urbanwide (commented):
Looks good - a few comments. Presume you will go on to put the task contents into a service as it gets bigger.

.ebextensions/cronjob.config
tbhi: yeah - the card says to try in the afternoon so not to overlap with morning jobs - not sure if there's a better time

lib/tasks/school_groups/generate_impact_reports.rake
tbhi: added

urbanwide (approved):

PR #5239, opened 2026-04-29
Handle custom contract lengths and prorata contracts in pricing calculations

ldodds:
Revises price calculation to apply custom contract lengths and pro_rata licence periods.

* Prices are always for a full year (365 days, see below)
* Standard contracts (`licence_period='contract'`) can have start/end dates which define a duration of more or less than a year. So prices need to be multiplied up by the length of the licensed period. 
* Custom contracts (`licence_period=`custom`) generate licences with lengths based on the `licence_years` field. Again, prices need to be multiplied up based on the length of the licensed period
* Standard contracts can have provision for pro-rata licences (`invoice_terms='pro_rata'`. These licences may run for a short period than the total length of the contract (e.g. school joins mid-way through the contracted period so should only be charged for the shorter period). So prices need to be prorated based on the length of the licenced period within the contract.

When calculating licensed periods we ignore leap days, so a year is always 365 days regardless of the number of days in the licenced period. This applies to calculating the number of days in a period and in generating licence end dates for custom contracts.

The `ContractPriceCalculator` now generates prices using the above rules and there is a revised set of specs that should cover all scenarios.

The `PriceCalculator`, which is just used to produce estimates, now applies the contract length rules but doesn't calculate prorated licences. Its only used for estimates and for now we're assuming the estimated is for the full contracted period.

I've also removed used of the `PriceCalculator` on the school licence pages. We're revising the UI here to not show estimated renewal prices, instead we'll show actual pricing for any renewals. This will be added in a later PR. For now, removing that allows the `PriceCalculator` code to be simplified.


app/models/commercial/contract.rb
```diff
@@ -153,6 +153,15 @@ def invoiced?
 
     private
 
+    def leap_days_between(period_start, period_end)
+      (period_start.year..period_end.year).count do |year|
+        next false unless Date.leap?(year)
```
tbhi: perhaps `Date.leap?(year) && Date.new(year, 2, 29).between?(period_start, period_end)`

app/models/commercial/licence.rb
```diff
@@ -87,6 +93,15 @@ def deletable?
       !invoiced?
     end
 
+    private_class_method def self.leap_days_between(period_start, period_end)
```
tbhi, line 96: reuse above?

app/models/meter.rb
```diff
@@ -136,6 +136,10 @@ class Meter < ApplicationRecord
 
   scope :for_admin, ->(admin) { where(school: { school_groups: { default_issues_admin_user: admin } }) }
 
+  scope :main_meter_counts_by_school, lambda  {
+    where(pseudo: false, meter_type: Meter::MAIN_METER_TYPES, active: true).group(:school_id)
```
tbhi: overlap with main_meter.active

app/services/commercial/contract_price_calculator.rb
```diff
@@ -28,11 +29,42 @@ def calculate_per_school
...
+      ).to_f / 365.0
+    end
+
+    def prorata_multiplier(licence_start_date, licence_end_date)
+      return 1.0 if @contract.custom? || @contract.full?
+
+      licence_days = Commercial::Licence.licence_period_days(licence_start_date, licence_end_date).to_f
+      full_days = Commercial::Licence.licence_period_days(@contract.start_date, @contract.end_date).to_f
+
+      licence_days.to_f / full_days
```
tbhi: `.to_f` looks not required

tbhi (approved):

app/models/commercial/contract.rb
ldodds: Old method, I moved the code to Licence. Have applied the suggested change there.

PR #5242, opened 2026-04-29
 fix csv and user filtering on admin meters report 

tbhi:
- changes behaviour of the report (/admin/reports/admin_user_meter_report) so it will not load data until the filter button is pressed - as it seems quite slow - perhaps should be paginated in future - previously showed Claudia's meters by default
- allows filtering by school group for any admin
- fixes use of user parameter for csv download

### DIFF
--- a/app/views/admin/reports/base_meter_reports/_report_filters.html.erb
+++ b/app/views/admin/reports/base_meter_reports/_report_filters.html.erb
@@ -29,7 +29,7 @@
                         controller: controller_path,
                         action: :index,
                         only_path: true,
-                        **params.permit(:school_group, :user, :meter_type).merge(format: :csv)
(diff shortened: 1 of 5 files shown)


app/views/admin/reports/base_meter_reports/_report_filters.html.erb
```diff
@@ -29,7 +29,7 @@
                         controller: controller_path,
                         action: :index,
                         only_path: true,
-                        **params.permit(:school_group, :user, :meter_type).merge(format: :csv)
+                        **params.permit(:school_group, :admin, :meter_type).merge(format: :csv)
```
tbhi, line 32: this is the main fix

ldodds (approved):

PR #5243, opened 2026-04-30
My meters section

LukeGouldson:
Adds navigation and templates for ‘Meters with…’, ‘New data for inactive meters’, ‘Meters & readings’, ‘Meters for admin user’, ‘baseload anomalies’, ‘manually read meters’ and ‘pupil number updates’. Not yet added to dashboard navbar as the reports need optimising first


ldodds (commented):
Looks good. 

Some of the reports are relatively fast when filtered by admin, rather than showing all issues. So I think it might be OK to start linking them in. Take a look at the `frequency` method in the controller, if its `daily` then they should definitely be fine. The blank/lagging/zero reads meter reports are the worst so maybe avoid for now. 

Baseload anomalies isn't here, does that need similar changes?

LukeGouldson:
Bit of a conflict as another PR edited `base_meter_reports/index.html.erb`

ldodds:
See my earlier comment about adding in a few of the links and whether baseload/gas anomaly reports need any changes.

ldodds (approved):
Looks good. There's still some improvements that can be made to the newly linked reports when viewed through the dashboard:

- remove the admin user filter
- remove the admin name column

Suggest doing those in a separate PR

PR #5245, opened 2026-05-01
generate potential savings metrics for group impact page

tbhi:


ldodds (approved):

lib/tasks/school_groups/generate_impact_reports.rake
```diff
@@ -7,11 +7,14 @@ namespace :school_groups do
...
-       [ImpactReport::Metric::ENGAGEMENT_METRICS, :engagement]].each do |metrics, metric_category|
-        metrics.each do |metric_type|
-          run.metrics.create!(enough_data: true, number_of_schools: report.overview.visible_schools, metric_category:,
-                              metric_type:, value: report.public_send(metric_category).public_send(metric_type))
+      %i[overview engagement potential_savings].each do |metric_category|
+        ImpactReport::Metric.metrics(metric_category).each do |metric_type|
+          value = report.value(metric_category, metric_type)
+          next if value.nil?
+
+          run.metrics.create!(enough_data: true, metric_category:, metric_type:,
```
urbanwide: We need fuel_type populating here, as the front end needs to pick metrics based on fuel type.
Also - does enough_data need to be considered for potential savings?

app/services/school_groups/impact_report/potential_savings.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
```
urbanwide, line 1: Does it make sense to also break out overview, engagement etc into their own files at this point too, for consistency?

urbanwide (commented):
Looks good - a couple of questions / comments

lib/tasks/school_groups/generate_impact_reports.rake
ldodds: For consistency and to ensure transparency in the admin interface, we'd have to add a metric with enough_data as false for any of the metrics not generated here. If the underlying alerts haven't run, then there will be no data for a specific priority.

On the group advice pages and dashboards we just display all priorities, but for the impact report we might want to only show those with >1 school. But that's for the UX I think.

db/schema.rb
```diff
@@ -30,7 +30,7 @@
...
+  create_enum "impact_report_metric_types", ["visible_schools", "data_visible_schools", "users", "active_users", "pupils", "enrolled_schools", "enrolling_schools", "activities", "actions", "points", "targets", "total_savings", "electricity_out_of_hours_gbp", "electricity_out_of_hours_co2", "electricity_out_of_hours_kwh", "electricity_peak_gbp", "electricity_peak_co2", "electricity_peak_kwh", "electricity_use_gbp", "electricity_use_co2", "electricity_use_kwh", "electricity_baseload_gbp", "electricity_baseload_co2", "electricity_baseload_kwh", "solar_panels_gbp", "solar_panels_co2", "solar_panels_kwh", "thermostatic_control_gbp", "thermostatic_control_co2", "thermostatic_control_kwh", "gas_out_of_hours_gbp", "gas_out_of_hours_co2", "gas_out_of_hours_kwh", "gas_use_gbp", "gas_use_co2", "gas_use_kwh", "heating_early_gbp", "heating_early_co2", "heating_early_kwh", "heating_down_gbp", "heating_down_co2", "heating_down_kwh", "heating_off_gbp", "heating_off_co2", "heating_off_kwh", "storage_heaters_off_gbp", "storage_heaters_off_co2", "storage_heaters_off_kwh", "insulate_pipes_gbp", "insulate_pipes_co2", "insulate_pipes_kwh"]
```
ldodds: As the metric has a `fuel_type` then the enum values could be normalised here a bit more, e.g. `out_of_hours_kwh` which would be recorded for both electricity and gas. That keeps the enum tidier I think. Some metrics like `baseload` are already fuel type specific so don't necessarily need a prefix.

urbanwide: Also do we only need the gpb metric given we are talking about potential savings? I can't remember what we said now.

ldodds: I said to generate all three values, so we have more options when it comes to the UX. We might just end up using gbp for now.

urbanwide: Ok, makes sense. Thanks

lib/tasks/school_groups/generate_impact_reports.rake
```diff
@@ -5,15 +5,9 @@ namespace :school_groups do
...
+      generator = SchoolGroups::ImpactReport::Generator.new(school_group)
       run = ImpactReport::Run.create!(school_group:, run_date: Date.current)
-      [[ImpactReport::Metric::OVERVIEW_METRICS, :overview],
-       [ImpactReport::Metric::ENGAGEMENT_METRICS, :engagement]].each do |metrics, metric_category|
-        metrics.each do |metric_type|
-          run.metrics.create!(enough_data: true, number_of_schools: report.overview.visible_schools, metric_category:,
-                              metric_type:, value: report.public_send(metric_category).public_send(metric_type))
-        end
-      end
+      generator.metrics.each { |attributes| run.metrics.create!(**attributes) }
```
urbanwide: You could create a method on generator that would encapsulate this? Something like:

```
def create_metrics!
  run = ImpactReport::Run.create!(school_group:, run_date: Date.current)
  metrics.each { |attributes| run.metrics.create!(**attributes) }
end

```

Then call here:

```
generator = SchoolGroups::ImpactReport::Generator.new(school_group)
generator.create_metrics!
```

db/migrate/20260501135818_add_potential_savings_to_metric_type_enum.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class AddPotentialSavingsToMetricTypeEnum < ActiveRecord::Migration[8.1]
+  def up
+    SchoolGroups::ImpactReport::PotentialSavings::METRICS.each do |metric|
```
urbanwide: When I'm running this migration, I'm getting:

```
NameError: uninitialized constant SchoolGroups::ImpactReport::PotentialSavings::METRICS (NameError)

    SchoolGroups::ImpactReport::PotentialSavings::METRICS.each do |metric|
                                                ^^^^^^^^^
```

urbanwide (commented):
I'm getting an error when running the migration? (see comment + one other)

db/migrate/20260501135818_add_potential_savings_to_metric_type_enum.rb
tbhi: fixed - not updated since last refactor

urbanwide:
I'm good with this being merged now if you are @ldodds ?

PR #5247, opened 2026-05-04
Impact reporting admin interface

urbanwide:
Does a number of things:

* Fleshes out the /admin/impact_reports admin interface
* Hooks in the new settings to the impact report page, so different sections can be switched on / off feature settings changed
* Component-ises all the different sections of the impact report page (apart from notes) and adds spec and previews.
* Adds school group settings nav to the admin/edit page
* Updates PageNav to work with bootstrap 5
* Updates TogglerDropdown component to work with bootstrap 5 (used in the admin interface previews)
* Adds impact report config items to the manage school group menu / school group settings left nav


app/controllers/admin/impact_reports_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    def edit
+      @configuration = @school_group.impact_report_configuration || @school_group.build_impact_report_configuration
+    end
+
+    def update
+      @configuration = @school_group.impact_report_configuration || @school_group.build_impact_report_configuration
+      @configuration.attributes = configuration_params if params[:impact_report_configuration].present?
+
+      if @configuration.save
+        redirect_to url_from(params[:redirect_back]) || admin_impact_reports_path,
```
ldodds: I visited from the Manage Group menu and ended up on the group dashboard which isn't what I was expecting. Think it would be better to either redirect to the impact report, to immediately review changes, or back to the form itself. Both of those makes it easier to (pre)view the changes.

ldodds (approved):
Looks good. One comment about the redirect behaviour

app/controllers/admin/impact_reports_controller.rb
urbanwide: Good idea, I've changed it to go back to the form itself, feels better this way.

PR #5255, opened 2026-05-06
Impact reports - use overview metrics

urbanwide:


app/components/impact_reports/overview/stats_component.rb
```diff
@@ -4,7 +4,7 @@ module ImpactReports
   module Overview
     class StatsComponent < ImpactReports::BaseComponent # rubocop:disable ViewComponent/PreferComposition
       def show_enrolling?
-        @impact_report.overview.enrolling_schools > @impact_report.overview.enrolled_schools
+        run.overview(:enrolling_schools).value >= run.overview(:enrolled_schools).value
```
ldodds: There's a bug here, which is partly due to ambiguity in the spec. We have enrolled_schools which is implemented as number schools enrolled in last 12 months, and enrolling_schools which is those still enrolling.

If no schools have joined in last 12 months and no enrolling, then we're still seeing this section.

I think the check might be as simple as is it non-zero?

We also have to decide how to ranke enrolled_schools and enrolling_schools. At the moment "enrolled_schools" is shown if "enrolling_schools" is not. But both might be nil. Maybe do: enrolling first if non zero, then enrolled if non zero. So we may end up with only three blocks at the top.

ldodds (commented):
Code looks ok, one bug to fix.

Not really important for this initial PR but for the other sections you're going to want to filter out Metrics where `enough_data` is false, so you're only looking at displaying/ranking metrics with sufficient data to display. 

Adding a `valid` or `displayable` scope to `Metrics` which can be used to filter metrics (e.g. `enough_data && value.present?`) might help, but then calling code need to allow for missing metrics when applying display logic.

app/components/impact_reports/overview/stats_component/stats_component.html.erb
```diff
@@ -5,48 +5,39 @@
                                        component_classes: 'theme theme-dark') do |grid| %>
     <% grid.with_stats_card do |card| %>
       <%= card.with_header title: impact_t('overview.cards.schools.title') %>
-      <%= card.with_figure number_with_delimiter(@impact_report.visible_schools_count) %>
+      <%= card.with_figure number_with_delimiter(overview(:visible_schools).value) %>
       <%= card.with_subtext do
             impact_t('overview.cards.schools.subtext',
-                     count: @impact_report.data_visible_schools_count)
+                     count: overview(:data_visible_schools).value)
```
ldodds, line 11: What if non of the schools are data visible yet? E.g. new group is joining but data isn't available yet. Maybe have a different message "With access to our education programme and pending energy analysis" (wording will need review)

app/components/impact_reports/overview/stats_component/stats_component.html.erb
```diff
@@ -5,48 +5,39 @@
...
           end %>
     <% end %>
 
     <% grid.with_stats_card do |card| %>
       <%= card.with_header title: impact_t('overview.cards.users.title') %>
-      <%= card.with_figure number_with_delimiter(@impact_report.overview.users) %>
+      <%= card.with_figure number_with_delimiter(overview(:users).value) %>
```
ldodds, line 17: Just double-checked and think the scope used for the metric needs tweaking. Maybe one for @tbhi:

- we do want cluster users
- we don't want unconfirmed users
- we don't want users that are not marked as `active`, as they've been blocked/disabled or marked as inactive when a school has been archived/deleted
- we don't want to count pupil logins (but OK to count `student` role).

app/components/impact_reports/overview/stats_component.rb
urbanwide: I took an educated guess at the logic here, thanks for clarifying, have updated 👍🏻

app/components/impact_reports/overview/stats_component/stats_component.html.erb
urbanwide: Believe Tristan is looking at this

app/components/impact_reports/overview/stats_component/stats_component.html.erb
urbanwide: Have updated to include the new text for 0 value here.

urbanwide:
> Not really important for this initial PR but for the other sections you're going to want to filter out Metrics where `enough_data` is false, so you're only looking at displaying/ranking metrics with sufficient data to display.
> 
> Adding a `valid` or `displayable` scope to `Metrics` which can be used to filter metrics (e.g. `enough_data && value.present?`) might help, but then calling code need to allow for missing metrics when applying display logic.

OK thanks. Was leaving this out until we got to the metrics that needed filtering, however have now added support as suggested here too just for consistency.

app/components/impact_reports/overview/stats_component/stats_component.html.erb
```diff
@@ -1,52 +1,43 @@
 <%= tag.div id: id, class: classes do %>
   <%= render Layout::GridComponent.new(id: 'overview-cards',
-                                       cols: 4, classes: 'mb-4',
+                                       cols: cols, classes: 'mb-4',
                                        cell_classes: 'mb-4',
                                        component_classes: 'theme theme-dark') do |grid| %>
     <% grid.with_stats_card do |card| %>
```
urbanwide: This needs a:

`if: overview(:visible_schools).displayable?
`
Forgot to commit it but just doing something else

ldodds (approved):
Happy for you to merge after the displayable check is added for visible schools.

PR #5259, opened 2026-05-08
Change meters for admin user to include all meters

LukeGouldson:
Adds an 'active' filter to this report so that users can view active, inactive or all meters. Defaults to active


ldodds (approved):

ldodds:
Going to merge so we can release today.

PR #5265, opened 2026-05-11
Handle SolarEdge Pending Site Status

ldodds:
When a SolarEdge site is still in "Pending" status the data period API call will return a nil for data from which data is available.

This PR:

- ensures that nil dates are parsed correctly, allowing the site details to be properly read and stored (this allows admins to view the site status)
- cause the API client to raise an exception if there is an incomplete data period. This stops us from attempting to load data when there isn't any available

Added a spec for the parsing behaviour. There's already specs in calling code to ensure that exceptions are caught and logged correctly.

### DIFF
--- a/lib/data_feeds/solar_edge_api.rb
+++ b/lib/data_feeds/solar_edge_api.rb
@@ -33,10 +33,8 @@ module DataFeeds
     end
 
     def site_start_end_dates(site_id)
-      dates = get_data("/site/#{site_id}/dataPeriod")
-      start_date = Date.parse(dates['dataPeriod']['startDate']) if dates['dataPeriod']['startDate']
-      end_date = Date.parse(dates['dataPeriod']['endDate']) if dates['dataPeriod']['endDate']
-      [start_date, end_date]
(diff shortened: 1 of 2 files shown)


lib/data_feeds/solar_edge_api.rb
```diff
@@ -33,7 +34,9 @@ def sites
 
     def site_start_end_dates(site_id)
       dates = get_data("/site/#{site_id}/dataPeriod")
-      [Date.parse(dates['dataPeriod']['startDate']), Date.parse(dates['dataPeriod']['endDate'])]
+      start_date = Date.parse(dates['dataPeriod']['startDate']) if dates['dataPeriod']['startDate']
+      end_date = Date.parse(dates['dataPeriod']['endDate']) if dates['dataPeriod']['endDate']
+      [start_date, end_date]
```
tbhi: perhaps ```
def site_start_end_dates(site_id)
  period = get_data("/site/#{site_id}/dataPeriod").fetch('dataPeriod', {})
  %w[startDate endDate].map { |key| Date.parse(period[key]) if period[key] }
end
```

tbhi (approved):

PR #5269, opened 2026-05-12
Fix bug in SolarEdge loading

ldodds:
This fix: https://github.com/Energy-Sparks/energy-sparks/pull/5265 included a new bug. It was only checking for the dataPeriod for the server if a start/end date wasn't provided when attempting to load data.

However we only do this when first loading from an installation or when doing a reload. Otherwise we're loading all recent data. As the dataPeriod wasn't being checked, we were always throwing an exception.

Would have been caught by better test coverage. So I have:

- reworked the API client specs to use Faraday test stubs to match other clients
- added additional specs for the `smart_meter_data` method, ensuring that is confirms that the `dataPeriod` and `energyDetails` API clients are both called
- added variants to test behaviour with and without an explicit date range

This improves the test coverage and confirms that the bug is fixed: we now always check the available `dataPeriod`.

I've also fixed another minor bug in the SolarEdgeController to help with debugging requests in the browser. And also ensured that we only attempt to do an immediate load if we have a valid `dataPeriod`.

Have also confirmed, via manual testing that data is loading as expected, and that new installations can be setup correctly, even when pending


app/services/solar/solar_edge_installation_factory.rb
```diff
@@ -3,53 +3,55 @@
...
 
       installation.update(information: information)
 
       # Retrieve two days worth of data, just to get the meters set up and ensure some data comes back
-      SolarEdgeDownloadAndUpsert.new(
-        installation: installation,
-        start_date: first_reading_date,
-        end_date: first_reading_date + 1.day
-      ).perform
+      if first_reading_date
```
ldodds, line 48: the only functional change

lib/data_feeds/solar_edge_api.rb
```diff
@@ -178,7 +179,7 @@ def process_raw_values(raw_values)
     end
 
     def dates(meter_id, start_date, end_date)
-      sd, ed = site_start_end_dates(meter_id) if start_date.nil? || end_date.nil?
+      sd, ed = site_start_end_dates(meter_id)
```
ldodds, line 182: Fix for the newly introduced loading bug, always check dataPeriod.

urbanwide (approved):

PR #5270, opened 2026-05-12
Increase batch insert size for unvalidated readings

ldodds:
Increases batch insert size from 100 to 500.

Looking at last 24 hours or so, only a few percent of files contain more than 500 rows. So this keeps batching in place for the very largest files but allows the majority (95%+) to be inserted in one statement.


ldodds:
Deployed to test to check impact

tbhi (approved):

PR #5281, opened 2026-05-12
Add format_price helper and use on commercial pages

ldodds:
Replace use of the FormatUnit helper when displaying product and contract pricing. The existing helper rounds to significant figures which isn't correct for driving invoicing, etc.

Added a custom helper to do the formatting rather than using `number_to_currency` as that doesn't handle the formatting options I needed. Thought it was better to have a single helper rather than have the public pricing page use different rules.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,6 +1,7 @@
 module ApplicationHelper
   include Pagy::Frontend
   include ActionView::Helpers::TagHelper
+  include ActionView::Helpers::NumberHelper
 
   def nice_date_times(datetime, options = {})
     return '' if datetime.nil?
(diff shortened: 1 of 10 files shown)


app/helpers/application_helper.rb
```diff
@@ -265,6 +266,22 @@ def format_school_time(school_time)
     format('%04d', school_time).insert(2, ':')
   end
 
+  def format_price(price, decimals: true)
```
urbanwide, line 269: Was going to ask if you'd seen the option round_mode: :down for number_to_currency i.e.

```
> helper.number_to_currency(120000.599, unit: '£', precision: 2, round_mode: :down)
 => "£120,000.59"
```

But then I saw that you need to do this in your test:

`expect(helper.format_price(1234.567)).to eq('£1,234.567')
`

So I can see why you have not used number_to_currency now. Just thought I would leave a comment seeing as though I'd looked.

urbanwide (approved):
Looks good. I looked at number_to_currency too and can see why you've not used it now (see comment).

PR #5283, opened 2026-05-13
Allow filtering of contract and licence list by date

ldodds:
Updated the lists of Contracts linked from /admin/commercial to include date filters for lists where this would be useful (e.g. expired, expiring)

Revise the list of Licences linked from same page to use the same style of presentation as the Contracts (controller actions, rather than tabbed view). 

This will make it easier to add more lists later, optimise the individual pages and add additional filters if needed.

### DIFF
--- a/app/views/admin/commercial/contracts/_contract_list.html.erb
+++ b/app/views/admin/commercial/contracts/_contract_list.html.erb
@@ -11,6 +11,22 @@
   </div>
 </div>
 
+<% if local_assigns[:date_label] %>
+  <div class="row mb-4">
+    <div class="col">
+      <%= simple_form_for :filters, url: path, method: 'GET', html: { class: 'form-inline' } do |f| %>
+        <%= f.input :date,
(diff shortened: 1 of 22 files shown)


app/views/admin/commercial/contracts/_contract_list.html.erb
```diff
@@ -11,6 +11,22 @@
...
+  <div class="row mb-4">
+    <div class="col">
+      <%= simple_form_for :filters, url: path, method: 'GET', html: { class: 'form-inline' } do |f| %>
+        <%= f.input :date,
+                    as: :tempus_dominus_date,
+                    default_date: @date,
+                    label: date_label,
+                    label_html: { class: 'd-block mr-2' },
+                    required: false %>
+        <%= f.submit 'Filter', class: 'btn btn-primary btn-sm ml-auto' %>
```
tbhi: ms-auto looks better for bootstrap 5, also gap seems quite large

tbhi (approved):

PR #5285, opened 2026-05-13
Integrate potential savings metrics

urbanwide:
@ldodds would you mind having a quick look at this please? The tests aren't written / are broken but I'm wondering if I have done the metric picking right before I carry on and tidy up.

At the moment, I'm only looking at the gbp metrics, sorting them into fuel types, then by value descending. Then picking one from each of the fuel types until I have 3.

Not sure what the plan is for using the non-gbp metrics but It might be easier for the metric to have the unit as another field, rather than it being within the metric_type label. I'm currently splitting out the unit from the metric_type at the moment, which feels clunky.


ldodds:
I suggested generating all 3 metrics (kwh, gbp, co2) in case we needed them. If we don't then we could just simplify the metric generation down to just the gbp values. 

I suspect that "these are the costs savings we've identified" is likely to have the biggest impact. So perhaps we can confirm that in internal review, then streamline the metrics after. So fine to split out metric type as you are now.

In terms of metric selection:

- ignore storage heaters. There's so few schools with them that this is just going to end up not showing or just having a savings for 1-2 schools. we're not showing any storage heater metrics in the energy efficiency section
- suggest just taking the top saving each for electricity and gas for now, so just two metrics
- in the spec/wireframe I suggested picking solar to give variety, so maybe fall back to that to give two card, if there's only one of the above
 
I think we need more discussion about what to do if:

- there aren't any potential savings, or they are all for single fuel type
- the savings are for a small number of schools in the group
- there's a solar saving but its less than some of the other electricity / gas savings we've identified, or if its only for a small % of schools
- there's several big savings for electricity or gas, so showing an additional metric or two would be better

So just pick the top one each for electricity and gas for now.

This section is meant to be a prompt for them to click through to the full list of opportunities, so that's the key action here, rather than trying to get a perfect ranking.

ldodds (approved):

PR #5287, opened 2026-05-14
Update admin meter statuses

LukeGouldson:
An after party task which changes:

Meters with statuses 'Comms issue - requires site visit', 'Comms issue - site visit planned', 'Comms issue - under investigation',  'Meter issue - action required by Energy Sparks', 'Meter issue - action required by school/MAT/Council', and 'Meter issue - action required by supplier/procurement/MOP' to have status 'Meter issue raised'

And meters with status 'Meter no longer required' to have status 'Not required'


LukeGouldson:
Just to note, the card also suggests adding a delete button to the meter statuses page, but it seems that this already exists

ldodds (approved):

PR #5289, opened 2026-05-14
Changed logic for alerts for group users

LukeGouldson:
In the alert column on the user_list_component, group admins and group managers now have either 'No' or 'Yes (X)' where X is the number of schools they have alerts on for.


ldodds (commented):
Change looks fine but can you add a test to the component spec? You could use the `'it contains the expected data table'` shared example to do a better test of the rendering with and without group users and alerts.

ldodds (approved):

PR #5301, opened 2026-05-19
Add caching to admin dashboards

LukeGouldson:
Adds caching to the quick link and prompt components rendering on the dashboard homepage. Prompts have a timestamp for when the last issue was updated


LukeGouldson:
Not sure how to go about testing this, other than printing the keys in the terminal. Also I believe caching works differently on developer env/test/production? Let me know if there's a good way to write specs for this.

ldodds (approved):

PR #5303, opened 2026-05-19
Changed submit to button

LukeGouldson:
Replaces submit button with standard button, so the qucik links form no longer submits upon pressing. Therefore a new tab can be opened as many times as desired


LukeGouldson:
This has had the effect of replacing the query param 'commit' with the param 'button' on the MPXN and URN pages. I couldn't find a convenient way to fix this, but perhaps it's not an issue? Let me know

ldodds (approved):

PR #5320, opened 2026-05-21
Impact report notes updates

urbanwide:
Basic updates to make the notes viable.

Almost certainly don't have the date ranges right - sorry didn't have time to dig round for the proper way of doing it...


ldodds (approved):
Lets go with this for now as an improvement and I'll take a look at the date ranges again when I add the card for the extra footnote metric

PR #5322, opened 2026-05-22
Add exception report prompts

ldodds:
Adds prompts to the commercial index page to highlight content from the exception reports.


app/components/commercial/exception_report_prompt_component/exception_report_prompt_component.html.erb
```diff
@@ -0,0 +1,30 @@
+<%= tag.div id: id, class: classes do %>
+  <%= render PromptListComponent.new do |list| %>
+    <% list.with_title(level: 3) { 'Exception reporting' } %>
+    <% if unlicensed_schools.positive? %>
```
LukeGouldson, line 4: Could implement this section with an array of hashes, with checks for whether each one should be displayed found in the rb file? To save repeated lines.

LukeGouldson (approved):

app/components/commercial/exception_report_prompt_component/exception_report_prompt_component.html.erb
ldodds: Will leave for now and refactor when we add more reminders. Will need to revisit this component when it comes to implementing invoicing.

PR #5330, opened 2026-05-27
Add ability to tag issues and notes

LukeGouldson:
Adds the IssueTag model, which has an association with Issue. Initial tags are 'group review', 'onboarding status' and 'solar setup'.

'Group review' has a system id and therefore is not deletable by admin users - other tags which are created by admins can be deleted.

Issues can be filtered by tag


db/migrate/20260522074513_create_issue_tags_issues.rb
ldodds, line 1: Not a big issue but this could have been combined with the other migration.

app/views/admin/issue_tags/index.html.erb
```diff
@@ -0,0 +1,37 @@
...
+      </thead>
+      <tbody>
+        <% @issue_tags.each do |issue_tag| %>
+          <tr>
+            <td><%= issue_tag.label %></td>
+            <td class="fit nowrap text-end">
+              <%= link_to 'View issues with this tag', admin_issues_path(issue_tag:),
+                          class: 'btn btn-secondary btn-sm edit' %>
+              <%= link_to 'Edit', edit_admin_issue_tag_path(issue_tag), class: 'btn btn-secondary btn-sm edit' %>
+              <%= link_to 'Delete',
```
ldodds: You the DeleteButtonComponent here so the button only appears if its possible to delete the tag

app/views/admin/issue_tags/index.html.erb
```diff
@@ -0,0 +1,37 @@
+<div class="d-flex justify-content-between align-items-center">
+  <h1>Issue Tags</h1>
+</div>
+
+<%= link_to 'New issue_tag', new_admin_issue_tag_path, class: 'btn btn-primary mb-2' %>
+
+<div id='issue-tags-list'>
+  <% if @issue_tags.any? %>
+    <table class="table table-sorted">
+      <thead>
```
ldodds, line 10: Suggest adding the system id so we can see that

app/views/admin/issue_tags/index.html.erb
```diff
@@ -0,0 +1,37 @@
...
+        <tr>
+          <th>Label</th>
+          <th></th>
+        </tr>
+      </thead>
+      <tbody>
+        <% @issue_tags.each do |issue_tag| %>
```
ldodds: Think you're relying on the default index action on the controller to just load the tags. It'd be better to explicitly load them with an order (sorted by label)

spec/models/issue_tag_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+require 'rails_helper'
+
+RSpec.describe IssueTag do
+  describe 'deletable' do
+    let!(:sys_issue_tag) { create(:issue_tag, system_id: :sys_tag) }
+    let!(:other_issue_tag) { create(:issue_tag) }
+
+    it 'deletes tags without a system id' do
+      other_issue_tag.destroy
+      expect(described_class.count).to eq(1)
```
ldodds: You can write specs like:

```
expect { other_issue_tag.destroy }.to change(described_class, :count).by(1)
```

And:

```
expect { sys_issue_tag.destroy }.not_to change(described_class, :count)
```

app/views/admin/issue_tags/index.html.erb
ldodds, line 1: Need a link to this index page on admin/index. Maybe add under "Other" for now.

ldodds (changes requested):
Couple of small changes needed, but this is looking great. Can review again if you fix today.

ldodds (approved):

PR #5331, opened 2026-05-28
Add button to confirm contracts

ldodds:
Adds a button to the contract page and component to allow a contract (and its licences) to be immediately confirmed.

Avoids need to use the edit contract form for a common task.


app/models/commercial/contract.rb
```diff
@@ -37,6 +37,10 @@
 #  fk_rails_...  (created_by_id => users.id)
 #  fk_rails_...  (product_id => commercial_products.id)
 #  fk_rails_...  (updated_by_id => users.id)
+#
+#  fk_rails_...  (created_by_id => users.id)
```
tbhi: duplicated?

app/controllers/admin/commercial/contracts_controller.rb
```diff
@@ -111,6 +111,16 @@ def destroy
         end
       end
 
+      def confirm
+        if @contract.update(status: :confirmed)
```
tbhi, line 115: worth  a transaction for the update & cascade?

tbhi (approved):

app/models/commercial/contract.rb
ldodds: I can't figure out why annotate keeps adding these duplicate rows. Seems to happen again whenever it runs. Deleted content again for now.

app/controllers/admin/commercial/contracts_controller.rb
ldodds: There's a transaction wrapping the cascades, I think that's OK for now? Will add a card to look at that, as it'd need to be done for create and update too.

PR #5332, opened 2026-05-28
Update Funder reports and related pages

ldodds:
Improve how we report on Funder information using the new Contract and Licensing data.

- Adds a new "Contract Holders" report that shows the numbers of schools per contract holder. The format is consistent with the current "Funder allocations" summary
- Adds a summary of licensed schools to the funder pages, along with some smaller display and navigation tweaks
- Adds a new "Contract Holder" column to the Funder Allocation Report CSV to list the funder name in a way that is consistent with the older Funder column. Intention is to simply this report further once we've transitioned to the new model
- Add some notes to the funder fields for onboarding and school configuration


app/controllers/admin/funders_controller.rb
```diff
@@ -7,6 +7,7 @@ def index
     end
 
     def show
+      @funding_summary = ::Commercial::Contract.current_contract_holder_summaries.first { |s| s[:id] = @funder.id }
```
tbhi: looks like should be `==`

tbhi (approved):

PR #5334, opened 2026-05-29
Remove kwh and co2 potential savings metrics

tbhi:


ldodds (approved):

db/migrate/20260529105327_add_more_potential_savings_metric_type_enum.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddMorePotentialSavingsMetricTypeEnum < ActiveRecord::Migration[8.1]
+  def up
+    %i[peak use heating_down heating_early heating_off insulate_pipes thermostatic_control solar_panels]
+      .each do |metric|
+      add_enum_value :impact_report_metric_types, metric
+    end
```
urbanwide, line 8: Wonder if you are going to remove the old metric enums?

urbanwide (approved):
Looks good. One minor question / comment

db/migrate/20260529105327_add_more_potential_savings_metric_type_enum.rb
tbhi: yes will look, guess we can drop old runs/metrics

urbanwide: Great, good idea. I meant remove the old metric_types from the enums in the database? Think your first migration added the new ones but didn't remove the old ones?

tbhi: I'll think all the old metrics will need to be migrated or removed before doing something like https://github.com/Energy-Sparks/energy-sparks/blob/00894985124a785a21d5fb1594da0d82efe48349/db/migrate/20260501135818_add_potential_savings_to_metric_type_enum.rb#L23 so looks like will need a follow PR to complete removal

urbanwide: Ok no problem, makes sense. We perhaps need a card in Trello so we don't forget

PR #5337, opened 2026-05-29
Add unit field to impact report metrics

tbhi:


app/models/impact_report/run.rb
```diff
@@ -74,14 +74,11 @@ def potential_savings
 
     def energy_efficiency
```
urbanwide: I prefer this way of sorting the energy efficiency metrics, rather than using the index in this case. If we are going to do this, we shouldn't store the energy efficiency in the metrics_index below though. I don't mind tidying this up later though (will probably move potential savings out of the index at some point too).

app/components/impact_reports/energy_efficiency/metrics_component/metrics_component.html.erb
```diff
@@ -7,21 +7,19 @@
         component_classes: 'theme theme-dark'
       ) do |grid| %>
     <% displayable.each do |metric| %>
-      <% grid.with_stats_card(if: metric.energy_efficiency_type == :annual_saving) do |card| %>
-        <%= card.with_header title: impact_t("energy_efficiency.metric_types.#{metric.metric_type}.title",
-                                             fuel_type: t("advice_pages.fuel_type.#{metric.fuel_type}")) %>
-        <%= card.with_figure format_unit(metric.value, metric.unit.to_sym, false) %>
+      <% grid.with_stats_card(if: metric.metric_category == 'energy_efficiency' &&
```
urbanwide: If you merge in master, there are more metrics integrated now, which might break this. So worth a merge in first and I will re-review then.

urbanwide (commented):
Couple of comments - will review again after master has been merged in again if that's ok. Some new metrics have been added which might break things. Thanks

app/components/impact_reports/energy_efficiency/metrics_component/metrics_component.html.erb
tbhi: updated with master

urbanwide: All looks good. As you have probably seen, there are more conflicts again but good to merge otherwise

urbanwide (approved):
Happy for this to be merged once the conflicts are sorted.

PR #5340, opened 2026-06-01
Add other issues to data source page

LukeGouldson:
Adds a 'Meter Issues' tab to data source show page, which displays all issues associated with meters for that data source


app/views/admin/data_sources/show.html.erb
```diff
@@ -43,21 +43,36 @@
...
+                                                                 .status_open.by_priority_order,
+                                             issuable_type: 'Data source' %>
+    <% else %>
+      <div class="bg-light p-2">No meter issues for <%= @data_source.name %>.</div>
+    <% end %>
+    <div class='pt-3'><%= render 'admin/issues/issues_links', issueable: @data_source %></div>
+  <% end %>
+<% end %>
+
+<script>
```
urbanwide: I wonder if this javascript could go in a different (shared) place. I know there is the same piece of code on the admin/school_groups/show.html.erb page...

urbanwide (changes requested):
Looks good - just wonder if the javascript at the bottom of the page can go in a better place? the same javascript features on another page too.

I am not sure if you have time to do this, but it might also be good to update the data sources admin interface to use Bootstrap 5 while you are there. Give me a shout if you need any hints on this.

urbanwide (approved):

PR #5342, opened 2026-06-01
Add impact report link and tidy

urbanwide:
- Adds link to report on config admin page.
- Also adds link to index of reports.
- Also tidies up a few other bits.

We probably also need a link in the group/settings left nav to the report and also in the manage group menu, but assume this is for another time.

### DIFF
--- a/app/models/commercial/contract.rb
+++ b/app/models/commercial/contract.rb
@@ -37,6 +37,10 @@
 #  fk_rails_...  (created_by_id => users.id)
 #  fk_rails_...  (product_id => commercial_products.id)
 #  fk_rails_...  (updated_by_id => users.id)
+#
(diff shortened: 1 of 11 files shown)


app/models/commercial/contract.rb
```diff
@@ -37,6 +37,10 @@
 #  fk_rails_...  (created_by_id => users.id)
 #  fk_rails_...  (product_id => commercial_products.id)
 #  fk_rails_...  (updated_by_id => users.id)
+#
```
ldodds: issue with annotate duplicating comments here, Tristan has a PR with a possible fix.

urbanwide:
One spec failing -  should be ok after this is merged: https://github.com/Energy-Sparks/energy-sparks/pull/5341

ldodds (approved):

PR #5359, opened 2026-06-04
Add review reminders to admin dashboard

LukeGouldson:
Adds two new admin dashboard prompts: overdue group reviews and monthly group reviews


app/components/admin/dashboard_prompt_component.rb
```diff
@@ -81,11 +98,23 @@ def prompt_for_missing_alert_contacts?
...
     end
 
     def weekly_issues_count
-      @weekly_issues_count ||= user.owned_issues.where(review_date: Date.current...(Date.current + 7)).count
+      @weekly_issues_count ||= user.owned_issues.exclude_issue_tag(group_review_tag)
+                                   .where(review_date: Date.current...(Date.current + 7)).count
+    end
+
+    def overdue_group_reviews_count
+      @overdue_group_reviews_count ||= user.owned_issues.for_issue_tag(group_review_tag)
```
ldodds, line 112: I've just realised we have a bug that affects all the issue related counts. An Issue has a status of open/closed. When counting issues to produce reminders we should be excluding those that have a closed status.

This will be most obvious for group reviews as at the moment it'll be warning about overdue reviews even if they've been marked as closed (done), but it affects all the prompts. 

Can you revise the scopes used on all the counts and update the specs?

ldodds (changes requested):

ldodds (approved):

PR #5362, opened 2026-06-04
rubocop safe autocorrect on some of /spec

tbhi:


ldodds (approved):
Hard to review this given its size. But if only safe corrections have been applied and tests are passing then think it should be OK.

PR #5365, opened 2026-06-08
Integrate benchmark metrics

urbanwide:
Integrated long_term, out_of_hours, heating_control, baseload benchmark metrics

Is "Annual [gas/electricity] use" mentioned in the spec document, the same as long_term @ldodds ?

There is no heating timing benchmark metric available as yet (it is mentioned in the document). Was this not included intentionally?

Also added a validation to metric, to ensure there is only one per metric_type, metric_category, fuel_type combination per run.


ldodds:
@urbanwide Yes long_term will be the benchmark for annual gas and electricity use, as shown on the long term usage page for each fuel type.

There is no heating timing benchmark as yet, the spec mentions we'd need to implement an entirely new benchmark if we wanted to report on this, so ignore for now.

ldodds (approved):

urbanwide:
> @urbanwide Yes long_term will be the benchmark for annual gas and electricity use, as shown on the long term usage page for each fuel type.

OK thanks @ldodds . Do you want me to change this text then, or is it ok for now? :

```
long_term:
  subtext: Schools achieving a well managed or exemplar rating for managing long term %{fuel_type} use
  title: Reducing long term %{fuel_type} use

```

> 
> There is no heating timing benchmark as yet, the spec mentions we'd need to implement an entirely new benchmark if we wanted to report on this, so ignore for now.

Ahh, not sure how I missed the highlighted text :-) Thanks

PR #5367, opened 2026-06-08
Add impact report links to group pages

urbanwide:
Displays the amount of schools from the impact report metrics rather than the latest visible schools count for group (so it matches report). Is this the right thing to do @ldodds?

Group dashboard:
image

Group advice index:
image


ldodds:
Yes, makes sense to use the same figures here.

ldodds (approved):

PR #5368, opened 2026-06-08
remove old metric types 

tbhi:


urbanwide (approved):
Looks good but please can total_savings also be removed from ENERGY_EFFICIENCY_METRICS in metric.rb?

```
    ENERGY_EFFICIENCY_METRICS = (
      %i[total_savings] + # not sure we need this now. Remove from DB too?
      ENERGY_EFFICIENCY_GENERATORS.flat_map { |type| type::METRICS }
    ).freeze
```

PR #5376, opened 2026-06-09
Fix ordering for metric reports created on the same day

urbanwide:


app/controllers/school_groups/impact_controller.rb
```diff
@@ -11,7 +11,6 @@ class ImpactController < ApplicationController
     before_action :redirect_unless_authorised
     before_action :redirect_unless_visible
     before_action :redirect_not_enough_data
-    before_action :enable_prototype_page
```
urbanwide, line 14: This is the 'prototype' tag from the top left of the page. Don't think we need this on the impact report page now.

ldodds (approved):

PR #5381, opened 2026-06-10
Add uniqueness validation to product and contract names

ldodds:
There's already a unique index but active record validation was missing.

### DIFF
--- a/app/models/commercial/product.rb
+++ b/app/models/commercial/product.rb
@@ -36,8 +36,7 @@ module Commercial
 
     self.table_name = 'commercial_products'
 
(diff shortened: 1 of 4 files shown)


app/models/commercial/product.rb
```diff
@@ -36,14 +36,16 @@ class Product < ApplicationRecord
 
     self.table_name = 'commercial_products'
 
-    validates_presence_of :name
+    validates :name, presence: true
```
tbhi: `validates :name, presence: true, uniqueness: true` would work?

tbhi (approved):

PR #5382, opened 2026-06-10
Allow meter attributes to be restored

LukeGouldson:
Adds a restore option to all deleted meter attribute across the site, removing their 'deleted_by'. Note that this currently does not change anything else about the attribute, such as updated_by etc - let me know if we want anything else about the attribute changing too


ldodds (approved):
Looks good, and nice tidy up of the tests 👍

PR #5383, opened 2026-06-10
Fix text on engagement featured school component

urbanwide:
In summary, this now does the following:
* Changes the left text on the engagement featured default school - it no longer shows zero activity or action counts. Also handles singular counts better.
* In response to Leigh's reply below, made the following podium changes:
  * Now links to current school shown on podium (if not already on that page) and makes it bold
  * Links schools to adult dashboard, unless on the pupil dashboard page.
  
Keeping text below so Leigh's reply makes sense:

---

What I currently have (what it defaulted to when changed to using the school scoreboard):
image

Becca asked for:
`Your highest scoring school is in X place among the x region and holds X place nationally.
`
Do we want it to be:
`Your highest scoring school is in 5th place on the South-West England scoreboard and holds 38th place nationally.
`

What is the preferred text @ldodds ? Thanks

A few other questions (could be done in a different card):
* The links at the bottom of the podium to the schools, always go to the pupil dashboard. Do we think this is still right? Or should the links be to be to the adult dashboards by default and then to pupil dashboards when on a pupil dashboard?
* The highest scoring school at the bottom does not have a link to the school. Think this would be useful to add in (certainly when not on the same school dashboard). Maybe could make the link bold to distinguish it from the other two.
* Do we want to also update the podium on the school group dashboard to have the same text?

### DIFF
--- a/app/components/impact_reports/engagement/featured_school_component.rb
+++ b/app/components/impact_reports/engagement/featured_school_component.rb
@@ -24,30 +24,36 @@ module ImpactReports
       end
 
       def default_description
-        impact_t('engagement.featured.description',
+        key = if activity_count.zero?
+                'actions'
+              else
+                action_count.zero? ? 'activities' : 'both'
+              end
+
+        impact_t("engagement.featured.description.#{key}",
                  school: school.name,
-                 activities: activity_count,
-                 actions: action_count)
+                 activities: impact_t('engagement.featured.activities', count: activity_count),
+                 actions: impact_t('engagement.featured.actions', count: action_count))
       end
 
-      def today
-        @today ||= Time.zone.today
+      def now
+        @now ||= Time.zone.now
       end
 
       def twelve_months_ago
-        @twelve_months_ago ||= today - 12.months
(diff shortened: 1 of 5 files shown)


ldodds:
* I think we should be using the same text and presentation on the group impact page as the group dashboard, so no extra variant
* The group dashboard version of the podium ranks the schools _within the group_ (they might not all be on the same regional scoreboard and group admins wanted to see the rankings across their group not the whole site). So I'm not sure the suggested text revision for the title makes sense. Especially as you might have 3 schools all at number 1 across different regions. The group dashboard version is currently using the group name and that's what I see on the impact page, not a regional scoreboard name (like in your screenshot above). The link in the title goes to the school group scoreboard so group admins can see rankings in the group. I suggest leaving the title on the podium as it currently is and address the other suggestions
* Makes sense to adjust the link target for different contexts. So on pupil dashboard go to other pupil dashboards, otherwise adult dashboards
* Adding a link for the featured school also makes sense. Making it bold seems helpful.

ldodds (approved):

PR #5385, opened 2026-06-10
Add component for producing a funding summary for a group

ldodds:
Adds a new component that is capable of producing a high-level summary of the funding situation in a group. This is used to provide a shorter summary on the group licensing summary page, like this:

Screenshot from 2026-06-10 14-23-53

The component supports a text based template for producing a similar version of the text which also includes the names of each school, e.g:

```
2 schools with fees due for the full academic year
Avebury Boys' Academy
Avebury Girls' Academy

2 schools self funding
Southpark Primary Academy
Castle Rock Academy
```

The latter is provided to allow the text to be copied into email correspondence with groups during renewals.

### DIFF
--- a/app/components/commercial/range_funding_summary_component/range_funding_summary_component.text.erb
+++ b/app/components/commercial/range_funding_summary_component/range_funding_summary_component.text.erb
@@ -1,0 +1,9 @@
+<% CATEGORIES.each do |category| %>
+  <% if categorised_schools[category]&.any? %>
+<%= category_description(category) %>
+<% categorised_schools[category].sort_by(&:name).each do |school| %>
+<%= school.name %>
+<% end %>
+
+  <% end %>
+<% end %>
--- a/app/components/commercial/range_funding_summary_component/range_funding_summary_component.text.erb
+++ b/app/components/commercial/range_funding_summary_component/range_funding_summary_component.text.erb
@@ -1,9 +1,8 @@
-<% CATEGORIES.each do |category| %>
-  <% if categorised_schools[category]&.any? %>
+<%- CATEGORIES.each do |category| %>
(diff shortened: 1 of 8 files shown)


app/components/commercial/range_funding_summary_component/range_funding_summary_component.text.erb
```diff
@@ -0,0 +1,9 @@
+<% CATEGORIES.each do |category| %>
+  <% if categorised_schools[category]&.any? %>
```
urbanwide: You could use `<%-` here to strip leading whitespace (and in a few other places) and then you could keep some of the indentation for the if/end and loops.
`<%- if categorised_schools[category]&.any? %>`

urbanwide (approved):
Looks good. One small comment

app/components/commercial/range_funding_summary_component/range_funding_summary_component.text.erb
ldodds: Not sure i knew about that!

app/components/commercial/range_funding_summary_component/range_funding_summary_component.text.erb
```diff
@@ -0,0 +1,8 @@
+<%- CATEGORIES.each do |category| %>
+  <%- if categorised_schools[category]&.any? %>
+<%= category_description(category) %>
+<%- categorised_schools[category].sort_by(&:name).each do |school| %>
+<%= school.name %>
+<% end %>
```
urbanwide, line 6: You could do it with the `<%- end %>` tags too - if you want everything to be indented.
Doesn't work with `<%=` though I don't think

PR #5388, opened 2026-06-11
load all devices in n3rgy response to deal with meter replacement

tbhi:


ldodds (approved):
Feels a bit fragile to assume that the second device in the response has readings that should override the first, as we don't know the ordering used by n3rgy. But I don't see any other information available to us to handle it differently.

Some examples I've seen have no overlaps. Another has 2 devices with exactly the same readings duplicated for an entire day, so summing the half-hours is not correct.

Think we go with this for now.

PR #5392, opened 2026-06-11
Impact reports: reorder energy efficiency metrics

urbanwide:
* Reorders energy efficiency metrics as discussed (I hope I have it right!)

Also:
* Adds unit of :gbp to potential savings metrics when they generated, to align with others.
* Removes footnotes from metric category enum, as is no longer needed
* Tidies up admin interface a bit
* Fixes an error that was happening if there was no latest run for a group

I think it might be useful / more efficient to add the visible_schools count directly to run model by the way, as it's being used on pages where we don't need all the metrics to check if there is enough data to display the report. Just a thought.


ldodds (approved):
Tested with a few groups and looks good. The admin interface tweaks also look good

urbanwide:
Thanks Leigh, did you see my last comment (in the description) about adding visible schools count to the run model? I'll merge now anyway.

ldodds:
> Thanks Leigh, did you see my last comment (in the description) about adding visible schools count to the run model? I'll merge now anyway.

I did, but didn't reply. Seems like a sensible refactor if you want to do that as a follow on.

PR #5393, opened 2026-06-12
migrate impact report metric out_of_hours to use units

tbhi:
missed from previous PR - not required if previous runs have been removed


ldodds:
I've deleted all previous runs.

PR #5394, opened 2026-06-12
Invoice creation

ldodds:
Implements process for generating invoices.

The process has two steps:

1. Raising an invoice. This involves choosing which schools will be included in the new invoice. This might be all of those with licences that are pending invoice, or a subset chosen by the user based on status of the school
2. Confirming the invoice creation, providing a purchase order number (which may vary between invoices)

Have added a new component to generate a simplified school status that is useful when raising an invoice, the status is one of:

- Not data visible
- No data (if not regenerated yet)
- No recent data if all fuel types are > 14 days out of date
- One utility, if at least one of electricity or gas is up to date
- All utilities, if both electricity and gas up to date


app/controllers/admin/commercial/contracts/invoices_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+
+          redirect_to admin_commercial_invoice_path(@invoice)
+        rescue ActiveRecord::RecordInvalid
+          redirect_to admin_commercial_contract_path(@contract), notice: 'Failed to create invoice'
+        end
+
+        private
+
+        def format_price_without_symbol(price)
+          format_price(price).delete('£')
```
tbhi: looks like format_price adds commas? does that cause problems? perhaps need separate round and format functions

app/views/admin/commercial/contracts/invoices/new.html.erb
```diff
@@ -0,0 +1,96 @@
...
+      </thead>
+      <tbody>
+          <%= f.simple_fields_for :line_items do |item_form| %>
+            <tr>
+              <td>
+                <%= item_form.input :licence_id, as: :hidden %>
+                <%= item_form.object.licence.school.name %>
+              </td>
+              <td class="col-1">
+                <%= item_form.input_field :number_of_meters,
```
tbhi, line 50: on firefox I see spinner controls with these 

Image

spec/system/admin/commercial/manage_invoices_spec.rb
```diff
@@ -98,4 +100,121 @@
...
+              licence.start_date.to_fs(:es_short),
+              licence.end_date.to_fs(:es_short),
+              format_price(contract.product.small_school_price),
+              'Edit'
+            ]
+          ]
+        end
+      end
+
+      context 'when then proceeding' do
```
tbhi: perhaps worth only have `:js` for this bit for speed

app/controllers/admin/commercial/contracts/invoices_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+
+module Admin
+  module Commercial
+    module Contracts
+      class InvoicesController < AdminController
+        include ApplicationHelper
+
+        load_and_authorize_resource :contract, class: 'Commercial::Contract'
+        def raise_invoice
+          @licences = @contract.licences.pending_invoice.by_start_date
```
tbhi: perhaps worth adding `.includes(school: :school_group)`

tbhi (approved):

app/views/admin/commercial/contracts/invoices/new.html.erb
ldodds: changed to be strings

app/controllers/admin/commercial/contracts/invoices_controller.rb
ldodds: Just switched to use standard round and format functions. At this point am just trying to ensure that what we store is what was previously displayed on-screen to avoid any later issues with rounding or formatting.

PR #5395, opened 2026-06-15
Don't convert bad codes from m3

tbhi:
- preserves the bad data value which seems useful for counting in Aggregation::ValidateAmrData.remove_dcc_bad_data_readings


ldodds (approved):
Leaving in place does mean the handling of codes for gas and electricity are consistent. 

Although if set to nil the validation logic would apply similar rules and either substitute the whole day or interpolate values. 

But lets leave it consistent now for known error codes.

PR #5399, opened 2026-06-16
Add new supplier model and forms

LukeGouldson:
Adds model and initial pages for suppliers


db/migrate/20260610150521_add_supplier_id_to_meter.rb
ldodds, line 1: Could be combined with the earlier migration.

app/views/admin/data_sources/show.html.erb
```diff
@@ -60,7 +60,7 @@
     <% if @data_source.active_meter_issues.issue.any? %>
       <%= render 'admin/issues/issues_list', issues: @data_source.active_meter_issues.issue
                                                                  .status_open.by_priority_order,
-                                             issuable_type: 'Data source' %>
+                                             issueable_type: 'Data source' %>
```
ldodds: Might be missing something, but I can't see where this var is used in the partial. If its not used then explains why the incorrectly named var didn't cause an error.

spec/factories/suppliers.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :supplier do
+    name { 'supplier' }
```
ldodds: Better to use something like `sequence(:name, 'Supplier AAAAA1')` then there's a unique name for each supplier when created.

spec/models/supplier_spec.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Supplier do
+  describe 'deletable' do
+    let!(:supplier) { create(:supplier, name: 'supplier') }
```
ldodds: Don't need the name here?

spec/system/admin/dashboards_spec.rb
```diff
@@ -137,6 +137,28 @@
           end
         end
 
+        describe 'my suppliers' do
+          let!(:user_supplier) { create(:supplier, name: 'user supplier', owned_by: user) }
```
ldodds: With unique names in the factory you could remove the `name` here and leave assertions the same.

app/views/schools/meters/_active_meters.html.erb
ldodds, line 1: The `_inactive_meters` partial will also need the new column.

ldodds (commented):
Some minor comments but main issue is updating the inactive meters list too.

The meters table is quite tight on space so adding supplier may cause some issues. There was a suggestion in Trello to rename a couple of the columns:

* change ‘meter system’ to ‘meter type’
* change ‘meter status’ to just ‘status’

Can you do that too to give a bit more space but we'll need to look at the table again when we load data in.

ldodds (approved):

PR #5409, opened 2026-06-16
Revise funder allocation report and user export

ldodds:
Revises the Funder Allocation report and main admin user exports to have separate columns for Current and Future Contract Holder.

The Current Contract Holder is the organisation who owns the contract associated with their current licence. This is summarised to be either a Funder name, or "MAT Funding" or "School self funding".

The Future Contract Holder is as above but looking at the first licence for the next academic year.


spec/models/user_spec.rb
```diff
@@ -448,6 +456,50 @@
...
     end
+
+    context 'when the school has a current and future licence' do
+      let!(:current_licence) do
+        create(:commercial_licence,
+               school:,
+               contract: create(:commercial_contract, contract_holder: create(:funder)))
+      end
+
+      let!(:next_academic_year) do
```
tbhi: looks not needed?

tbhi (approved):

PR #5410, opened 2026-06-16
Fix LGAP bug

ldodds:
The system has a rule that if there's a gap of >50 days data in the readings for a meter, then we truncate the gappy data as its too much to substitute.

This has been in the system since 2018 (or earlier). However there is also code that auto inserts a correction for gas meters to set missing weekend data to zero. 

There is an ordering problem with meter corrections being applied first, then the check for gaps. So gas meter data is not being truncated and we are attempting to substitute longer gaps.

Looks like bug was [originally noted and fixed here](https://github.com/Energy-Sparks/energy-sparks_analytics/commit/b9dcc60b7c63d2a18b8ee30b4faee996193ed126) but was then [reintroduced here](https://github.com/Energy-Sparks/energy-sparks_analytics/commit/1cccbac5958a4d25c6869d643966092eca5d3184#diff-b629e82abe37cc1a94f9222deeb82e2530ea0ee9a894f32f4e23484dccf83cbeL40-R45) and has been left in place since. 

This PR has the following functional changes:

- changes the ordering in which the validation steps are applied: we remove invalid data, check for gaps and then apply substitutions. The `do_validations` method has been broken up to make this order explicit and avoid similar bugs being introduced again
- removes the rule for automatically adding a substitution for missing weekend gas data. This seemed to have originally been added during to issues with early data feeds. It can still be configured manually if needed but shouldn't otherwise be done.

As test for the LGAP handling was added, confirming it works for both gas and electricity meters.

There are some other non-functional changes:

- reworked ValidateAmrData constructor to have default max number of days missing, and move constant to this class
- remove unused/unnecessary debug code from this and related classes
- change logging levels to `.debug` and use block style of logging (`.debug {...}`) throughout

The functional change broken an old unrelated test. I've replaced that with a version that doesn't rely on the checked in meter collection data.

I've identified all the gas meters in the system that will be affected by the changes and provided a report to the ops team.


spec/services/aggregation/validate_amr_data_spec.rb
```diff
@@ -245,4 +243,61 @@ def arbitrary_night_readings(meter)
...
+        (amr_data.start_date..amr_data.end_date).each do |date|
+          expect(amr_data.days_amr_data(date).type).to eq('ORIG')
+        end
+      end
+    end
+
+    context 'with a gap' do
+      let(:after_gap_date) { Date.new(2025, 12, 31) }
+      let(:amr_data) do
+        amr_data = build(:amr_data, :with_date_range, start_date: Date.new(2025, 1, 1),
```
tbhi: super() to reuse the original amr_data?

app/services/aggregation/validate_amr_data.rb
```diff
@@ -14,55 +14,43 @@ class UnexpectedGroupRule < StandardError; end
...
-    def validate(debug_analysis: false)
-      logger.debug '=' * 150
-      logger.debug "Validating meter data of type #{@meter.meter_type} #{@meter.name} #{@meter.id}"
-      logger.debug "Meter data from #{@meter.amr_data.start_date} to #{@meter.amr_data.end_date}"
-      logger.debug "DCC Meter #{@meter.dcc_meter}"
+      # Extracts the :meter_corrections from the meter attributes and stores them
+      # as meter_correction_rules, for later processing.
+      #
+      # The corrections are applied in a later step
+      process_meter_attributes
```
tbhi: preferable to not mutate meter in the constructor? perhaps not needed on the meter at all?

spec/services/aggregation/validate_amr_data_spec.rb
```diff
@@ -245,4 +243,61 @@ def arbitrary_night_readings(meter)
...
+  context 'when checking for long gaps' do
+    let(:amr_data) do
+      build(:amr_data, :with_date_range, start_date: Date.new(2025, 1, 1), end_date: Date.new(2025, 3, 31))
+    end
+
+    let(:meter) { build(:meter, type: :electricity, amr_data:, meter_collection: build(:meter_collection)) }
+
+    before { travel_to Date.new(2025, 12, 31) }
+
+    context 'with no gaps' do
```
tbhi, line 265: perhaps check either side of the 50 day limit

tbhi (approved):

app/services/aggregation/validate_amr_data.rb
ldodds: I've remove the method from meter and revised the tests that rely on that code. Can you take another quick look please?

tbhi (approved):

PR #5412, opened 2026-06-17
Impact reports: Add visible schools to run

urbanwide:
This adds visible_schools to the ImpactReport::Run model, and uses it where necessary to work out of there is enough data to show the report. This is used on dashboards (and other places), so seemed like a reasonable thing to do.

* Also now displays if there is enough data (i.e. visible_schools >= 2) on the admin index for reports and uses a grey background for any reports that currently wouldn't display (due to not being visible or having enough data).
* Also adds a column showing if any case studies are available (thought this would be useful for admins to get more of an overview of this).

image

While I was editing the admin interface, I added the extra edit button requested by Claudia:

image

Plus a bit of other minor refactoring


app/views/admin/impact_reports/index.html.erb
```diff
@@ -12,30 +15,51 @@
...
-      <th colspan="3">&nbsp;</th>
+      <th colspan="4">&nbsp;</th>
+      <th class="text-center bg-light border-start">Overview</th>
       <th colspan="2" class="text-center bg-light border-start border-end">Energy efficiency</th>
       <th colspan="2" class="text-center bg-light border-end">Engagement</th>
       <th></th>
     </tr>
     <tr>
       <th>Visible</th>
+      <th>Enough data</td>
```
tbhi: should be `</th>`

app/views/admin/impact_reports/index.html.erb
```diff
@@ -12,30 +15,51 @@
...
   <tbody>
     <% @school_groups.each do |school_group| %>
       <% config = school_group.impact_report_configuration %>
-      <tr class="<%= 'bg-light' unless config %>">
-        <td class="text-center"><%= checkmark(config&.visible) %></td>
+      <% run = school_group.latest_impact_report_run %>
+      <tr class="<%= 'text-muted bg-light' unless config&.visible && run&.enough_data? %>">
+        <td class="text-center">
+          <span data-bs-toggle="tooltip" title="<%= config&.visible ? 'Visible' : 'Not visible' %>">
+            <%= checkmark(config&.visible, off_class: 'text-muted') %></td>
```
tbhi: extra `</td>`

app/views/admin/impact_reports/index.html.erb
```diff
@@ -12,30 +15,51 @@
...
+        <td class="text-center">
+          <% title = run&.enough_data? ? "#{run&.visible_schools} visible schools in latest run" : 'Not enough data' %>
+          <span data-bs-toggle="tooltip" title="<%= title %>">
+            <% if run %>
+            <%= link_to checkmark(run&.enough_data?, off_class: 'text-muted'),
+                        school_group_impact_run_path(school_group, run) %>
+            <% else %>
+              <%= checkmark(run&.enough_data?, off_class: 'text-muted') %>
+            <% end %>
+            </td>
```
tbhi: looks unmatched

app/services/school_groups/impact_report/generator.rb
```diff
@@ -11,10 +11,12 @@ def self.metric_names = GENERATORS.flat_map { |generator| generator.metric_names
 
       def initialize(school_group)
         @school_group = school_group
+        @visible_schools = Base.new(@school_group).visible_schools.count
```
tbhi: wonder if it would make sense to calculate visible schools here and pass it into the generator

tbhi (approved):

app/services/school_groups/impact_report/generator.rb
urbanwide: Thanks. Good idea. Though would you mind checking I have implemented as intended? Thanks

app/services/school_groups/impact_report/generator/base.rb
```diff
@@ -28,14 +29,14 @@ def metrics
           end
         end
 
+        attr_reader :visible_schools
```
tbhi: can still be private?

app/services/school_groups/impact_report/generator/base.rb
```diff
@@ -15,8 +15,9 @@ def self.metric_names
 
         private_class_method def self.metric_category = self::METRIC_CATEGORY
 
-        def initialize(school_group)
+        def initialize(school_group, visible_schools = nil)
           @school_group = school_group
+          @visible_schools = visible_schools || @school_group.assigned_schools.visible
```
tbhi: perhaps simpler without the default

PR #5414, opened 2026-06-18
Data source and supplier csv import

LukeGouldson:
Adds a rake task which imports changes to meter data via csv, and an after party task to populate the suppliers table and update data sources


LukeGouldson:
Currently, the after party task does not delete any of the existing data source data, because this would cause us to lose a lot of existing information (eg related issues, and alert thresholds). Let me know how we want to go about managing this

lib/tasks/deployment/20260616112840_populate_suppliers_and_data_sources.rake
```diff
@@ -0,0 +1,112 @@
...
+    # Create a supplier model for above suppliers
+
+    suppliers.each do |s|
+      Supplier.create(name: s)
+    end
+
+    # Create a data source model for above data sources
+
+    data_sources.each do |d|
+      existing_data_source = DataSource.find_by(name: d)
```
ldodds: Could use find_or_create_by! https://guides.rubyonrails.org/v7.0.2/active_record_querying.html#find-or-create-by-bang

lib/tasks/meters/import_suppliers_data_sources_and_statuses.rake
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+namespace :meters do
+  desc 'Import suppliers and data sources from a CSV file.'
+  task :import_suppliers_and_data_sources, [:csv_file] => :environment do |_t, args|
+    puts "#{DateTime.now.utc} Loading suppliers and data sources from #{args.csv_file}"
+
+    importer = Meters::DataSourceAndSuppliersImporter.new
+    CSV.foreach(args.csv_file, headers: true) do |row|
+      meter = importer.import({
```
ldodds, line 10: Are we also updating the meter status? There's a comment in Trello about having an "Updated Meter Status" column too?

ldodds (approved):

lib/tasks/meters/import_suppliers_data_sources_and_statuses.rake
LukeGouldson: I believe this comment was left on a previous version of the file - the meter status is also updated here

ldodds (approved):

PR #5415, opened 2026-06-18
Add caching to impact reports

urbanwide:


ldodds (approved):
Could probably cache for much longer than an hour given that it'll invalidate if the config is changed.

urbanwide:
> Could probably cache for much longer than an hour given that it'll invalidate if the config is changed.

There are elements of the page - i.e. the Engagement::FeaturedSchool component and testimonials that use 'live' data. How about caching for a day? Or just not cache these sections?

ldodds:
> > Could probably cache for much longer than an hour given that it'll invalidate if the config is changed.
> 
> There are elements of the page - i.e. the Engagement::FeaturedSchool component and testimonials that use 'live' data. How about caching for a day? Or just not cache these sections?

Oh I see. Well these aren't changing that frequently, you could up to 3 or 6 hours, and I think it'd be fine.

PR #5416, opened 2026-06-18
include modelled solar attribute in aggregation

tbhi:


analytics/lib/dashboard/aggregation/accounting_costs.rb
ldodds, line 1: Looks fine, but unrelated to the solar changes?

ldodds (approved):
The test for the aggregate service changes could cover a few more cases. At the moment it's just checking that a generation meter is created from the modelled solar, but could also check, e.g. that it correctly combines the metered and modelled data. Or that if the modelled solar pv generation meter starts before the start date, then the overall aggregate range is based on the modelled data. Looks like its all working from manual testing but might be good to have some more tests around that.

analytics/lib/dashboard/aggregation/accounting_costs.rb
tbhi: moved to https://github.com/Energy-Sparks/energy-sparks/pull/5441

ldodds:
Refactoring looks good

PR #5417, opened 2026-06-22
Supplier pages improvements

LukeGouldson:
Adds columns to the supplier page, add suppliers to the lagging data source email, and adds suppliers to multiple reports


app/views/admin/suppliers/_email_report_button.html.erb
ldodds, line 1: In my PR [here](https://github.com/Energy-Sparks/energy-sparks/pull/5418) I've added a component to handle this, but also simplified how the buttons operate as there's too many clicks. Can tidy up after merging both.

ldodds (approved):

PR #5418, opened 2026-06-22
Change user export to emailable job

ldodds:
Change admin user download to run as a background job and send an email.

Introduces a simple component for generating these types of buttons. Button is then reused on this page and a few others. 

Ended up simplifying the buttons on data source and procurement route pages as they used a modal, but with no extra content and then button had an "are you sure?" prompt. Adds several unnecessary clicks.


app/mailers/admin_mailer.rb
```diff
@@ -69,6 +69,12 @@ def funder_allocation_report
     mail(to: to, subject: admin_subject(title))
   end
 
+  def user_export_report
+    to, csv = params.values_at(:to, :csv)
+    attachments['users.csv'] = { mime_type: 'text/csv', content: csv }
+    mail(to:, subject: "User export #{Time.zone.now.to_fs(:es_compact)}")
```
tbhi: could use admin_subject here

spec/system/admin/users_spec.rb
```diff
@@ -66,6 +66,23 @@
     end
   end
 
+  describe 'when exporting users' do
+    subject(:email) { ActionMailer::Base.deliveries.last }
```
tbhi: could use `last_email` from EmailHelpers here

app/mailers/admin_mailer.rb
```diff
@@ -69,6 +69,12 @@ def funder_allocation_report
     mail(to: to, subject: admin_subject(title))
   end
 
+  def user_export_report
```
tbhi, line 72: useful to have a preview for this

tbhi (approved):
also seem

app/mailers/admin_mailer.rb
ldodds: Want to get this released so will add a preview separately

PR #5428, opened 2026-06-23
Issue report mailer fix

LukeGouldson:
Pluralizes issue report mailer in send user report as per [error](https://app.rollbar.com/a/energysparks/fix/item/EnergySparksProduction/5151#detail)


tbhi (commented):
I think worth a small spec like `spec/tasks/meters/check_for_dcc_spec.rb`  although I would just check for a sent mail rather than mocking - has https://github.com/Energy-Sparks/energy-sparks/blob/d0bbd6d52682b88ad156ed891634942fc13ce439/spec/mailers/issues_report_mailer_spec.rb#L10 already so wouldn't check much other than say email subject

tbhi (approved):

PR #5429, opened 2026-06-23
Add admin link to impact reports config interface

urbanwide:
In the call we talked about putting it in /admin/reports but it's more of an admin page in my head? Definitely something with school group context, but not sure it goes with the other items in there. What do you think @ldodds? 

image

Alternative in the reports index, not sure it's right where it is here either.
image

### DIFF
--- a/app/views/admin/index.html.erb
+++ b/app/views/admin/index.html.erb
@@ -23,6 +23,7 @@
       <% SchoolGroup::RESTRICTED_GROUP_TYPES.sort.each do |group_type| %>
         <li><%= link_to "Manage #{group_type.to_s.humanize} Groups", admin_school_groups_path(group_type:) %></li>
       <% end %>
+      <li><%= link_to 'Manage Impact Reports', admin_impact_reports_path %></li>
     </ul>
 
     <h2>Schools and and scoreboards</h2>


ldodds:
Think I prefer it under groups on the main admin page, like your first image.

Eventually I'd like to get away from having a "Reports" page as I the distinction is a bit arbitrary sometimes.

We should also think about having a "My Impact Reports" section on the admin dashboard to provide a filtered view for admin users.

urbanwide:
> Eventually I'd like to get away from having a "Reports" page as I the distinction is a bit arbitrary sometimes.

I agree, think this is a good idea.

> We should also think about having a "My Impact Reports" section on the admin dashboard to provide a filtered view for admin users. 

I can add a card for this?

ldodds:
> > Eventually I'd like to get away from having a "Reports" page as I the distinction is a bit arbitrary sometimes.
> 
> I agree, think this is a good idea.
> 
> > We should also think about having a "My Impact Reports" section on the admin dashboard to provide a filtered view for admin users.
> 
> I can add a card for this?

Please add a card, but perhaps Luke can pick this up.

ldodds (approved):

PR #5431, opened 2026-06-23
Remove trailing spaces in reading dates

ldodds:
One of the data feeds has trailing whitespaces in the reading dates. The date_format for the config includes a space and they parse fine when loading data for validation.

But the meter loading report and unvalidated readings reports use SQL to parse the dates and the code isn't handling the trailing space correctly.

I've done several things here:

- trimmed whitespace from reading dates when parsed from CSV
- adjusted the custom SQL date parsing to handle the trailing space
- added tests that confirm this works

I've also manually checked that the trimmed dates still parse correctly with a date format that has a trailing space. So should be no impact on the load and ensures that both old and new data is handled correctly.


app/models/amr_data_feed_reading.rb
```diff
@@ -68,6 +68,7 @@ class AmrDataFeedReading < ApplicationRecord
     WHEN date_format='%d/%m/%Y' THEN to_date(reading_date, 'DD/MM/YYYY')
     WHEN date_format='%d/%m/%y' AND reading_date~'\\d{4}-\\d{2}-\\d{2}' THEN to_date(reading_date, 'YYYY-MM-DD')
     WHEN date_format='%d/%m/%y' THEN to_date(reading_date, 'DD/MM/YY')
+    WHEN date_format='%Y-%m-%d ' AND reading_date~'\\d{4}-\\d{2}-\\d{2}\s*$' THEN to_date(reading_date, 'YYYY-MM-DD')
     WHEN date_format='%Y-%m-%d' AND reading_date~'\\d{2}/\\d{2}/\\d{4}' THEN to_date(reading_date, 'DD/MM/YYYY')
     WHEN date_format='%Y-%m-%d' THEN to_date(reading_date, 'YYYY-MM-DD')
     WHEN date_format='%Y-%m-%d' THEN to_date(reading_date, 'YYYY-MM-DD ')
```
tbhi: remove this line while here? doesn't appear to do anything given condition is the same as previous line

tbhi (approved):
seems like it would be better to have no space in date_format though - to_date seems to ignore whitespace

PR #5432, opened 2026-06-24
Allow solar installations to be disabled

tbhi:


db/migrate/20260623091133_add_active_to_solar_installations.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddActiveToSolarInstallations < ActiveRecord::Migration[8.1]
+  def change
```
ldodds, line 4: Why not make everything active by default? Matches the current system behaviour and avoids needing an after party tasks too.

tbhi: changed default to true

ldodds (approved):

PR #5434, opened 2026-06-24
Fix impact report generator issue

urbanwide:
On test, was getting this error when generating impact reports:

https://app.rollbar.com/a/energysparks/fix/item/EnergySparksTestEnvironment/4978/occurrence/463277378379#detail

The rake task generates metrics for school groups with active schools - this occurred when there were active schools, but no visible schools for a school group. The benchmark generator passes all visible schools (empty in this case) to CategoriseSchools, which builds a sql query interpolating school ids directly into an 'in' clause. As the list of ids is empty, invalid sql was being generated: 'WHERE schools.id IN ()'.

Alternative would be just to generate impact reports for groups with visible schools but thought best not change this for now.


ldodds (approved):
Looks good. Agree that it'd be best to just generate reports for groups with visible schools, but so long as regenerate bug fixed, it doesn't seem urgent.

PR #5435, opened 2026-06-24
Add more flexibility when editing and renewing contracts

ldodds:
Further changes to the contract renewal workflow.

Invoice terms (full, pro_rata) can now be edited on standard contracts until they've been invoiced. The invoice terms can also be altered during renewal, if needed, but the defaults remain as 'pro_rata' for this contract type.

Allows changing from a custom to a standard contract, or vice versa when renewing a contract. This is handled by adding a new interstitial page so the user can confirm the choice of what contract type to create for the renewal. This is primarily there to allow for custom contracts to be changed to a more standard pro rata based contract.

### DIFF
--- a/app/controllers/admin/commercial/contracts_controller.rb
+++ b/app/controllers/admin/commercial/contracts_controller.rb
@@ -62,10 +62,19 @@ module Admin
         }
       end
 
+      def renew
+        @original = ::Commercial::Contract.find(params.expect(:original_contract_id))
+        @chosen_params = {
+          contract_holder_id: params[:contract_holder_id],
+          contract_holder_type: params[:contract_holder_type],
+          original_contract_id: params[:original_contract_id]
+        }
+      end
+
       def new
         if renewal_request?
           @original = ::Commercial::Contract.find(params.expect(:original_contract_id))
-          @contract = ::Commercial::Contract.as_renewal(@original)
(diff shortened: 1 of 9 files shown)


app/controllers/admin/commercial/contracts_controller.rb
```diff
@@ -62,10 +62,19 @@ def choose
...
+          contract_holder_type: params[:contract_holder_type],
+          original_contract_id: params[:original_contract_id]
+        }
+      end
+
       def new
         if renewal_request?
           @original = ::Commercial::Contract.find(params.expect(:original_contract_id))
-          @contract = ::Commercial::Contract.as_renewal(@original)
+          @contract = ::Commercial::Contract.as_renewal(@original, chosen_type: params.expect(:chosen_type)&.to_sym)
```
tbhi: `&.` not needed if using expect I think

app/models/commercial/contract.rb
```diff
@@ -200,22 +200,32 @@ def self.current_contract_holder_summaries
 
     def self.temporal_group_keys = %i[contract_holder_id contract_holder_type]
 
-    def self.as_renewal(original)
+    def self.as_renewal(original, chosen_type: :contract)
```
tbhi: is the default for chosen_type needed? might be clearer without

tbhi (approved):

app/models/commercial/contract.rb
ldodds: Have simplified.

PR #5436, opened 2026-06-24
school search feature

LukeGouldson:
Adds a find school by name feature, similar to the find school by MPXN

### DIFF
--- a/app/controllers/admin/find_school_by_name_controller.rb
+++ b/app/controllers/admin/find_school_by_name_controller.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module Admin
+  class FindSchoolByNameController < AdminController
+    def index
+      @schools = find_schools_by_name
+      @onboardings = find_onboarding_by_name
+    end
+
+    private
+
+    def find_schools_by_name
+      if params['query'].present?
+        School.search_by_name(params['query']).limit(20)
+      else
+        []
+      end
+    end
+
+    def find_onboarding_by_name
+      if params['query'].present?
+        SchoolOnboarding.incomplete.search_by_school_name(params['query']).limit(20)
+      else
+        []
+      end
+    end
+  end
+end
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -781,6 +781,8 @@ class School < ApplicationRecord
 
     return { title: 'Data Visible', classes: 'ms-1 badge rounded-pill text-bg-success' } if data_visible?
 
+    return { title: 'Visible', classes: 'ms-1 badge rounded-pill text-bg-success' } if visible?
+
     { title: 'Awaiting Activation', classes: 'ms-1 badge rounded-pill text-bg-danger' }
   end
 
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -131,7 +131,10 @@ class School < ApplicationRecord
   include Commercial::LicenceHolder
   include PgSearch::Model
 
-  pg_search_scope :search_by_name, against: :name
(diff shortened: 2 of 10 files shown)


app/views/admin/find_school_by_name/index.html.erb
```diff
@@ -0,0 +1,43 @@
...
+A maximum of 20 schools and 20 onboardings will be returned.
+</p>
+
+<%= simple_form_for admin_schools_search_find_school_by_name_index_path, method: :get, html: { class: 'form' } do |f| %>
+  <%= f.input :school_name, label: 'School name', autofocus: true, input_html: { name: :query, value: params[:query] } %>
+  <%= f.submit :Search, class: 'btn btn-primary' %>
+<% end %>
+
+<% if @schools.any? || @onboardings.any? %>
+  <table class="table">
```
ldodds: Add table sorted

app/views/admin/find_school_by_name/index.html.erb
```diff
@@ -0,0 +1,43 @@
...
+    <thead>
+      <th>School</th>
+      <th>Model</th>
```
ldodds: Maybe "Type"?

app/controllers/admin/find_school_by_name_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Admin
+  class FindSchoolByNameController < AdminController
+    def index
```
ldodds, line 5: This works, but I think what I was originally expecting was to see if we could use pg_search features to do a search across both models. Looking at the multi_search docs more closely there's a bit more setup to it, so it might not be worth it just for this use case.

app/models/school.rb
```diff
@@ -129,6 +129,9 @@ class School < ApplicationRecord
   include AlphabeticalScopes
   include Commercial::ContractHolder
   include Commercial::LicenceHolder
+  include PgSearch::Model
+
+  pg_search_scope :search_by_name, against: :name
```
ldodds: Can we tailor both search scopes a bit more? E.g. a search for "Abbot" doesn't return results for "Abbots", so think we need to specify a dictionary: https://github.com/Casecommons/pg_search#dictionary and maybe enable prefix search: https://github.com/Casecommons/pg_search#prefix-postgresql-84-and-newer-only

app/models/school.rb
```diff
@@ -768,6 +771,22 @@ def community_use_times_to_analytics
     school_times.community_use.map(&:to_analytics)
   end
 
+  def status_display
```
ldodds, line 777: Think we can improve on this a bit, could you:

- Add some styling, so these are pills like we have in _school_status_buttons, I think we have an indicator somewhere for archived. Would like buttons/text to be consistent across the site
- Just display single status. If school is not data visible, just show if they are visible or awaiting activation

Can we also add some statuses for onboarding, e.g. "Not started" if no events, "In progress" if its been started but not yet finished

app/controllers/admin/find_school_by_name_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+      @schools = find_schools_by_name
```
ldodds, line 6: There's a bug here. If a user starts the onboarding process and completes the school form, then the school is added to the system, but the onboarding is not yet complete.

If I take an onboarding to that stage, I get duplicate results, one for the school search and one for the onboarding. 

In this case the school should show up in the onboarding list, not as a proper school yet. It's not actually awaiting activiation. So probably need a scope on the school query.

Image

ldodds (approved):

PR #5437, opened 2026-06-26
Make heating model cache functional

ldodds:
When analysing gas data we use a heating model. A heating model is produced using a linear regression over the aggregated gas data for the school. In general we end up using the best fitting of two (sometimes more?) different models. So the process involves generating multiple fits to the data, then using the best one.

The analytics has long has some code (`AnalyseHeatingAndHotWater::ModelCache`) that was intended to cache the results of generating models, to allow them to be reused. E.g. when generating the gas alerts for a school we run multiple analyses that all run the same model. Caching avoids re-fitting the model each time.

But while looking at some performance issues yesterday I noticed that regenerating schools were repeatedly refitting the models. Investigation showed that there's a nested series of bugs that are causing the intended caching to work.

To summarise:

-  `AnalyseHeatingAndHotWater::ModelCache` uses a hash to store the models, with a compound key that reflects the configuration of the model being fitted. We should only reuse a model if it was generated with same configuration (e.g. type, date range, etc). The model cache used a hash as the hash key, but one of the members of the key (`SchoolDatePeriod`) does not implement `==`, `eql?` or `hash` so any cached model would never be retrieved
- The `SchoolDatePeriods` used to configure the regression were using inconsistent types, so there were fewer cache hits even when the date ranges and other configuration were consistent
- `AlertAnalysisBase#initialize` had a call to clear the model cache, so the cache was reset every time a new alert was run. So it was effectively disabled.
- The `Schools::SchoolRegenerationService` populates the rails cache with the aggregated meter collection before the alerts are run. Prior to that point no heating model would be generated, so any benefits of the cache during the rest of the regeneration process would not be available to the application. The advice pages and charts would therefore still run regressions

This PR addresses all of the above by:

- Implementing the necessary equality methods on `SchoolDatePeriod`
- Removing the call to clear the cache in the alert base class
- Aligning the `:type` used when creating a `SchoolDatePeriod` so that we get more cache hits
- Adding a second call to cache the meter collection in the regeneration process

There are quite a few RSpec tidy ups here, but have tried to comment on the main changes. Also removed some unused code, including a separate model cache that was only used in the old analysis testing framework.

Tried to do the minimal changes to make the cache functional and maybe help to reduce CPU load/bottlenecks during regeneration.. The heating model code needs a proper refactor.


analytics/lib/dashboard/advice/heating_model_fitting_advice.rb
```diff
@@ -125,12 +128,8 @@ def advice_valid?
       end
     end
 
-    def heating_model(model_type)
```
ldodds, line 128: This is the 'functional' change here. Can't find this being used anywhere and it calls `.model_cache` on the meter collection (`@school`) here. But that is never populated. So looks safe to remove.

This old advice is only used on the "Expert analysis" page which is only used internally.

analytics/lib/dashboard/alerts/common/alert_analysis_base.rb
```diff
@@ -32,35 +23,31 @@ def initialize(school, _report_type)
...
-      log_stack_trace(e)
-      @calculation_worked = false
-    rescue StandardError => e
-      log_stack_trace(e)
-      Rollbar.warning(e, alert_type: self.class.name, asof_date: asof_date, school: @school.name) if ENV["LOG_ALL_ANALYTICS_ERRORS"] == 'true'
-      @calculation_worked = false
     end
-  end
-
-  def self.test_mode
```
ldodds, line 61: The actual functional change

analytics/lib/dashboard/alerts/gas/alert_model_cache_mixin.rb
```diff
@@ -1,15 +1,6 @@
 module AlertModelCacheMixin
-  # during analytics testing store model results to save recalculating for different alerts at same school
-  private def model_cache(meter, asof_date)
```
ldodds, line 3: Remove legacy class based cache, which is now unused.

analytics/lib/dashboard/modelling/heating/heating_regression_models.rb
ldodds, line 1: Majority of these changes are just rubocop tidying. I have removed one redundant method and also changed all .info debug messages to be of the form `logger.debug { ... }`. We don't need this level of output in production and use of the block avoids creating the strings if not needed.

lib/school_date_period.rb
```diff
@@ -1,20 +1,32 @@
...
+  end
+
+  # Implement equality such that a period can be used in a composite hash key
+  # Title is treated as a comment, so only checking type and dates
+  def ==(other)
+    other.is_a?(SchoolDatePeriod) && type == other.type && start_date == other.start_date && end_date == other.end_date
+  end
+  alias eql? ==
+
+  def hash
```
ldodds, line 28: Implementing `==`, `eql?` and `hash` is to allow SchoolDatePeriod to be used as part of a composite hash key when storing heating models.

analytics/app/models/dashboard/meter.rb
```diff
@@ -246,7 +246,7 @@ def heating_only?
 
     def up_to_one_year_model_period
       start_date = [amr_data.end_date - 364, amr_data.start_date].max
-      SchoolDatePeriod.new(:up_to_1_year_meter, 'Current Year', start_date, amr_data.end_date)
+      SchoolDatePeriod.new(:analysis, 'Current Year', start_date, amr_data.end_date)
```
ldodds, line 249: Through this PR, focusing on any SchoolDatePeriods that are used to create heating models, I've adjust the period type to be `:analysis`. A range of different values were being used, but these need to align for the heating model caching to work. The types aren't very meaningful anyway, so thought this was fine.

analytics/app/services/heating/seasonal_control_analysis_service.rb
```diff
@@ -127,7 +127,7 @@ def amr_end_date
 
     def calculate_heating_model
       # The period could be alerted here to parameterise the analysis based on date
-      period_start_date = [amr_end_date - 365, amr_start_date].max
+      period_start_date = [amr_end_date - 364, amr_start_date].max
```
ldodds, line 130: Was inconsistent with alerts. Aligned so the cached model will be reused.

app/services/schools/school_regeneration_service.rb
```diff
@@ -14,6 +14,7 @@ def perform
       if continue
         cache_meter_collection(meter_collection)
         regenerate_school_metrics(meter_collection)
+        cache_meter_collection(meter_collection)
```
ldodds, line 17: The cached heating models in the aggregate gas meter will only be available for later use in the application, e.g. when displaying advice pages or running charts, if the meter collection is updated to capture any cached/memoized objects.

Caching after regeneration is necessary if we want to gain some performance from avoiding having to re-run the regressions when displaying a gas advice page (pre-caching) or a gas chart. 

I've left the cache before the regenerate in place as otherwise if a user visits the school dashboard or advice page, the web application will serve a holding page whilst is aggregates the school and populates the cache on demand. Caching twice has some overheads but avoids potentially multiple requests trying to populate the cache on demand whilst the alerts for the school are still running.

ldodds:
@tbhi would be interested in an early review of this

analytics/lib/dashboard/alerts/gas/alert_model_cache_mixin.rb
```diff
@@ -1,15 +1,6 @@
...
-    return call_model(asof_date) unless AlertAnalysisBase.test_mode
-
-    @@model_cache_results ||= {}
-
-    return @@model_cache_results[meter.object_id] if @@model_cache_results.key?(meter.object_id)
-
-    # limit cache size
-    @@model_cache_results.delete(@@model_cache_results.keys.first) if @@model_cache_results.length > 20
-    @@model_cache_results[meter.object_id] = call_model(asof_date)
+  private def model_cache(_meter, asof_date)
```
tbhi: seems like also worth removing the `_meter` argument

tbhi (approved):

PR #5438, opened 2026-06-26
Bulk load meter data

ldodds:
Currently the unvalidated and validated meter factories run one query per meter to load in the required readings.

PR changes the code to load all meter data in a single query, then construct the meters from the pre-loaded readings. Reduces the number of SQL queries for schools with lots of meters.

Also adds `.includes(:meter_attributes)` to the scopes that load the meters to avoid N+1 queries when fetching attributes. Some meters may have a lot of corrections.


ldodds:
Deployed to test

tbhi (approved):

PR #5440, opened 2026-06-26
Bulk insert validated data

ldodds:
Try reducing the number of inserts required to save validated meters. Initial draft includes modified upserted that can handle multiple meters, and changes to do a single upsert per fuel type, rather than one upsert per meter.

May not be much of a trade-off of multiple inserts vs larger inserts and postgres having to do more conflict detection. Need some testing.


ldodds:
deployed to test

PR #5443, opened 2026-06-29
Update activity page to reflect latest design

urbanwide:
New version with controversial 9:3 split in the header. Have tried to get this to look right with 6:6 split, but the page looks wonky, the image just doesn't always look great and the tags sometimes wrap to two lines. 9:3 version below:

Screenshot 2026-07-06 at 12 17 15

Here is an example of the 6:6 version - hard to get the image right too, though it's ok with this one:

image

For reference, here are all the possible iterations of the "record an activity"  or "login to record" prompts:

Screenshot 2026-07-06 at 12 16 25

See what you think about the dark blue backgrounds on the prompts too @ldodds


ldodds:
Based on the pictures the 9:3 split looks much better. I think the dark blue works on the prompts.

Wonder if "Sign in to record this activity" should just be "Record this activity" as the wording indicates you need to sign in.

urbanwide:
Updated prompt button for login prompt. Also shows the prompt for if activity has been recorded previously (as well as other prompts):
image

ldodds (approved):
I've just reviewed the UX here, as the follow on PR will refactor further. Will take a closer look at the code then.

I've noticed that a few of the images are cropping differently (e.g. `/activity_types/178`) but think that's probably acceptable and means we probably just need better image. The layout looks a lot better.

PR #5444, opened 2026-06-29
Supplier import changes

LukeGouldson:
the 'updated meter status' column is now allowed to be blank, and the meter will still update - so long as the existing status of the meter is Manual Request


ldodds (approved):
Probably safest option to do what you have here, i might have allowed it to just be blank regardless but the extra sanity check does help.

LukeGouldson:
Great, I'll send stacie an email now

PR #5447, opened 2026-06-29
Add impact report to dashboard

LukeGouldson:
Adds filtered impact report to admin dashboard


app/views/admin/impact_reports/index.html.erb
```diff
@@ -1,20 +1,27 @@
-<% content_for :page_title, 'Impact Reports' %>
-<h1>Impact Reports</h1>
+<% if @dashboard_user %>
+  <% content_for :title do %>
+        <%= "#{@dashboard_user.display_name}'s Impact Reports" %>
+    <% end %>
+    <%= link_to 'View all impact reports', admin_impact_reports_path, class: 'btn btn-secondary btn-sm' %>
```
urbanwide: Noticed a bit of inconsistency with buttons sizes at the top of each dashboard sub-page - some are btn-sm and some are just normal sized. e.g. these are small:

Image

These are normal sized (also not quite lined up on my screen):
Image

Just might be better if they are all the same. Thanks

spec/system/admin/dashboards_spec.rb
```diff
@@ -115,6 +115,31 @@
...
+          end
+
+          let!(:non_user_school_group) { create(:school_group, :with_active_schools, count: 2) }
+
+          before do
+            click_on 'My Impact Reports'
+          end
+
+          it 'links to the impact reports page' do
+            expect(page).to have_current_path("/admin/dashboards/#{user.id}/impact_reports")
```
urbanwide, line 136: Think these two should be in separate tests (try and keep one expect statement per test).

urbanwide (commented):
Looks good with a couple of minor comments

urbanwide (approved):

PR #5449, opened 2026-06-30
Fix blank columns in suppliers import

LukeGouldson:
Allows for blank fields in data source suppliers and meter status - and will still update the meter with the other fields if 1 of its fields is left blank

### DIFF
--- a/app/services/meters/supplier_data_source_status_importer.rb
+++ b/app/services/meters/supplier_data_source_status_importer.rb
@@ -4,22 +4,62 @@ require 'csv'
 
 module Meters
   class SupplierDataSourceStatusImporter
-    def import(data) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity
+    def import(data)
       meter = Meter.find_by(mpan_mprn: data[:meter])
       return if meter.blank?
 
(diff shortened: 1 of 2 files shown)


ldodds (changes requested):
I don't think this quite does it. If the new Data Source is set to blank, we should update the meter to remove any existing association. I suggest just setting association to nil if the value is nil/empty string, otherwise try to find the Data Source, Meter Status, etc.

Also simplify the meter status code to just set it to blank regardless of the original status.

I'd also like to see a single database update with all changes, otherwise the import is going to take a long time. So just build a hash of the attributes being changed and do a single update at the end.

Finally, if there is a DataSource, Supplier or AdminMeterStatus to set and then we can't find it, then I need that logged to stdout out (`puts`) along with the `mpan_mprn` so I can identify any incorrect/missing data during the import. The current revisions mean I won't see any issues during the dry-run. If the other object can't be ffound, then don't include that attribute in the update. I think the Rake task is still logging if a Meter can't be found.

ldodds (approved):
This will do the job. Could be refactored to be a bit more concise, but fine for a temporary tool. If you merge I'll release to test once I've updated the database.

PR #5458, opened 2026-06-30
Add xero import download for single invoice

ldodds:
To simplify generating invoices in Xero this PR provides an option for downloading a CSV with the information for a single invoice. The CSV confirms to a Xero standard template.

The descriptions still need confirming, so will be changed following review.

The PR also includes a new XeroAccountCode model which is used to associate an invoice with a specific code in our account. New codes can be added and associated with individual contracts. This information is included in the Xero download to ensure that line items are associated with the correct accounting code.


app/services/commercial/xero_invoice_exporter.rb
```diff
@@ -0,0 +1,97 @@
...
+        description,
+        1,
+        rounded_price(price),
+        invoice.contract.xero_account_code&.code,
+        TAX_TYPE
+      ]
+    end
+
+    def rounded_price(price)
+      format('%.2f', price.round(2))
```
urbanwide: Think you can just do the following, as it already does rounding:
`format('%.2f', price)`

app/services/commercial/xero_invoice_exporter.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+module Commercial
+  class XeroInvoiceExporter
+    TAX_TYPE = '20% (VAT on Income)'
```
urbanwide, line 5: Could add:
`private_constant TAX_TYPE`

app/services/commercial/xero_invoice_exporter.rb
```diff
@@ -0,0 +1,97 @@
...
+      "Analysis of data for #{line_item.number_of_meters} meters for #{line_item.school.name}"
+    end
+
+    def self.private_account_fee_description(line_item)
+      "Private account fee for #{line_item.school.name}"
+    end
+
+    private
+
+    def headers
```
urbanwide, line 37: These could also be a private constant but also fine this way

urbanwide (approved):
Looks good. A few very minor comments

app/services/commercial/xero_invoice_exporter.rb
ldodds: I'm referencing this and the label methods in the spec rather than duplicating the strings, so not sure it makes sense to make them private.

app/services/commercial/xero_invoice_exporter.rb
ldodds: Will leave for now. This has to be tested against Xero, I've had to take a guess at the minimal set of columns based on their docs, but might need to be revised along with the descriptions.

PR #5462, opened 2026-07-02
enable rails 7.2 defaults, enables YJIT

tbhi:
enables
- https://guides.rubyonrails.org/configuring.html#config-yjit
- https://guides.rubyonrails.org/configuring.html#config-active-record-postgresql-adapter-decode-dates


tbhi:
running on test

ldodds:
Not seeing any notable impact on test server. Memory usage since deployed to test seems higher (as expected) but also less variable. Should be within capacity for production, so think this is OK to release.

ldodds (approved):

PR #5467, opened 2026-07-02
Add activities and intervention csv and filters

LukeGouldson:
Adds filters to the activities and interventions pages, and allows csv download


app/controllers/admin/reports/activities_controller.rb
```diff
@@ -1,8 +1,36 @@
...
+
+      def index # rubocop:disable Metrics/AbcSize
+        @activities = Activity.includes(:observations,
+                                        :school,
+                                        :activity_type,
+                                        observations: :created_by,
+                                        school: :school_group,
+                                        rich_text_description: { embeds_attachments: :blob })
+                              .recorded_in_last_year.order(created_at: :desc)
+        @activities = @activities.for_school_group(params[:school_group]) if params[:school_group].present?
```
urbanwide: Maybe these common filters (they are the same for activities and actions?) could be put in a concern - something (with a better name! Maybe admin_task_fiterable.rb or similar):

```
# app/controllers/concerns/filterable.rb
module Filterable
  extend ActiveSupport::Concern

  private

  def apply_filters(scope)
    scope = scope.for_school_group(params[:school_group]) if params[:school_group].present?
    scope = scope.for_admin(params[:admin]) if params[:admin].present?
    scope = scope.for_school(params[:school]) if params[:school].present?
    scope = scope.for_user_role(params[:user_role]) if params[:user_role].present?
    scope
  end
end
```
Then use
```
  include Filterable # top of file

    # inside method:
    @activities = apply_filters(@activities)
```

app/models/activity.rb
```diff
@@ -53,16 +54,23 @@ class Activity < ApplicationRecord
   validates :happened_on, presence: true
 
   scope :for_activity_type, ->(activity_type) { where(activity_type: activity_type) }
-  scope :for_school, ->(school) { where(school: school) }
+  scope :for_school, ->(school) { where(school:) }
```
urbanwide: These common scopes could also go in a concern used by both activity and observation?

app/views/admin/reports/activities/index.html.erb
```diff
@@ -36,4 +36,56 @@ The <%= fa_icon('check-circle text-success') %> icon in the images column indica
 images provided by the user
 </p>
 
+<div class="clearfix bg-light py-2 px-2 my-2 border rounded">
+  <%= form_tag admin_reports_activities_path, method: :get do |f| %>
+    <div class="d-flex justify-content-between align-items-center">
```
urbanwide: All these common form elements for activities and interventions could go in a shared partial.

app/views/admin/reports/activities/index.html.erb
```diff
@@ -36,4 +36,56 @@ The <%= fa_icon('check-circle text-success') %> icon in the images column indica
...
+        </span>
+      </div>
+      <div class="d-flex justify-content-between">
+        <%= submit_tag 'Filter', class: 'btn btn-sm' %>
+        <%= link_to 'CSV for below table',
+                    url_for(
+                      controller: controller_path,
+                      action: :index,
+                      only_path: true,
+                      **params.permit(:search, :school_group, :admin, :school, :user_role).merge(format: :csv)
```
urbanwide: params.permit etc would normally be done in the controller...

app/models/activity.rb
```diff
@@ -80,6 +88,25 @@ def academic_year = school.academic_year_for(happened_on)
...
+  def display_title
+    title.nil? ? activity_type.name : title
+  end
+
+  def self.csv_headers
+    ['School Group', 'Admin', 'School', 'User', 'User Role', 'User Staff Role', 'Recorded', 'Happened', 'Title',
+     'Images?']
+  end
+
+  def self.csv_attributes
```
urbanwide: I am in two minds whether or not the csv export should be done this way, as it's quite problem specific (and not a general export of the model). 

Recommend creating a .csv.ruby file for the export (the comparisons stuff does this a lot if you need an example)

urbanwide:
The paging isn't working properly on the activities or interventions page. If I click on any page other than the first, there is no records shown...

Also not a fan of the filter and csv buttons stretching vertically to fit the max height. Can they just be regular sized? Perhaps the csv button should just be called either 'CSV', 'Export CSV' (unless you have been told otherwise anyway!)

Screenshot 2026-07-07 at 13 31 32

urbanwide (changes requested):

LukeGouldson:
Yes, a lot of what I've done here was based on the existing filter/csv system from the admin meter reports as I thought that was most applicable, but I can see that these are older files and therefore might not have used best practice. Will do these edits, thanks!

LukeGouldson:
Fairly sure that the paging is working fine on both pages and Dashboard variants - are you looking at your personal dashboard page perhaps?

urbanwide:
> Fairly sure that the paging is working fine on both pages and Dashboard variants - are you looking at your personal dashboard page perhaps?

Yes it's on my dashboard...

app/controllers/admin/reports/activities_controller.rb
```diff
@@ -1,8 +1,44 @@
...
+                                        observations: :created_by,
+                                        school: :school_group,
+                                        rich_text_description: { embeds_attachments: :blob })
+                              .recorded_in_last_year.order(created_at: :desc)
+        @activities = apply_filters(@activities)
+
+        format
+      end
+
+      def format
```
urbanwide: I'd be inclined to remove the format method and swap things around as follows (more follows convention). Same for interventions controller too.

```
def index
  @activities = fetch_activities

  respond_to do |format|
    format.html do
      @pagy, @activities = pagy(@activities)
    end

    format.csv do
      @headers = headers
      response.headers['Content-Type'] = 'text/csv'
      response.headers['Content-Disposition'] = "attachment; filename=#{filename}"
      render partial: 'table'
    end
  end
end

private

def fetch_activities
  activities = Activity.includes(
    :observations,
    :school,
    :activity_type,
    observations: :created_by,
    school: :school_group,
    rich_text_description: { embeds_attachments: :blob }
  ).recorded_in_last_year
   .order(created_at: :desc)

  apply_filters(activities)
end

def filename
  EnergySparks::Filenames.csv('activities')
end
```

app/views/admin/reports/activities/index.html.erb
```diff
@@ -36,4 +36,6 @@ The <%= fa_icon('check-circle text-success') %> icon in the images column indica
 images provided by the user
 </p>
 
+<%= render 'admin/reports/shared/activity_interventions_filters' %>
```
urbanwide: Can we call these 'recording_filters' please

urbanwide (commented):
Couple of other suggestions (along with the paging issue we've just been chatting about)

urbanwide (approved):
Looks good 👍🏻

PR #5474, opened 2026-07-03
Metered solar config report

tbhi:
- put the query in `models/report` - perhaps good location moving forward rather than having large models
- attributes links to all the meter attributes as that seemed more useful
- update report index and meter report to bootstrap 5
- improve templeate reuse


app/assets/stylesheets/_advice_page.scss
```diff
@@ -21,7 +21,7 @@
 .advice-table {
   thead {
     background-color: $header-dark !important;
-    color: $white;
+    color: $white !important;
```
tbhi, line 24: needed for text to be white with bs5

app/controllers/admin/reports/base_import_reports_controller.rb
```diff
@@ -27,7 +27,7 @@ def columns # rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength
           Column.new(:procurement_route,
                      ->(meter) { meter.procurement_route&.organisation_name },
                      ->(meter, csv) { csv && link_to(csv, admin_procurement_route_path(meter.procurement_route)) }),
-          Column.new(:meter_status,
+          Column.new(:admin_meter_status,
```
tbhi, line 30: to match filter

app/controllers/admin/reports/metered_solar_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      def title = 'Metered Solar'
+
+      def columns
+        column_names = %i[school_group admin school meter data_source supplier admin_meter_status]
+        columns = super.filter { |column| column_names.include?(column.name) }
+        columns.insert(column_names.index(:meter) + 1, BoolColumn.new(:active))
+        columns + [date_column(:start_date),
+                   date_column(:end_date),
+                   real_generation_meters_column,
```
ldodds, line 16: Now that I'm looking at it, I think it'd be best to revise some of the column ordering, so that they last few columns are grouped, e.g:

- real generation meters
- modelled generation meter 
- real export meter
- modelled solar
- solar override

ldodds (approved):

PR #5475, opened 2026-07-06
Added supplier and data source to reports

LukeGouldson:
Adds supplier and data source to both supplier and data source reports


app/controllers/admin/suppliers_controller.rb
```diff
@@ -13,7 +13,7 @@ def index
     def show; end
 
     def deliver
-      @supplier = Supplier.expect(:supplier_id)
+      @supplier = Supplier.find(params.expect(:supplier_id))
```
tbhi, line 16: I think worth adding `on: :member` to the route so `load_and_authorize_resource` does this

app/views/admin_mailer/school_data_source_report.html.erb
tbhi, line 1: looks like the 2 files are the same so could use a partial for them

tbhi, line 1: perhaps the csv should be updated with the same new columns?

tbhi (approved):

PR #5481, opened 2026-07-08
Update Solar PV Areas

ldodds:
The PV Live API has recently changed the boundaries they are using but we missed updating our definitions.

This PR:

- updates the system with the latest gsp ids, names and centroids
- reassigns all schools to the nearest areas
- removes some obsolete areas

Also some tidying of the admin interface to make it easier to check individual areas

Ideally need to run `rake data_feeds:solar_pv_tuos_back_fill` after latest release.


lib/tasks/deployment/20260708133854_solar_areas2026.rake
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+namespace :after_party do # rubocop:disable Metrics/BlockLength
+  desc 'Deployment task: solar_areas2026'
+  task solar_areas2026: :environment do # rubocop:disable Metrics/BlockLength
```
tbhi, line 5: worth putting in a transaction?

lib/tasks/deployment/gsp-areas-2026.csv
tbhi, line 1: is this supplied straight from PV Live?

tbhi (approved):

lib/tasks/deployment/gsp-areas-2026.csv
ldodds: No, i have to generate the lat/longs from the centroids of the polygons from NESO, then merge that with a CSV download from PV Live. I've not looked at automating that as we only do this 1-2 a year.

tbhi: just noticed there's a duplicate 279

ldodds: I must have missed that one. It will get removed during the update though.

Happens because the NESO GSP areas are sometimes non-contiguous geographic areas, which PV Live treats as a single location. For now I've just used centroid of one area to determine best area for a school.

A better approach might be to use geospatial containment, but needs some rethink of the process and probably added PostGIS extensions in order import/query boundaries.

PR #5482, opened 2026-07-08
Update action pages to latest designs

urbanwide:
This has both activity and action page changes in this PR (could change the base but think it is best to the activity one, but think best to review these together). Probably should go on test for a bit first


app/components/layout/cards/feature_component.rb
```diff
@@ -30,24 +30,32 @@ class FeatureComponent < LayoutComponent
...
 
-      def initialize(size: :md, **_kwargs)
+      SIZES = { xl: 1, lg: 2, md: 3, sm: 4 }.freeze
+      private_constant :SIZES
+      TAG_POSITIONS = %i[top bottom].freeze
+      private_constant :TAG_POSITIONS
+
+      def initialize(size: :md, tag_position: :top, **_kwargs)
         super
+        validate_inclusion size: size, in: SIZES.keys
```
ldodds: Very minor comment, but could use the newer style here, e.g. `validate_inclusion size:, in: SIZES.keys`

app/controllers/activity_types_controller.rb
```diff
@@ -34,7 +40,7 @@ def for_school
 
   private
 
-  def can_be_completed_for_schools(activity_type, user)
+  def available_schools_for_user(activity_type, user)
```
ldodds, line 43: The values here seem to be different than in the interventions controller? Although might result in the same actual list. Can we align them?

app/models/activity_type.rb
```diff
@@ -48,6 +48,7 @@ class ActivityType < ApplicationRecord
   TX_REWRITEABLE_FIELDS = [:description_cy, :school_specific_description_cy, :download_links_cy].freeze
 
   belongs_to :activity_category
+  alias category activity_category
```
ldodds, line 51: Noticed that you'd added the same to intervention type. Perhaps in a separate PR we should just extract all the common elements of both of these models into a common base class (`Task`?) or a concern?

app/views/activity_types/_activity_type_score.html.erb
ldodds, line 1: Not saying to do it here, but should probably make this a component at some point. The if statement its wrapped in could be moved to `render?`

app/views/activity_types/show.html.erb
```diff
@@ -3,60 +3,45 @@
...
-    <%= render 'shared/activities/prompt_login_to_record', label: 'activity', score: @activity_type.score,
-                                                           return_path: activity_type_path(@activity_type) %>
+    <%= render 'shared/tasks/prompts/login_to_record', task: @activity_type %>
   <% end %>
 
-  <% if can?(:create, Activity) && @can_be_completed_for_schools.present? %>
-    <% if @can_be_completed_for_schools.one? %>
-      <%= render 'activity_types/prompt_score', activity_type: @activity_type,
-                                                school: @can_be_completed_for_schools.first %>
+  <% if can?(:create, Activity) && @available_schools_for_user.present? %>
```
ldodds, line 26: Another potential candidate for a component at some point.

config/locales/cy/views/tasks/tasks.yml
ldodds, line 1: When this is merged, I'll need to update transifex with the new files and also manually add the translations where we've just moved things around. Will avoid the translators having to do them from scratch. Can you assign me a Trello card to do that?

ldodds (approved):
Looks good. The UX is good and nice to see some tidying and reuse.

As noted in a few of the components there's probably scope to do some follow-up changes to extract more common elements from ActivityType, etc into a common base class to highlight more of the similarities.

I also think that some of the partials you've added under `shared/tasks` could be components, but that's something we can add later. The PR is big enough already :)

There are still some images that get cropped, including some of our own illustrations. For example `/intervention_types/21` and `/intervention_types/30`. Perhaps we need to resize some of the images themselves if we can't get the fit right in the CSS. Do have an idea of what size/ration is best so we can work to that?

app/models/activity_type.rb
urbanwide: These are not common as they are aliasing to different method names. However, could make them methods to override in a base class / concern

app/controllers/activity_types_controller.rb
urbanwide: Was trying to mostly do cosmetic stuff mostly with a bit of tidying underneath.

ActivityTypesController excludes schools from the list where the activity type has already been completed this year. InterventionTypesController just picks schools that the current user has permissions for.

Given a user can complete activity types multiple times in a year, I am not sure restricting this way is correct anyway?

urbanwide:
As above - made these changes cosmetically and with a first step towards consolidation - definitely agree some new components would be nice here too. Perhaps can put a card in for this.

> There are still some images that get cropped, including some of our own illustrations. For example `/intervention_types/21` and `/intervention_types/30`. Perhaps we need to resize some of the images themselves if we can't get the fit right in the CSS. Do have an idea of what size/ration is best so we can work to that?

I've just changed this so rather than using fit (i.e. object-fit cover), that they are responsive in size. See what you think. Less requirements then to get the aspect ratio right for now. Typically I think 3:2 (or maybe 4:3) would work best though (landscape) in response to your question. Perhaps we could fix the images later then maybe go back to using fit after that if we think it works.

Would you mind having a quick look @ldodds. Also a question about consolidation of methods above

app/controllers/activity_types_controller.rb
ldodds: Yes doesn't make sense. But don't start changing that here, suggest adding trello card and revisiting in next phase of work in the autumn

app/models/activity_type.rb
ldodds: Yes, that's what I was thinking. But again, don't need to do it immediately.

ldodds:
Took a quick look at the images, but I'm not really seeing any differences in the images, the examples I linked to are both still cropped at the bottom, e.g.:

Screenshot from 2026-07-13 13-33-30

Again, this could be resolved separately.

urbanwide:
Hmm, I am seeing them like this:
image
image

ldodds:
I get that if I reduce window width to about 50% of my screen width but there's a breakpoint where it switches to being cropped. Can't figure how how to work that out at the minute.

urbanwide:
I am not seeing this, but not sure why. OK will leave as is for now if that's ok and perhaps put on test first? Or is it ok to merge straight to master (then test)?

ldodds:
It's only worth putting on test if you're going to be asking for internal feedback. The test server is locked down so there's no incidental crawler traffic, etc if you were hoping that might highlight some issues.

urbanwide:
Ahh ok, then will just merge to master. Thanks.

PR #5484, opened 2026-07-09
Ensure licences for existing schools have correct default status

ldodds:
Licences move from provisional to confirmed, then pending invoice and finally invoiced.

Licences for schools that are data enabled, should have a pending invoice status if their contract is confirmed. This is currently applied if a school is onboard and then made data enabled. But there's a bug where this isn't happening correctly for renewed/future licences.

- [x] Ensure that when contracts are confirmed that licences are updated to pending invoice if school is already data enabled
- [x] Ensure that when licence is added for a school, via the licence form, it has the right status
- [x] Ensure that when licence is added for a school, via the bulk licence editor, it has the right status
- [x] Add migration to fix status of existing licences for renewed contracts.

### DIFF
--- a/app/services/commercial/contract_manager.rb
+++ b/app/services/commercial/contract_manager.rb
@@ -43,12 +43,7 @@ module Commercial
     def update_licence_statuses
       # update status for those that have not already advanced
       @contract.licences.where.not(status: %i[pending_invoice invoiced]).find_each do |licence|
-        status = if @contract.confirmed? && licence.school.data_enabled?
-                   :pending_invoice
-                 else
-                   @contract.status
-                 end
-        licence.update!(status:)
+        licence.update!(status: @contract.status)
       end
     end
   end
--- a/app/services/commercial/contract_manager.rb
+++ b/app/services/commercial/contract_manager.rb
@@ -43,7 +43,12 @@ module Commercial
     def update_licence_statuses
       # update status for those that have not already advanced
       @contract.licences.where.not(status: %i[pending_invoice invoiced]).find_each do |licence|
-        licence.update!(status: @contract.status)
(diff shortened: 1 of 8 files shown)


app/services/commercial/contract_manager.rb
```diff
@@ -43,7 +43,12 @@ def update_licence_dates
     def update_licence_statuses
       # update status for those that have not already advanced
       @contract.licences.where.not(status: %i[pending_invoice invoiced]).find_each do |licence|
-        licence.update!(status: @contract.status)
+        status = if @contract.confirmed? && licence.school.data_enabled?
```
tbhi: is this needed with the before hook on license?

lib/tasks/deployment/20260709173551_fix_licence_status.rake
```diff
@@ -0,0 +1,23 @@
...
+  desc 'Deployment task: fix_licence_status'
+  task fix_licence_status: :environment do
+    puts "Running deploy task 'fix_licence_status'"
+
+    # Find confirmed contracts and licences where the school is data enabled and set the
+    # licence to be 'pending_invoice'. Corrects for an existing bug where this status should
+    # have been used already.
+    Commercial::Licence
+      .joins(:contract, :school)
+      .where(start_date: Time.zone.today..)
```
tbhi, line 13: should it apply to older licenses?

tbhi (approved):

lib/tasks/deployment/20260709173551_fix_licence_status.rake
ldodds: Yes, but actual status of these is unclear. Need to confirm whether any have now been invoiced, so will do a follow-up PR for those.

PR #5485, opened 2026-07-09
Tidy admin manage menus

LukeGouldson:
Ensures the options in the Group Settings admin section and Manage Group menu are the same, adding meter attributes and meter updates
Adds breadcrumbs and the Manage Group menu to admin school group pages, and the Manage Group menu to admin dashboard pages


ldodds:
The links, breadcrumbs, etc all look good, although think you've missed some:

- "Make sure all the pages under the Admin section are displaying breadcrumbs". Breadcrumbs are missing from "Edit group", "issues & notes", "users", "partners", "meter attributes", "meter updates". These pages aren't using the same layout so don't have the settings navigation either. Might be why the breadcrumbs aren't in place but would like to get all the pages aligned.

ldodds (approved):

PR #5491, opened 2026-07-10
Fix OneDayAMRReading.new constructor arguments

ldodds:
I'd missed changing a couple of calls to `OneDayAMRReading` constructor so loading data from RBee sites isn't working correctly.

I've added a very basic spec to test the `DataFeeds::LowCarbonHubMeterReadings` as there wasn't one at all for that or the underlying API. Have added a fixture with valid response and just confirmed that an API call is made and data is returned in expected shape.

That's enough to confirm the fix to the constructor. Have held off any further refactoring or changes to the API client code for now.


spec/lib/data_feeds/low_carbon_hub_meter_readings_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        'login' => 'username',
+        'serialNumber' => '218854673',
+        'startDate' => (Date.new(2026, 7, 4) - 1.hour).strftime('%Y-%m-%dT%H:%M:%S'),
+        'endDate' => (Date.new(2026, 7, 9) + 1.day).midnight.strftime('%Y-%m-%dT%H:%M:%S'),
+        'step' => 'tenmin'
+      )
+    end
+
+    it 'returns expected data' do
+      expect(response.keys).to include(:solar_pv, :electricity, :exported_solar_pv)
```
tbhi: think I'd `expect(response.transform_values { |h| h.slice(:mpan_mprn) }).to eq(..`

maybe also check a reading value

tbhi (approved):

PR #5504, opened 2026-07-13
ensure meter type school level data is set for automated pupil number updates

tbhi:
- made floor area optional for the attribute so allows only setting the pupil number
- added an integration test that the meter collection picks up value set by the updater


ldodds (approved):
Looks good. Can we get this onto test ahead of a production release, as once this is in place it'll be the first time we do a regeneration with automatically added attributes. There's unlikely to be any issues but worth checking.

PR #5521, opened 2026-07-16
Remove funder field from onboardings and schools

ldodds:
Removes the old direct associations between the Funder, SchoolOnboarding and School models. The relationship is now via the Contract and Licensing model. E.g. a School is associated with a Funder if they have a current licence covered by one of the Funder's contracts.

The PR:

- Removes the old associations, updating all the forms and code that uses them
- Remove the old Funder reports/view that summarised these associations, there are already existing replacements that use the C&L models
- Update reports and exports that identify the current Funder for a school to use the new "summarise current contract holder" method on School.
- Change the school comparison report interface to continue to allow filtering by Funders, but via the indirect relationships between the model

A later step will look at how we synchronise the new indirect relationships with Mailchimp, but will do that later when requirements are clear.

Have taken a back-up of the old associations using the existing Funder Allocation Report just in case needed.


app/models/intervention_type.rb
```diff
@@ -17,6 +17,12 @@
 #
 # Indexes
 #
+#  index_intervention_types_on_intervention_type_group_id  (intervention_type_group_id)
```
tbhi, line 20: looks like empty line previously causing this to be duplicated

spec/services/schools/funder_allocation_report_service_spec.rb
```diff
@@ -103,7 +101,7 @@
     # only basic data, helps to catch errors checking for nils
     let!(:school_2) do
       create(:school, visible: true, active: false, removal_date: nil, archived_date: Time.zone.today,
-                      school_group: create(:school_group), funder: funder_2)
```
tbhi, line 106: funder_2 now not used?

tbhi (approved):

app/models/school.rb
tbhi: also remove this?

tbhi: another

tbhi (commented):
also this? https://github.com/Energy-Sparks/energy-sparks/blob/90caa1a9407fc7180e62e8cfd7eeff0d993069cb/app/controllers/schools_controller.rb#L193

PR #5522, opened 2026-07-16
Update activity and action recording forms

urbanwide:
* Styled activity and action forms.
* New Tasks::Header component
* Have introduced a new yellow-very-pale for using on the pupil activity backgrounds, as the lightest one we have is a bit dark. Not sure it's right but see what you think @ldodds.
* Quite a bit of refactoring


ldodds (approved):

ldodds:
Looks good and I think the pale yellow is ok

PR #5523, opened 2026-07-16
Remove unused default_contract_holder field

ldodds:


app/controllers/schools/self_funded_controller.rb
tbhi, line 1: also app/views/schools/self_funded/

app/components/commercial/licences_component/licences_component.html.erb
tbhi, line 1: should @show_renewal_data also be removed from the component?

tbhi (approved):

app/components/commercial/licences_component/licences_component.html.erb
ldodds: No, the attribute is being used to control when to add more context in the display when displaying licences for renewal.

PR #5525, opened 2026-07-16
Implement new "our impact" page

ldodds:
Adds a new "Our impact" page which summarises our impact for the last academic year.

The page is backed by a new model which will hold the metrics that are calculated annually. This is done outside the application so we just need a way to store the results.

The page has separate blocks of metrics for total savings, average primary and secondary school savings, reach and behaviour change. It also features two testimonials which allow us to showcase some case studies.

The metrics shown on the home page are also now driven by the newly added model. This currently has a fall back to some static figures to make sure the page continues to work.


app/controllers/admin/organisation_statements_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    private
+
+    def organisation_statement_params
+      params.expect(
+        organisation_statement: %i[academic_year actions activities efficiency_report_link
+                                   first_testimonial_id second_testimonial_id
+                                   primary_carbon_saving primary_cost_saving primary_saving_electricity
+                                   primary_saving_electricity primary_saving_gas pupils schools
```
tbhi: looks like duplicate `primary_saving_electricity` - perhaps one per line would make it clearer?

spec/factories/testimonials.rb
```diff
@@ -7,8 +7,9 @@
     image { Rack::Test::UploadedFile.new(Rails.root.join('spec/fixtures/images/pupils-jumping.jpg'), 'image/jpeg') }
 
     # Translated attributes
-    title_en { 'Testimonial title in English' }
-    title_cy { 'Testimonial title in Welsh' }
+    sequence(:title_en) { |n| "Testimonial #{n} title in English" }
+    sequence(:title_cy) { |n| "Testimonial #{n} title in English" }
```
tbhi: in Welsh?

app/views/admin/organisation_statements/show.html.erb
```diff
@@ -0,0 +1,67 @@
...
+  </div>
+<% end %>
+
+<div class="row">
+  <div class="col-12 mb-4 bg-light rounded-2 pb-3">
+    <h3>Our reach</h3>
+    <p>Schools: <%= number_with_delimiter @organisation_statement.schools %></p>
+    <p>Pupils: <%= number_with_delimiter @organisation_statement.pupils %></p>
+    <p>Staff: <%= number_with_delimiter @organisation_statement.staff %></p>
+    <p>Activities: <%= number_with_delimiter @organisation_statement.activities %></p>
```
tbhi: only needed below?

app/models/impact_report/organisation_statement.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    belongs_to :second_testimonial,
+               optional: true,
+               class_name: 'Testimonial'
+
+    validates :academic_year, presence: true, uniqueness: true
+    validates :current, uniqueness: { message: 'already exists' }, if: :current? # rubocop:disable Rails/I18nLocaleTexts
+
+    scope :current_statement, lambda {
+      where(current: true).first
```
tbhi: perhaps makes more sense as static method since not chainable, e.g.
```
def self.current_statement
  find_by(current: true)
end
```

tbhi (approved):

PR #5530, opened 2026-07-17
Fix icon colour issue for dark themed cards

urbanwide:
This works but not sure it's as robust as I thought, as it's dependant on the order css is output by asset pipeline


ldodds:
Rather than changing theme-dark why not just add some CSS for the prompt component specifically for now to set the default colour for its icons only. That's the only place you needed it?

This will fix the icons on the home page (and elsewhere I've spotted). Can look at an alternative fix later?

ldodds (approved):

PR #5532, opened 2026-07-17
 Integrate with MeterZ API 

tbhi:
- consolidated code with other solar installation where possible
-


app/services/solar/meter_z_download_and_upsert.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    def parse_reading(reading)
+      datetime = DateTime.parse(reading['reading_timestamp'])
+      accumulated = reading['readings']['accumulated_kilowatt_hours'].to_f
+      [datetime.to_date, accumulated, hh_index(datetime)]
+    end
+
+    def convert_from_accumulated(readings_by_day)
+      readings_by_day.filter_map do |date, readings|
+        next if readings.any?(&:nil?)
```
tbhi, line 56: perhaps should be less strict

app/models/meter_z_installation.rb
```diff
@@ -0,0 +1,68 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_...  (amr_data_feed_config_id => amr_data_feed_configs.id) ON DELETE => cascade
+#
+class MeterZInstallation < ApplicationRecord
+  belongs_to :amr_data_feed_config
+
+  has_many :meters, dependent: nil
+  # has_many :meter_z_installation_schools, dependent: nil
```
ldodds: can remove?

app/services/solar/meter_z_download_and_upsert.rb
```diff
@@ -0,0 +1,61 @@
...
+      end
+    end
+
+    def convert_readings(readings)
+      readings_by_day = convert_to_readings_by_day(readings)
+      convert_from_accumulated(readings_by_day)
+    end
+
+    def convert_to_readings_by_day(readings)
+      readings_by_day = Hash.new { |hash, key| hash[key] = Array.new(49, nil) }
```
ldodds, line 39: 48?

lib/tasks/deployment/20260720101712_create_meter_z_config.rake
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+namespace :after_party do
+  desc 'Deployment task: create_meter_z_config'
+  task create_meter_z_config: :environment do
+    puts "Running deploy task 'create_meter_z_config'"
+
+    AmrDataFeedConfig.create!(description: 'MeterZ',
+                              identifier: 'meter-z',
+                              date_format: 'n/a',
```
ldodds: should still contain a date format, the unvalidated readings and meter loading reports rely on this field

lib/data_feeds/meter_z.rb
```diff
@@ -0,0 +1,68 @@
...
+    def initialize(api_key)
+      @api_key = api_key
+      @connection = FaradayHelper.connection(url: BASE_URL, headers: { 'x-api-key' => @api_key }) do |f|
+        f.adapter CustomAdapter
+        f.response :json
+      end
+    end
+
+    def readings(organisation_id, site_id, meter_id, start)
+      # readings returned from start_datetime in descending order to end_datetime so if you want the last week of
```
ldodds, line 31: The API does support changing the default order, so we could request/process in ascending order

lib/data_feeds/meter_z.rb
```diff
@@ -0,0 +1,68 @@
...
+      # readings end_datetime should be set to 1 week ago
+      if start.nil?
+        start_datetime = Date.tomorrow.iso8601
+      else
+        end_datetime = start.iso8601
```
ldodds, line 36: If the API is returning accumulated readings, then if requesting data for a specific date range, do you need one extra half-hourly reading to calculate the change?

In practice it'd only be around midnight so unlikely to be any actual generation, but just wondering.

ldodds (approved):
Couple of minor notes and a change needed to the feed config. 

Have tested setup of a single school and reviewed loaded data. All seems to be working as expected. Some overnight generation figures in the data, but these are coming from source.

app/services/solar/meter_z_download_and_upsert.rb
tbhi: it stores 2 midnight readings per day to calculate the change so I think 49 is right

lib/data_feeds/meter_z.rb
tbhi: The first reading it's calculated should be the 0030 reading minus the 0000 reading - think that's right?

ldodds: Yes, OK.

PR #5545, opened 2026-07-23
Fix context for school contracts page

ldodds:
Make sure there is a `@school` in scope when showing contracts for schools. Ensures that the standard breadcrumbs and second navigation bar are displayed.

### DIFF
--- a/app/controllers/admin/commercial/contract_holder_contracts_controller.rb
+++ b/app/controllers/admin/commercial/contract_holder_contracts_controller.rb
@@ -13,6 +13,9 @@ module Admin
           @school_group = @contract_holder
           breadcrumbs
           render :index, layout: 'group_settings'
+        elsif @contract_holder.is_a?(School)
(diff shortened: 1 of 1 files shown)


app/controllers/admin/commercial/contract_holder_contracts_controller.rb
```diff
@@ -13,6 +13,9 @@ def index
           @school_group = @contract_holder
           breadcrumbs
           render :index, layout: 'group_settings'
+        elsif @contract_holder.is_a?(School)
+          @school = @contract_holder
+          render :index
```
tbhi, line 18: is this the default so not needed?

tbhi (approved):

app/controllers/admin/commercial/contract_holder_contracts_controller.rb
ldodds: Could be a Funder

PR #5552, opened 2026-07-27
Remove unused alert template variables

ldodds:
The alert base class defines a number of template variables are are unused, e.g. school name or rating. But they aren't necessary as they're not used in the template text for displaying alerts or the alert framework already captures them on the alert record. 

They're just taking up space in the database as they're repeated for individual alerts and across all alert types. They also increase size of JSON produced when deserializing the variables in queries, etc.

They're mostly calculated on demand so removing the template variable declarations stops them from being calculated and saved.

The `AlertAdditionalPrioritisationData` alert also had some unused variables. These were used in the old benchmarking pages but we've since reimplemented that. I've just kept those that are currently used in `Comparison::View`s and removed the others.


analytics/lib/dashboard/alerts/common/alert_additional_prioritisation_data.rb
tbhi: looks like these need updating too

tbhi (approved):

PR #5562, opened 2026-07-30
Change Xero Account Codes to be Strings

ldodds:
Need to allow for optional leading zeros, so switch code to a string.

### DIFF
--- a/db/migrate/20260730142636_change_xero_account_code_to_string.rb
+++ b/db/migrate/20260730142636_change_xero_account_code_to_string.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+class ChangeXeroAccountCodeToString < ActiveRecord::Migration[8.1]
+  def up
+    add_column :commercial_xero_account_codes, :code_str, :string
+
+    execute <<~SQL.squish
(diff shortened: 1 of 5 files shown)


db/migrate/20260730142636_change_xero_account_code_to_string.rb
```diff
@@ -0,0 +1,29 @@
...
+class ChangeXeroAccountCodeToString < ActiveRecord::Migration[8.1]
+  def up
+    add_column :commercial_xero_account_codes, :code_str, :string
+
+    execute <<~SQL.squish
+      UPDATE commercial_xero_account_codes
+      SET code_str = code::text;
+    SQL
+
+    remove_column :commercial_xero_account_codes, :code
```
tbhi, line 12: think `change_column :commercial_xero_account_codes, :code, :string, using: 'code::text'` would also work

tbhi (approved):

db/migrate/20260730142636_change_xero_account_code_to_string.rb
ldodds: Thanks, wasn't aware of that. Will use that style in future.

PR #5563, opened 2026-07-30
Make overlapping invoices warning more prominent

ldodds:
- Move higher on the contract page
- Add to the raise invoice page


app/views/admin/commercial/contracts/_licences_warning.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if @overlapping_licences.any? %>
```
tbhi: think passing variable would recommended

tbhi (approved):

PR #5564, opened 2026-07-31
Destroy schools that were soft deleted more than 3 years ago

ldodds:
Adds a service (and monthly cronjob) to remove schools that have been archived or soft-deleted for over three years. This ensures that the school data is removed rather than accumulating over time.

Few things to note in PR changes:

- Have renamed the current SchoolRemover to Schools::Remover. That code is largely unchanged other than addressing some rubocop comments.
- The new Destroyer class has slightly different logic for archived and deleted schools to ensure that necessary changes are done before destroying the School
- In reviewing the school associations I've tidied them all up and added dependent clauses. Existing HABTM have been replaced with `has_many :through` so dependent clauses can be added
- I've added before_destroy calls to a number of classes to ensure that they can't be deleted where additional changes are required first, e.g. consent being removed before destroying a meter, or removing schools from cluster users
- I've implemented behaviour to allow contracts, licences and invoices to be deleted. But only where the contract holder is a school that has been archived/soft-deleted for over three years

I've written tests for the deletion constraints and done some manual testing of the Destroyer. But plan to run this on test first to confirm its works OK with real-world data.

In doing the work I noticed that a number of our foreign keys have delete cascade behaviour defined on them. This was done early in the project (around 2019). The relevant associations don't always have `dependent` behaviour defined, so the deletion is possibly unexpected.

I've declared the behaviour on all of the school associations, but suggest we consider removing that cascades from all foreign keys so that we can fully control behaviour in ActiveRecord. This will require a lot of index changes, so haven't included here as we may want to consider when that is run.


spec/models/meter_spec.rb
```diff
@@ -524,23 +527,42 @@
...
+      it 'cannot be destroyed' do
+        expect { meter.destroy }.not_to(change(described_class, :count))
+        expect(meter.errors[:base]).to include('Audit requirements mean consent must be withdrawn before removing meter')
+      end
+    end
+
+    context 'when consent has not been granted, or already withdrawn' do
+      let!(:meter) { create(:electricity_meter, consent_granted: false) }
+
+      it 'cannot be destroyed' do
```
tbhi: it 'can be destroyed'

tbhi (approved):

PR #5574, opened 2026-08-05
Remove unused columns in unvalidated readings table

ldodds:
The `AmrDataFieldReading` model has some fields which were added early in the project but are not needed.

These include:

- holding a description or name for the meter
- a total for the day
- a unique reference for the meter or reading
- the name of the school
- the postcode of the school

These were only ever all available for the first data feed config but a couple of fields have been added to a few later configs. However the fields aren't always inserted any more, and even when stored are never actually used when processing or exporting data.

This PR removes the columns and the related configuration.


app/models/amr_data_feed_reading.rb
```diff
@@ -38,7 +31,6 @@
 #  index_amr_data_feed_readings_on_created_at_and_meter_id      (created_at,meter_id)
 #  index_amr_data_feed_readings_on_meter_id                     (meter_id)
 #  index_amr_data_feed_readings_on_mpan_mprn                    (mpan_mprn)
-#  unique_meter_readings                                        (mpan_mprn,reading_date) UNIQUE
```
tbhi, line 41: this wasn't removed?

tbhi (approved):

app/models/amr_data_feed_reading.rb
ldodds: No. But annotaterb removed the annotation. Tried rerunning and adding `-i` flag but it still didn't add it back in. I've manually changed, but obviously something has changed in annotaterb recently that is causing all the comments to be revised.

PR #5575, opened 2026-08-06
Remove old readings

ldodds:
Adds a monthly cronjob to tidy up old readings from the database. Set to run on second of the month to avoid clashing with deleting soft-deleted schools.

Will result in

- AmrDataFeedReadings created 3 years ago and not associated with a meter, being removed
- AmrUploadedReadings created 3 years ago, being removed, along with associations
- AmrDataFeedImportLogs no longer referenced from AmrDataFeedReadings being removed, along with associations

Import logs can become "orphaned" through natural updates if readings are reloaded, so there's a separate scope and deletion for those rather than relying on an association.

Will need to `VACUUM ANALYSE amr_data_feed_readings` the first time this is run, as there's a large amount of data being removed. So suggest running immediately after the code is released.  When run monthly it should be OK to rely on the database autovacuum.

### DIFF
--- a/app/models/amr_data_feed_reading.rb
+++ b/app/models/amr_data_feed_reading.rb
@@ -52,6 +52,8 @@ class AmrDataFeedReading < ApplicationRecord
   belongs_to :amr_data_feed_import_log
   belongs_to :amr_data_feed_config
 
+  DELETE_THRESHOLD = 3.years
+
   CSV_HEADER_DATA_FEED_READING = 'School URN,Name,Mpan Mprn,Meter Type,Reading Date,Reading Date Format,Record Last Updated,00:30,01:00,01:30,02:00,02:30,03:00,03:30,04:00,04:30,05:00,05:30,06:00,06:30,07:00,07:30,08:00,08:30,09:00,09:30,10:00,10:30,11:00,11:30,12:00,12:30,13:00,13:30,14:00,14:30,15:00,15:30,16:00,16:30,17:00,17:30,18:00,18:30,19:00,19:30,20:00,20:30,21:00,21:30,22:00,22:30,23:00,23:30,00:00'.freeze
 
   PARSED_DATE = <<~SQL.squish.freeze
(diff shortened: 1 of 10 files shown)


app/models/manual_data_load_run.rb
```diff
@@ -19,7 +19,7 @@
 class ManualDataLoadRun < ApplicationRecord
   belongs_to :amr_uploaded_reading, dependent: :destroy
```
tbhi: should there be a `dependent: :destroy`  here? - now the inverse has `dependent: :destroy` - believe  usually only on the has_many

app/models/amr_uploaded_reading.rb
```diff
@@ -21,8 +21,9 @@
 
 class AmrUploadedReading < ApplicationRecord
   belongs_to :amr_data_feed_config
+  has_one :manual_data_load_run, dependent: :destroy
```
tbhi, line 24: does anything enforce  single?  should be has_many?

app/models/amr_data_feed_reading.rb
```diff
@@ -180,4 +182,11 @@ def self.sort_query_results_by(mpans, query_results)
       end
     end
   end
+
+  def self.delete_old_readings!(before_date: DELETE_THRESHOLD.ago)
+    AmrDataFeedReading.transaction do
+      AmrDataFeedReading.where(meter_id: nil).where(created_at: ...before_date).delete_all
+      AmrDataFeedImportLog.unused.delete_all # remove those no longer referenced from table
```
tbhi, line 189: perhaps should also look at created_at: - unreferenced logs be useful if there were errors?

spec/models/amr_uploaded_reading_spec.rb
```diff
@@ -14,4 +14,22 @@
...
+  describe '.delete_old_readings!' do
+    context 'with an old record' do
+      before { create(:amr_uploaded_reading, created_at: 3.years.ago) }
+
+      it 'deletes the record' do
+        expect { described_class.delete_old_readings! }.to change(described_class, :count).by(-1)
+      end
+    end
+
+    context 'with an newer unused record' do
```
tbhi: with a newer

spec/models/amr_data_feed_reading_spec.rb
```diff
@@ -143,5 +143,51 @@
...
+        it 'deletes the record' do
+          expect { described_class.delete_old_readings! }.to change(described_class, :count).by(-1)
+        end
+
+        it 'cleans up the import log' do
+          expect { described_class.delete_old_readings! }.to change(AmrDataFeedImportLog, :count).by(-1)
+        end
+      end
+
+      context 'with an newer unused record' do
```
tbhi: with a newer

tbhi (approved):

app/models/amr_uploaded_reading.rb
ldodds: You can only create one run per uploaded reading

app/models/amr_data_feed_reading.rb
ldodds: Logs can become unreferenced if the readings in the original import are replaced with a more recent upload, so wanted to clear our unused records not just those that are old.

PR #5576, opened 2026-08-07
Initial estimated data support

ldodds:
Some data feeds that supply data to us include status indicators that show whether a whole day, or individual half-hourly readings are estimated.

For some manual processing of data, estimates are manually removed, but in other cases we just load all the data. This means we're not tracking whether readings are estimated.

This PR revises the system to support capturing whether the readings for a day are estimated or actual. We track this at the day, rather than half-hourly level as we don't need to distinguish between actuals and estimates to that degree of precision.

The changes include:

- revisions to the data loader (specifically configs, data feed translator and upserter) to identify if a reading is estimated and ensure that is stored correctly. 
- updates to existing data feed configs to configure which column(s) hold the flags that indicate Actual/Estimated data
- passing the estimated flag to the analytics via the "EST" reading type, which is otherwise equivalent to "ORIG".
- storing "EST" status on validated readings
- updating admin interfaces (unvalidated data downloads, meter readings calendar, and a new report) to help identify which meters have estimated readings
- include "EST" status in how we calculate meter monthly summaries, so presence of estimated data can be shown on SECR reports

A few things to note:

- some feeds have a single column to indicate the whole day is estimated, others track it at half-hour level. We now capture a set of "reading_status_fields" if any of these indicate there are estimates we treat the whole day as estimated
- the majority of fields use as a standard set of flags, e.g. "E" or "Estimated", so have built those in. But the changes allow adding custom flags, e.g. "Yes" to a config. This helps avoid ambiguity when the flags might have different meanings in different feeds
- while most formats have a clearly labelled separate set of columns for adding flags to readings, in some cases the columns all have the same name, e.g. 48 "Type" columns. And in one case the flag columns have the same names as the readings (e.g. block of HH1-HH48 columns for readings, then HH1-HH48 repeated for flags). Code handles this, the latter case with an extra flag.
- the DataFeedUpserter logic needed changing where we have feeds that are merging in partial days worth of readings. Some complexity in ensuring that we're properly maintaining the estimate flag

I decided not to go with capturing flags at half-hourly level because:

- not all feeds have that level of detail
- 99% of our use cases involve identifying whether any reading in a day is estimated rather than individual half-hours
- the implementation at half-hourly is likely to be less efficient to store and query, and more complex (e.g. we'd have to start tracking not just estimates but also our own half-hourly substitutions)

We could also include that level of detail later if needed but this is a big step forward.

Later improvements will involve exposing more of this to users, e.g. annotating charts and tables.


app/models/amr_data_feed_reading.rb
```diff
@@ -76,24 +76,6 @@ class AmrDataFeedReading < ApplicationRecord
     END parsed_date
   SQL
 
-  def self.download_all_data
```
ldodds, line 79: Unused.

app/views/admin/reports/base_meter_reports/index.html.erb
```diff
@@ -29,7 +29,12 @@
       </p>
       <% if lookup_context.exists?('help', lookup_context.prefixes, true) %>
       <p>
-        <a class="btn btn-sm btn-info" data-toggle="collapse" href="#help" role="button" aria-expanded="false" aria-controls="help">
+        <a class="btn btn-sm btn-info"
```
ldodds, line 32: Just fixing button under bootstrap 5

spec/models/meter_monthly_summary_spec.rb
ldodds, line 1: Reworked the spec to test more scenarios and make it easier to test for ORIG and EST data.

app/views/admin/reports/amr_validated_readings/show.html.erb
```diff
@@ -55,7 +56,7 @@
           <li>
             <span style='background:<%= colour %>;'></span>
             <div class="description" style="font-size: 17px;">
-              <small><%= (@amr_types[type][:name]).to_s + (type == 'MISSING' ? '' : "(#{type})") %></small>
+              <small><%= @amr_types[type][:name]).to_s + (type == 'MISSING' ? '' : "(#{type})" %></small>
```
tbhi, line 59: invalid brackets here? although seems to be rendering

app/controllers/admin/reports/estimated_reads_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+                     ->(meter) { meter.latest_est_reading_date&.iso8601 }),
+          Column.new(:total,
+                     ->(meter) { meter.total }),
+          Column.new(:recent,
+                     ->(meter) { meter.recent_total })
+        ]
+      end
+
+      def results
+        30.days.ago
```
tbhi, line 39: not used?

app/views/admin/reports/base_meter_reports/index.html.erb
```diff
@@ -29,7 +29,12 @@
...
+           data-toggle="collapse"
```
tbhi, line 33: old attribute can be removed?

spec/system/admin/reports/estimated_reads_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      ]
+    end
+    let(:expected_rows) do
+      [[school.school_group.name, school.default_issues_admin_user.name, school.name, meter.mpan_mprn.to_s, meter.name,
+        '', 'NHH AMR', meter.supplier.name, meter.data_source.name, '',
+        Time.zone.yesterday.iso8601, Time.zone.yesterday.iso8601, '60', '30']]
+    end
+  end
+
+  it 'allows csv download'
```
tbhi, line 51: missing

app/controllers/admin/reports/amr_validated_readings_controller.rb
tbhi, line 60: clause needed for zero usage and EST?

app/services/amr/data_feed_translator.rb
```diff
@@ -139,5 +140,14 @@ def find_meter_by_serial_number(meter_serial_number)
...
       @map_of_fields_to_indexes ||= @config.map_of_fields_to_indexes
     end
+
+    def estimated?(amr_data_feed_row)
+      return false unless @config.reading_status_fields.any?
+
+      flags = @config.estimate_flags.any? ? @config.estimate_flags : AmrDataFeedConfig::ESTIMATED_STATUS
+
+      statuses = @config.reading_status_indexes.map { |reading_index| amr_data_feed_row[reading_index] }
+      statuses.any? { |s| flags.include?(s) } # rubocop:disable Style/ArrayIntersect
```
tbhi, line 150: `statuses.intersect?(flags.to_a)` works?

tbhi (approved):

app/views/admin/reports/base_meter_reports/index.html.erb
ldodds: Reports not switch over to bs5 yet so still needed.

app/controllers/admin/reports/amr_validated_readings_controller.rb
ldodds: Added

PR #5592, opened 2026-08-13
Add parsed_date column to amr_data_feed_reading

ldodds:
Currently we store unparsed dates in `AmrDataFeedReading.reading_date`. Not sure why this decision was made early in development, perhaps to make it easier to clean up data if CSV formats change. But the current strategy has a lot of limitations:

- makes querying the table awkward, e.g. for producing meter loading reports, we've partly worked around this by replicating some of the date parsing logic in postgres
- makes validating dates harder, as they need to be reparsed and SQL/Ruby logic differs
- adds overhead when loading data for validation as dates have to be reparsed each time
- the Date parsing code was letting through invalid and incorrectly parsed dates (this is partly fixed by code extracted from this PR and already merged in master, see below)
- results in duplicate rows stored (we currently upsert using a combined index on mpan and reading_date, so if formats vary we can have multiple rows)

This PR changes the `AmrDataFeedReading` table to add a `parsed_date` column. It also updates the loader to ensure that column is populated for any newly loaded data. The API loaders (solar, n3rgy, perse) have also been updated to populate the new column.These changes are fairly straightforward as they involve adding a `parsed_date` key to the hash of values used to upsert unvalidated readings.

These changes should be backwards compatible as the rest of the application will continue to use `reading_date` for the moment. 

This is an initial step towards further refactoring, most of the above issues will be addressed once we can build on this PR.

## Validation 

The loader was already rejecting unparseable dates and future dates, although was using logic inconsistent to that used to parse dates when loading unvalidated readings. It was also parsing the dates multiple times.

Problems with using `Date.strptime` with incorrect date formats, and assumptions in `Date.parse` fallback can lead us to producing dates like `0020-01-01`. These should mostly be resolved (see below) but might still happen.

The validation logic now just relies on `parsed_date` and includes an extra check to reject dates earlier than 1900. 

## Backfilling `parsed_date`

There's also a script to backfill `parsed_date` for all ~20m readings in the system. It iterates over the table to parse and update dates in batches to try and improve performance. The script has to use the same Ruby parsing logic to avoid misinterpreting the data.

The script will be run overnight or over a weekend as it takes approximately 3 hours even when running locally. We can run on test server to help benchmark.

## Date parsing issues

Reviewing the backfilled dates highlighted some Date parsing issues that have been resolved by: https://github.com/Energy-Sparks/energy-sparks/pull/5609.


lib/tasks/database/populate_parsed_date.rake
```diff
@@ -0,0 +1,46 @@
...
+        parsed_date = AmrDataFeedConfig.date_from_string_using_date_format(row.reading_date, row.date_format)
+        next unless parsed_date
+
+        # Build a VALUES tuple: (id, 'parsed_date')
+        "(#{row.id}, #{conn.quote(parsed_date)}::date)"
+      end
+
+      next if rows.empty?
+
+      sql = <<~SQL.squish
```
ldodds, line 27: This is to avoid touching the created_at/updated_at timestamps. We rely on updated_at during loading of unvalidated reading to order the readings so we only end up using the most recently loaded. Doing the insert in batches to try and improve performance.

tbhi (approved):

tbhi:
should loaders be a part of this or perhaps a separate PR? e.g. https://github.com/Energy-Sparks/energy-sparks/blob/6a0ad755d684d1a774a70b5fc6e67f7b13c2c2ab/app/services/amr/data_feed_upserter.rb#L116 could be updated to set parsed_date

ldodds:
> should loaders be a part of this or perhaps a separate PR? e.g.
> 
> https://github.com/Energy-Sparks/energy-sparks/blob/6a0ad755d684d1a774a70b5fc6e67f7b13c2c2ab/app/services/amr/data_feed_upserter.rb#L116
> 
> could be updated to set parsed_date

Realised this morning that the solar, n3rgy and other loaders need updating. I'm planning to do that in this PR.

PR #5594, opened 2026-08-17
Resolve issues destroying schools

ldodds:
Fixes some issues when destroying archived schools:

- disabling users fails when model is invalid, so disable validations on `disable!` and `enable!`
- ensure rtone variant installations are removed from meter, these will be destroyed when school destroyed
- remove meter issues before destroying school. Currently failing for some schools. Seems to be due to an Issue referencing both schools and meters, so destroying with the school fails as not yet destroyed the meter. Added explicit call to remove the meter association, the issue will then be destroyed when school destroyed

### DIFF
--- a/app/models/meter.rb
+++ b/app/models/meter.rb
@@ -76,7 +76,7 @@ class Meter < ApplicationRecord
   has_many :issues, through: :issue_meters
   has_many :meter_attributes, dependent: :destroy
   has_many :meter_monthly_summaries, dependent: :destroy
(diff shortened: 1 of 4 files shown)


tbhi:
does `has_one :rtone_variant_installation, required: false, dependent: :nullify` work since rtone_variant_installations.meter_id is not null?

ldodds:
No you're right that didn't fix it. I've change the association behaviour and it works now.

tbhi (approved):

PR #5595, opened 2026-08-17
Change default number precision to three significant figures

tbhi:
main change for three significant figures and test fixes


ldodds (approved):

spec/lib/format_unit_spec.rb
ldodds, line 8: Could probably remove this too? There's a reference to :ks2 still in `analytics/test_support/run_benchmarks.rb` but that can be removed.

tbhi: will remove in 2nd PR

PR #5603, opened 2026-08-18
Basic SolarEdge v2 OAuth workflow

ldodds:
Implements the basic Oauth workflow documented in [the SolarEdge v2 API](https://api-docs.solaredge.com/docs/developer-platform/b087f82c79d78-authentication#the-flow).

Specifically:

- provides ability for admins to see a "Connect SolarEdge Site" link which links to SolarEdge Connect to start the workflow. (Later PR will add ability to automatically email links to users)
- provides a callback handler at `/oauth/solar-edge` to handle flow where users have agreed to give us access

The callback handler exchanges the `code` provided by SolarEdge and uses that to create or update a `SolarEdgeInstallation` adding the access and refresh tokens provided by the SolarEdge API. We use the `external_id` to provide a School id that can be used to create/find the installation.

The user is then shown a success page or an error page if there are issues including missing params, school cannot be found, or API errors from SolarEdge. Admins will get an email notification when site is successfully connected.

I've had to remove some validations and tweak behaviour around managing SolarEdgeInstallations to allow us to create installations without an MPAN or API key. This is to allow us to continue with v1 until November (we may still need to add some sites) but begin adding new sites using v2.

Follow-up PRs will:

- Add ability for admins to email link to users/other email to request that site is connected
- Handling of behaviour if user clicks "Deny" rather than giving consent. Currently there's no redirect to our site when this happens. User just stays on SolarEdge domain. Have reported a bug/requested clarification from SolarEdge
- Query/report to confirm which installations are not consented
- Implementing actual support for the v2 API. Access tokens are only valid for 2 hours so we'll need to refresh them frequently. Am storing timestamp on model.

For now the goal here is to allow us to start getting consent for access sites via OAuth so we can complete the other changes.


app/views/admin_mailer/solar_edge_site_connected.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  </p>
+<% else %>
+  <p>
+    This is an new SolarEdge site without an API key. We will start loading data for this site once we have implemented
+    v2 of the Solar Edge API.
+  </p>
+<% end %>
+
+<p>
+  <% link_to 'View solar feeds', school_solar_feeds_configuration_index_url(@installation.school) %>
```
tbhi: `<%= `

app/controllers/oauth/solar_edge_controller.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+module Oauth
+  class SolarEdgeController < ApplicationController
+    skip_before_action :authenticate_user!
+
+    def callback
+      @site_id, @code, @school_id = params.expect(:site_id, :code, :external_id)
```
tbhi, line 8: is it ok to trust the parameters?

tbhi (approved):

app/controllers/oauth/solar_edge_controller.rb
ldodds: I wasn't sure what else to do here. Normally I'd have expected the URL to be signed or something. I don't think there's any SQL injection issues and it will only create or update the installation if the exchanging `code` for a refresh token works successfully. Not sure there's anything useful we can pass via `external_id` to sign/validate the request ourselves.

PR #5608, opened 2026-08-20
Add index to amr_validated_reading

ldodds:
Added a combined index on meter_id, status. Speeds up the queries that look for specific statuses, e.g. EST, PROB.

### DIFF
--- a/db/migrate/20260820134134_add_status_index_to_validated_reads.rb
+++ b/db/migrate/20260820134134_add_status_index_to_validated_reads.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class AddStatusIndexToValidatedReads < ActiveRecord::Migration[8.1]
+  disable_ddl_transaction!
+  def up
+    add_index :amr_validated_readings,
+              %i[meter_id status],
+              name: 'idx_amr_validated_reading_meter_status',
+              algorithm: :concurrently
+  end
+
+  def down
+    remove_index :amr_validated_readings,
+                 name: 'idx_amr_validated_reading_meter_status',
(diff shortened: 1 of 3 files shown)


db/migrate/20260820134134_add_status_index_to_validated_reads.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class AddStatusIndexToValidatedReads < ActiveRecord::Migration[8.1]
+  disable_ddl_transaction!
+  def up
```
tbhi, line 5: def change would work?

tbhi:
> Added a combined index on meter_id, status. Speeds up the queries that look for specific statuses, e.g. EST, PROB.

any particular queries? I note `Meters.latest_est_date_sql` but looks like that would benefit from reading_date in the index too

tbhi (approved):

db/migrate/20260820134134_add_status_index_to_validated_reads.rb
ldodds: Would need to use reversible block, so not really much in it? Will leave for now.

ldodds:
> > Added a combined index on meter_id, status. Speeds up the queries that look for specific statuses, e.g. EST, PROB.
> 
> any particular queries? I note `Meters.latest_est_date_sql` but looks like that would benefit from reading_date in the index too

Yes that query also the report for PROB data, which doesn't use the date. Explain plan and performance is better with the EST query with this index. I'm going to leave as it for now until have had feedback on report structure from ops team, as might end up reworking/refactoring whole query.

PR #5609, opened 2026-08-21
Fix date parsing issues

ldodds:
We parse unvalidated reading dates when loading from database. The parsing is done against the configured date format, with a fallback to `Date.parse` is there are errors.

During testing of https://github.com/Energy-Sparks/energy-sparks/pull/5592 I noticed some issues with the current behaviour, specifically:

```
Date.strptime('12-05-2022', "%Y-%m-%d") => Fri, 20 May 0012
```

And

```
Date.strptime('12-05-22', "%d-%m-%Y") => Fri, 20 May 0012
```

Neither of those calls to `strptime` generate errors but they result in misinterpreted dates. 

In cases where `strptime` throws an error, `Date.parse` can still misinterpret two digit years:

```
Date.parse('12-05-22')  => Tue, 22 May 2012
```

So there's an additional check in the fallback to also handle this.


app/models/amr_data_feed_config.rb
```diff
@@ -90,6 +90,56 @@ class AmrDataFeedConfig < ApplicationRecord
...
+  end
+
+  private_class_method def self.safe_parse_date(date_string, date_format)
+    # Avoids this: Date.strptime('12-05-2022', "%Y-%m-%d") => Fri, 20 May 0012
+    return Date.parse(date_string) if YEAR_MONTH_DAY_FORMATS.include?(date_format) && date_string.match(FOUR_DIGIT_YEAR)
+
+    # Avoids this: Date.strptime('12-05-22', "%d-%m-%Y") => Fri, 20 May 0012
+    # And this: Date.parse('12-05-22') => Tue, 22 May 2012
+    if DAY_MONTH_YEAR_FORMATS.include?(date_format) && date_string.match(TWO_DIGIT_YEAR)
+      format = date_string.include?('/') ? '%d/%m/%y' : '%d-%m-%y'
```
urbanwide: These two lines are a repeat of two earlier. Perhaps they can be put in a method?

app/models/amr_data_feed_config.rb
```diff
@@ -90,6 +90,56 @@ class AmrDataFeedConfig < ApplicationRecord
...
+  # parse the dates differently.
+  #
+  # In the second we fallback to Date.parse
+  def self.date_from_string_using_date_format(date_string, date_format)
+    return nil if date_string.blank?
+
+    safe_parse_date(date_string, date_format)
+  rescue ArgumentError
+    begin
+      if date_string.match(TWO_DIGIT_YEAR) # Avoid ruby default of assuming year/month/day order
```
urbanwide: Think you could use match? rather match here (and in some other places?)

urbanwide (approved):
Couple of minor suggestions / comments

PR #5631, opened 2026-09-02
Only attempt contact update if user is in Mailchimp audience

ldodds:
We occasionally get an error from Mailchimp when updating a contact. It appears to be due to the email address not being in the audience. All users should be in the audience, with preferences set. But users may have opted out or been removed. Which might explain why the API returns an error.

Changes just ensures that when updating a contact we first check if they are present, then do the update.

When deleting a user we only remove tags if the user is in the audience.

### DIFF
--- a/app/services/mailchimp/audience_manager.rb
+++ b/app/services/mailchimp/audience_manager.rb
@@ -1,5 +1,5 @@
 module Mailchimp
-  class AudienceManager
+  class AudienceManager # rubocop:disable Metrics/ClassLength
     def initialize(client = Rails.configuration.mailchimp_client)
       @client = client
     end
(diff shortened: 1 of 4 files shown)


app/services/mailchimp/audience_manager.rb
```diff
@@ -53,6 +53,9 @@ def remove_tags_from_contact(email_address, tags_to_remove)
     end
 
     def update_contact(mailchimp_contact, original_email = nil)
+      contact = get_list_member(mailchimp_key(mailchimp_contact, original_email))
```
tbhi, line 56: perhaps get_list_member should distinguish between member not found (404?) and other errors as seems now all would be ignored

app/services/mailchimp/audience_manager.rb
```diff
@@ -53,6 +53,9 @@ def remove_tags_from_contact(email_address, tags_to_remove)
...
+      return unless contact
```
tbhi, line 57: other callers of update_contact need updating for nil? seems like app/jobs/mailchimp/email_updater_job.rb#L11 and app/services/mailchimp/audience_updater.rb#L18

tbhi (approved):

PR #5634, opened 2026-09-03
Update annotations and schema following upgrades

ldodds:
This PR updates annotated model and db schema due to recent upgrades.

Upgrade to `annotaterb` has changed default format for columns with Rails backed enums. [See report](https://github.com/drwl/annotaterb/issues/373). This is likely to be fixed in future release, but changes now reduces noise.

I've also added `show_enums` option to annotaterb config to include postgres backed enums in the annotations. The enum used for the column is now named and all enums used by the table are listed in the annotations along with their values. The enum information is always present/obvious in the model files so I think useful to have included by default.

There is also an change to `db/schema.rb`. The upgrade to Postgres 18 apparently includes changes to how SQL is extracted and serialised which means the SQL includes in `db/schema.rb` is slightly different. The key changes are:

- In materialised views used for comparison reports we were typically using `data` as the record_type provided to `jsonb_to_record` to extract variables. Postgres now seems to prefer using `json` as the variable name and now quotes that everywhere
- In the baseload anomaly query the SQL export has removed some unnecessary column qualification

Neither of these are functional changes.

### DIFF
--- a/app/models/amr_validated_reading.rb
+++ b/app/models/amr_validated_reading.rb
@@ -24,7 +24,6 @@
 #  idx_amr_validated_reading_meter_status                    (meter_id,status)
 #  index_amr_validated_readings_on_meter_id_and_one_day_kwh  (meter_id,one_day_kwh)
 #  index_amr_validated_readings_on_reading_date              (reading_date)
-#  unique_amr_meter_validated_readings                       (meter_id,reading_date) UNIQUE
 #
 # Foreign Keys
 #
--- a/.annotaterb.yml
+++ b/.annotaterb.yml
@@ -5,3 +5,4 @@
 :exclude_tests: true
 :sort: true
 :hide_limit_column_types: 'integer,boolean'
+:show_enums: true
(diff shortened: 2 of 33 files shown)


app/models/amr_validated_reading.rb
```diff
@@ -24,7 +24,6 @@
 #  idx_amr_validated_reading_meter_status                    (meter_id,status)
 #  index_amr_validated_readings_on_meter_id_and_one_day_kwh  (meter_id,one_day_kwh)
 #  index_amr_validated_readings_on_reading_date              (reading_date)
-#  unique_amr_meter_validated_readings                       (meter_id,reading_date) UNIQUE
```
tbhi, line 27: is this right?

tbhi (approved):

app/models/amr_validated_reading.rb
ldodds: It's not right, the index is still there but `annotaterb` keeps removing the annotation. I can't see a reason why its not included and it's a bit annoying to keep manually adding back in. Easy to miss during a commit. What would you suggest?

tbhi: agree, hopefully will be re-added by a future version

PR #5635, opened 2026-09-03
Optimise duplicate reading checks during loading

ldodds:
Currently we check whether an incoming data file has duplicate readings for the same combination of mpxn and reading_date. We then end up rejecting all but the last duplicate, so we only load the last entry in the file

Checking the database shows that that the validation is being triggered so is worth keeping. Although might need to investigate root cause of the duplicates.

However the way we're doing the duplicate checks involves repeatedly slicing and scanning the parsed readings to look for any that have the same mpxn/reading_date. This is going to be very inefficient for larger files as we're checking every row against all subsequent rows.

This PR fixes that by building an index that captures which readings are duplicated. They uses that to add a warning to all but the last one.

Hopefully this will have a big improvements for the larger TGP files (30mb+) files in particular.

### DIFF
--- a/app/models/amr_reading_data.rb
+++ b/app/models/amr_reading_data.rb
@@ -78,6 +78,8 @@ class AmrReadingData
   end
 
   def invalid_row_check
+    create_duplicate_checking_index
+
     @reading_data.each_with_index do |reading, index|
       reading_date = reading[:reading_date]
       readings = reading[:readings]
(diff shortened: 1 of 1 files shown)


app/models/amr_reading_data.rb
urbanwide, line 97: I know this isn't something you've touched, but this could also be removed, if easy enough to do anyway.

urbanwide (approved):
Looks good. Could a basic test be added to test this new behaviour?

app/models/amr_reading_data.rb
ldodds: I've removed that in https://github.com/Energy-Sparks/energy-sparks/pull/5592

ldodds:
> Looks good. Could a basic test be added to test this new behaviour?

There's an existing test which confirms that a) duplicates are flagged and b) that only one duplicate is flagged:

```
    context 'with duplicate rows' do
      let(:reading_data) do
        mpan_mprn = '1234050000000'
        readings = Array.new(48, '0.0')
        [
          { mpan_mprn:, readings:, reading_date: '2019-01-01' },
          { mpan_mprn:, readings:, reading_date: '2019-01-01' }
        ]
      end

      it_behaves_like 'it has a warning' do
        let(:warnings) { 1 }
        let(:warning_type) { :duplicate_reading }
      end
    end
```

I thought that was enough to confirm that the existing behaviour was preserved?

urbanwide:
Yes - makes sense now I've re-read the top description. Good to go 👍🏻

PR #5636, opened 2026-09-04
Change system to use parsed_date

ldodds:
Once we've back-filled the `AmrDataFeedReading` table on production using the script in https://github.com/Energy-Sparks/energy-sparks/pull/5592 we can merge these changes.

- [x] Start using `parsed_date` when loading unvalidated readings.
- [x] Change unvalidated reading report to use parsed_date
- [x] Change meter loading report to use parsed_date
- [x] Review and update any other reports where we can use parsed_date

At a later date we could also change parsed_date column to be not null, to match reading_date. Haven't done this here as want to confirm all rows populated successfully after backfill and ongoing loading.

### DIFF
--- a/app/services/amr/analytics_unvalidated_amr_data_factory.rb
+++ b/app/services/amr/analytics_unvalidated_amr_data_factory.rb
@@ -41,7 +41,7 @@ module Amr
              .pluck(
                :meter_id,
                :amr_data_feed_config_id,
-               :reading_date,
+               :parsed_date,
                :created_at,
                :readings,
                :estimated
@@ -66,7 +66,7 @@ module Amr
     def reading_if_valid(_meter_id, reading)
       return if reading_invalid?(reading)
 
-      reading_date = date_from_string_using_date_format(reading)
+      reading_date = reading[1]
       return if reading_date.nil?
 
       OneDayAMRReading.new(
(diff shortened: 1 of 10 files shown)


ldodds:
@tbhi can you review for tomorrow's release please?

spec/factories/amr_data_feed_readings_factory.rb
```diff
@@ -1,6 +1,7 @@
 FactoryBot.define do
   factory :amr_data_feed_reading do
     reading_date  { Date.yesterday.strftime('%b %e %Y %I:%M%p') }
```
tbhi: perhaps re-use parsed_date here

app/services/amr/analytics_unvalidated_amr_data_factory.rb
```diff
@@ -66,7 +66,7 @@ def build_meter_data(active_record_meter)
     def reading_if_valid(_meter_id, reading)
       return if reading_invalid?(reading)
 
-      reading_date = date_from_string_using_date_format(reading)
+      reading_date = reading[1]
```
tbhi, line 69: bigger change but seems clearer to just pass through as arguments rather than having an array

tbhi (approved):

app/services/amr/analytics_unvalidated_amr_data_factory.rb
ldodds: Will do this in a separate PR

PR #5638, opened 2026-09-04
limited data report

tbhi:


ldodds (approved):
Changes to add the Limited Data report look good. The Trello card discussion also included a note about adding a "Stale data" report. Could you add that in a separate PR?

PR #5641, opened 2026-09-07
Flag up if the same school has recorded the same activity as taking place on the same date

urbanwide:
This is how it currently looks, with placement under the date field (didn't think being above would work, as the form field would be pushed down).

image

In Trello, it is suggested perhaps to 'tweak the save button'. Not sure if this means, add a confirmation message once save has been clicked @ldodds? - something like 'This activity has already been recorded for this school on this date, are you sure you want to save another one?'


ldodds:
re: "tweak the save button", I was thinking instead of always "Record action" something like "Record duplicate action"/"Record action again". But was just thinking out loud.

btw, for activities we have "Save activity" but actions as "Record action". Should perhaps be "Record" for both?

urbanwide:
Ready for review

In conversations with Nick yesterday, he reminded me that sometimes full classes log the same activity, on the same date, which now would bring up this warning for all but the first. Presume we are ok with this?

ldodds:
> Ready for review
> 
> In conversations with Nick yesterday, he reminded me that sometimes full classes log the same activity, on the same date, which now would bring up this warning for all but the first. Presume we are ok with this?

Yes, I've highlighted this as well. So I assume we are OK with this.

ldodds (approved):
Changes look good. Tested locally with activity recording and works as expected.

PR #5642, opened 2026-09-08
Use school switcher on choose school page

tbhi:


app/views/redirects/choose_school.html.erb
```diff
@@ -19,7 +19,7 @@
       <ul>
         <% @schools.each do |school| %>
           <li>
-            <%= link_to school.name, school_redirect_path(school, @path) %>
+            <%= link_to school.name, school_switcher_path(school_id: school.id, path: @path), method: :post %>
```
ldodds: Only school admins with multiple schools should go via the switcher and have their association change. Group admins that are shown the "choose school" page should just follow a redirect link as before. We don't really want group admins to be linked directly to schools.

ldodds (changes requested):

ldodds (approved):

PR #5652, opened 2026-09-10
Move activities controller and views to be scoped to schools

urbanwide:
Doesn't change the url, just moves the controller to be schools/activities_controller.rb and views to /schools/activities/
(as per interventions)


app/controllers/schools/activities_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+          ActivityType.find(params.expect(:activity_type_id))
+        elsif @activity.present?
+          @activity.activity_type
+        end
+    end
+
+    def set_breadcrumbs
+      return unless activity_type
+
+      activity_type.category
```
tbhi: remove?

tbhi (approved):

PR #5667, opened 2026-09-15
correct school impact enrolled number - school group not always set on onboarding

tbhi:


tbhi:
replaced with https://github.com/Energy-Sparks/energy-sparks/pull/5669

PR #5669, opened 2026-09-15
correct school impact enrolled number - school group not always set on onboarding

tbhi:
Changed query to find onboarding from school rather than school group.  For the school group with an issue 2 of the 3 onboarding objects don't have school group id set.  Not sure if that's correct but seems safe to go via the school.


app/services/school_groups/impact_report/generator/overview.rb
```diff
@@ -31,21 +31,17 @@ def three_months_ago = @three_months_ago ||= generated_at - 3.months
...
-            .joins(:events)
-            .where(
-              school_onboarding_events: {
-                event: SchoolOnboardingEvent.events[:onboarding_complete],
-                created_at: twelve_months_ago..
-              }
-            )
-            .distinct
-            .count
+          school_group.schools.joins(school_onboarding: :events)
```
urbanwide: Should we go through assigned_schools here rather than schools?

app/services/school_groups/impact_report/generator/overview.rb
```diff
@@ -31,21 +31,17 @@ def three_months_ago = @three_months_ago ||= generated_at - 3.months
...
+                               event: SchoolOnboardingEvent.events[:onboarding_complete],
+                               created_at: twelve_months_ago..
+                             })
+                      .distinct
+                      .count
         end
 
         # schools still enrolling
-        def enrolling_schools = school_group.onboardings_for_group.incomplete.count
+        def enrolling_schools = school_group.schools.joins(:school_onboarding).merge(SchoolOnboarding.incomplete).count
```
urbanwide: I think the issue here could be that in the early stages of onboarding, there are some points when an onboarding doesn't have an associated school. I am not sure what this point is, but worth checking if we care about these.

app/services/school_groups/impact_report/generator/overview.rb
tbhi: corrected

app/services/school_groups/impact_report/generator/overview.rb
tbhi: makes sense - have reverted this part

urbanwide (approved):
Looks good

PR #5671, opened 2026-09-17
Select featured school based on activity over last 12 months

urbanwide:
It was incorrectly selecting based on activity this academic year
Also updates podium text as suggested by Claudia


spec/components/impact_reports/engagement/featured_school_component_spec.rb
```diff
@@ -47,6 +47,16 @@
       it { expect(page).to have_text('recorded 1 pupil activity in the last 12 months') }
     end
 
+    context 'with an action this academic year and an activity in a previous academic year but in the last 12 months' do
+      before do
+        create_list(:observation, 1, :intervention, school: school, at: yesterday)
```
tbhi: could just be create for 1?

app/components/impact_reports/engagement/featured_school_component.rb
tbhi, line 73: wonder if there's an issue as podium only seems looks at current academic year? a default school with points only in the previous academic year wouldn't render?

tbhi (approved):

app/components/impact_reports/engagement/featured_school_component.rb
urbanwide: I did look at this, but believe the podium needs to be for this academic year. So because the podium won't have any points for the default school, it would be best not to display the whole component without it.
The podium code is very convoluted and this might need a revisit but no time now.

PR #5672, opened 2026-09-18
Basic activity index page

urbanwide:
Adds feature flag, controller, routes, spec.
Puts header at top of page and breadcrumbs

Likely to change but a starting place to hang things off


config/locales/views/activities/activities.yml
```diff
@@ -33,6 +33,10 @@ en:
       update_activity: Update activity
       update_duplicate_activity: Update duplicate activity
       when_did_you_complete_this_activity: When did you complete this activity?
+    index:
+      description: Energy Sparks provides extensive support to teachers and pupils when learning about energy and climate change within the context of your own school
+      description_action_html: Use the links below to explore <strong>%{count}</strong> freely available activities
```
tbhi, line 38: plurization needed?

app/controllers/activities_controller.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class ActivitiesController < ApplicationController
+  before_action :enable_bootstrap5
+  before_action :enable_feature
+  before_action :set_breadcrumbs
+
+  # load_and_authorize_resource
```
tbhi: remove?

tbhi (approved):

config/locales/views/activities/activities.yml
urbanwide: Probably not - there will always be more than 1 activity 👍🏻

